PR #1, opened 2015-01-29
HWKALERTS-1 Prepare an initial version

lucasponce:
Inital import with following requeriments:

1. Create a hawkular-alerts-api as public API for hawkular-alerts-engine.
2. hawkular-alerts-engine is a decoupled component that can work in embedded/bus/third party scenarios.
3. Integration with hawkular-bus is defined into hawkular-alerts-bus component.
4. hawkular-alerts-api can be accessed via REST through hawkular-alerts-rest wrapper.
5. Bus distribution is packaged in a .ear archive.
6. Create a hawkular-notifiers-api as a common library to use by notifiers plugins.
7. Notifiers plugins will be separate artifacts to be deployed into the bus.

About the flow:
a. hawkular-alerts-bus listens on MetricsTopic topic, and send alerts data into hawkular-alerts-engine.
b. hawkular-alerts-bus registers a callback/listener into hawkular-alerts-engine to process specific notifications flow.
c. Notifications callback/listener is responsible to send a notification message into NotificationsTopic topic.
d. Each plugin is listening into NotificationsTopic with proper filtering to process incoming notifications.
e. Registration of new plugins sends a registration message into NotifierTypeRegisterQueue.
f. hawkular-alerts-bus listens for new plugins registration requests and update definitions service to access through alerts API to the list of notifiers/type (for UI purposes).


jshaughn:
Lucas, I'm sure this is a good start, I think we should just use this as our initial content on go from there.

PR #7, opened 2015-02-04
Add Data timestamps to allow time-ordered insertions.

jshaughn:
May also help with potential temporal reasoning, correlation, and more robust notification. The
timestamps are supplied by the app sending the Data and will default to
system time.
Also:
- Change XxxConditionEval constructors to pass in the relevant Data objects.
  A little cleaner in the rules, and a bit more flexible.
- update and standardize the model's toString() impls given the new fields
  and variations in the previous impls. Used Eclipse's generator.
- update equals/hashcode impls due to field additions. Used Eclipse generators.


lucasponce:
It seems OK.
I haven't fully check if all REST tests passed but no problem, go ahead with the merge, REST api has to be updated with some missing methods, so I will review it and fix any potential flaw.

PR #9, opened 2015-02-06
HWKALERTS-3 Implement an Email notification sender

tsegismont:
Initial implementation.

The PR is not ready yet.

I don't know how I can get configuration information for the EmailListener:
- from address
- to address

### DIFF
--- a/hawkular-notifiers-email/src/main/java/org/hawkular/notifiers/email/notifications/EmailListener.java
+++ b/hawkular-notifiers-email/src/main/java/org/hawkular/notifiers/email/notifications/EmailListener.java
@@ -16,14 +16,22 @@
  */
 package org.hawkular.notifiers.email.notifications;
 
-import org.hawkular.bus.common.consumer.BasicMessageListener;
-import org.hawkular.notifiers.api.log.MsgLogger;
-import org.hawkular.notifiers.api.model.NotificationMessage;
-import org.jboss.logging.Logger;
-
+import javax.annotation.Resource;
 import javax.ejb.ActivationConfigProperty;
 import javax.ejb.MessageDriven;
 import javax.jms.MessageListener;
+import javax.mail.Address;
+import javax.mail.Message;
+import javax.mail.Message.RecipientType;
+import javax.mail.MessagingException;
+import javax.mail.Session;
+import javax.mail.Transport;
+import javax.mail.internet.InternetAddress;
+import javax.mail.internet.MimeMessage;
+
+import org.hawkular.bus.common.consumer.BasicMessageListener;
+import org.hawkular.notifiers.api.log.MsgLogger;
+import org.hawkular.notifiers.api.model.NotificationMessage;
 
 /**
  * An example of listener for emails processing.
@@ -37,9 +45,26 @@ import javax.jms.MessageListener;
         @ActivationConfigProperty(propertyName = "messageSelector", propertyValue = "NotifierType like 'email'")})
 public class EmailListener extends BasicMessageListener<NotificationMessage> {
     private final MsgLogger msgLog = MsgLogger.LOGGER;
-    private final Logger log = Logger.getLogger(EmailListener.class);
+
+    @Resource(mappedName = "java:jboss/mail/Default")
+    Session mailSession;
 
     protected void onBasicMessage(NotificationMessage msg) {
-        msgLog.infoNotificationReceived("Email", msg.toString());
+        try {
+            Message message = createMimeMessage(msg);
+            Transport.send(message);
+        } catch (MessagingException e) {
+            msgLog.errorCannotSendMessage("email", e.getLocalizedMessage());
+        }
+    }
+
+    Message createMimeMessage(NotificationMessage msg) throws MessagingException {
+        Message message = new MimeMessage(mailSession);
+        message.setFrom(new InternetAddress("noreply@hawkular"));
+        Address toAddress = new InternetAddress("root@localhost");
+        message.addRecipient(RecipientType.TO, toAddress);
+        message.setSubject("Hawkular alert");
+        message.setContent(msg.getMessage(), "text/plain");
+        return message;
     }
 }
--- a/hawkular-notifiers-email/src/main/java/org/hawkular/notifiers/email/notifications/EmailListener.java
+++ b/hawkular-notifiers-email/src/main/java/org/hawkular/notifiers/email/notifications/EmailListener.java
@@ -63,7 +63,7 @@ public class EmailListener extends BasicMessageListener<NotificationMessage> {
         message.setFrom(new InternetAddress("noreply@hawkular"));
         Address toAddress = new InternetAddress("root@localhost");
         message.addRecipient(RecipientType.TO, toAddress);
-        message.setSubject("Hawkular alert");
+        message.setSubject("Hawkular alert " + msg.getNotifierId());
         message.setContent(msg.getMessage(), "text/plain");
         return message;
     }
--- a/hawkular-notifiers-email/src/main/java/org/hawkular/notifiers/email/notifications/EmailListener.java
+++ b/hawkular-notifiers-email/src/main/java/org/hawkular/notifiers/email/notifications/EmailListener.java
@@ -61,9 +61,9 @@ public class EmailListener extends BasicMessageListener<NotificationMessage> {
     Message createMimeMessage(NotificationMessage msg) throws MessagingException {
         Message message = new MimeMessage(mailSession);
         message.setFrom(new InternetAddress("noreply@hawkular"));
-        Address toAddress = new InternetAddress("root@localhost");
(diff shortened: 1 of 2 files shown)


lucasponce:
Hi,
The initial idea is that the NotificationMessage should have a notifierId.
The notifierId is an abstraction of the specific details of a destination.
The plugin should register notifiers with generic properties, as notifierId can be reused by several triggers.
In the current status, the registration is not yet implemented, there is a dummy initialization via files here:
https://github.com/hawkular/hawkular-alerts/blob/master/hawkular-alerts-engine/src/main/resources/hawkular-alerts/notifiers.data
This has to evolved to something more robust, but the idea is that once you get a message, the plugin should get from the notifierId the needed data for processing (i.e. from/to addresses, phone number, decoration text, etc).
I return on monday and I can explain better the idea.

lucasponce:
Hi, for the "to", "from", "templates" and specific notifier properties I suggest the following flow:

1.- The plugin registers via RegistrationInit, so then, the hawkular-alerts-engine can have all properties shared per this plugin.
2.- The DefinitionsService, via Java or REST API, can register new notifiersId, once this request comes to the alert-engine, the plugin would expect a NotifierRegistrationMessage with specific details per that notifierId (here it's where fieds like "to"/"from"/"template" should be expected).
3.- The notification plugin can register locally these properties, I guess this can be an optional task and it can depend of the implementations defails of the plugins (Another option is to create a dependency, with the alerts-engine, so the notification plugin can consume DefinitionsService to query this specific data).
4.- A NotificationMessage can be processed and with notifierId info, plugin can get all properties to process the message.

tsegismont:
>Hi, for the "to", "from", "templates" and specific notifier properties I suggest the following flow:
>
>1.- The plugin registers via RegistrationInit, so then, the hawkular-alerts-engine can have all properties shared per this plugin.
>2.- The DefinitionsService, via Java or REST API, can register new notifiersId, once this request comes to the alert-engine, the plugin would expect a NotifierRegistrationMessage with specific details per that notifierId (here it's where fieds like "to"/"from"/"template" should be expected).

So this means that in `org.hawkular.notifiers.email.registration.RegistrationInit`, I have to send the list of properties needed to send a notification, correct?

>3.- The notification plugin can register locally these properties, I guess this can be an optional task and it can depend of the implementations defails of the plugins (Another option is to create a dependency, with the alerts-engine, so the notification plugin can consume DefinitionsService to query this specific data).

So a new MDB is needed to listen to the new "NotifierRegistrationMessage"? And the MDB would take care of storing the options in memory?

>4.- A NotificationMessage can be processed and with notifierId info, plugin can get all properties to process the message.

From your previous points, I understand that we can get the default configuration for all notifications, but how would that map to notifierId?

lucasponce:
>
> Hi, for the "to", "from", "templates" and specific notifier properties I
> suggest the following flow:
>
> 1.- The plugin registers via RegistrationInit, so then, the
> hawkular-alerts-engine can have all properties shared per this plugin.
> 2.- The DefinitionsService, via Java or REST API, can register new
> notifiersId, once this request comes to the alert-engine, the plugin would
> expect a NotifierRegistrationMessage with specific details per that
> notifierId (here it's where fieds like "to"/"from"/"template" should be
> expected).
>
> So this means that in
> org.hawkular.notifiers.email.registration.RegistrationInit, I have to
> send the list of properties needed to send a notification, correct?
>
RegistrationInit should send all properties shared by the plugin. So, in
this way, the alerts-engine through the DefinitionsService will have all
properties related to the UI.
Some of these properties can make sense for read-only purposes, other can
be read-write.
For example:
- An email plugin should send here the name of the host/port.
- An SMS, should send the name of the gateway/service.


>  3.- The notification plugin can register locally these properties, I
> guess this can be an optional task and it can depend of the implementations
> defails of the plugins (Another option is to create a dependency, with the
> alerts-engine, so the notification plugin can consume DefinitionsService to
> query this specific data).
>
> So a new MDB is needed to listen to the new "NotifierRegistrationMessage"?
> And the MDB would take care of storing the options in memory?
>
This is the goal of this component:

https://github.com/hawkular/hawkular-alerts/blob/master/hawkular-notifiers-email/src/main/java/org/hawkular/notifiers/email/registration/RegisterListener.java


To register notifiers.
A new notifier can be:
"notifierA", "Send to admin1@corp.org and admin2@corp.org" a notification
message with "templateA".

So, from the UI, DefinitionsService will register a new notifierId for a
notifierType with some properties that are specific for this notifier.
Plugin can store in memory/cache these properties, but in any case, the
alerts-engine persistence will have these properties too.
NotifierID can be shared across multiple triggers definitions, so it is
good to have this separation.

>  4.- A NotificationMessage can be processed and with notifierId info,
> plugin can get all properties to process the message.
>
> From your previous points, I understand that we can get the default
> configuration for all notifications, but how would that map to notifierId?
>
In the RegisterListener, each time we have a new notifierId, the plugin
should receive a NotifierRegistrationMessage, so it can be cached/stored, I
guess this is open to the plugin's implementation.
Or, another option is that each time we receive a  NotificationMessage ,
the plugin can query to DefinitionService for the properties of specific
notifierId.

Both options are valid, it's open, the idea is that plugin notifier
skeleton can handle:
- Registration of the plugin -> So, alerts-engine will be aware.
- Listen for new notifiers coming from the UI.
- Listen for notifications to process specific send (email/sms/whatever).

tsegismont:
@lucasponce I believe the configuration for notifiers topic deserves its own JIRA and PR. If you agree, I'll create a new ticket and will continue the discussion in the ML/forum (as you and @jshaughn prefer).

I just pushed a new commit for this PR in particular. It takes into account that the mail recipient should be the notifierId (in the current model). I have a Postfix server installed on my laptop and I verified that email is sent to root@localhost (with modified notifiers and trigger data files).

Is that sounds good to you, can you merge this PR?

lucasponce:
I am ok to merge this PR now and refine later.
And we can put these comments in a JIRA to continue the discussion there.

jshaughn:
I'm OK if we merge this if it helps for an interim solution.  But overall I think we need more discussion...

I'm not sure I'm completely happy with the notifier-id approach.  That may be because I don't fully understand yet what is being proposed. But, from what I can tell the idea is to have pre-defined notifier-ids with formats that are particular to its relevant notifier:

#notifier id, notifier type,description
SNMP-Trap-1,snmp,description
admin@email.com,email,description
SMS-555-12345,sms,description

It may be useful to have pre-defined notifierIds, which from what I can tell are notifier configurations under a different name. But if we go that way we'll need to also have security/tenants, etc to ensure only the allowable notifier-ids are available to the Trigger definer.

I tend to think that configuring Notifiers for Triggers should be done in a fashion not unlike RHQ, where the user adds the desired notifiers to the Trigger, configuring each as needed. There may be some redundancy, but predefined notifier configurations seems like overkill.  If necessary, I suppose we could always query for other, existing [and accessible] notifier configurations and present them, in the presentation layer.

Going a bit further, I'm not sure I like the idea of Notifier CRUD via the API.  This may create a disconnect between the deployed notifiers (MDBs) and their definitions.  Maybe we should consider having a deployed notifier register itself by dropping a registration message on an Alerts Queue. As in the API, the notifier would declare it's type (basically a unique name), and define it's configurable properties.  And maybe those properties should be exactly that, a set of properties (name, description, default-value) requiring name-value pairs to be defined. That could lead to a simple UI.  The registration could be ignored for existing notifiers, or maybe the properties could be updated if necessary.

So, for a Trigger there would be 0 or more configured notifiers (maybe 1 or more?).  When a trigger fires an alert we would publish an Alert message to an Alerts topic with a header containing the configured notifier types and a body containing a bunch of properties, the notifier configurations and the actual Alert information. (Are property values sufficiently powerful?).  The notifier configuration properties could be prefixed with the notifier type.  The notifier subscribers would then consume the alert message and do their work.

In the code today we send a message to each notifier. That approach is still OK but I think there are advantages to just publishing a single message to an Alerts topic:
- The engine does not need to know anything about the notifiers other than the type name to be placed in the headers.
- The engine needs only send one message to a Topic that is managed by alerting, so failure mgmt may be simplified.
- The Alerts topic could be subscribed to easily by any other consumer of alerts.

On a related note, I think it will probably be better to have the rules engine generate the Alerts, but process them outside of the engine.  Meaning, read Alerts from the global, which Lucas has already created for us.  After firing the rules, check the global, process the Alerts in the Java code (persist, send message(s), etc). And then clear the global for the next rules execution.

I think we should meet to discuss this further. I'm not sure a long PR thread is the best way to hash out the design.

tsegismont:
Le 10/02/2015 23:32, Jay Shaughnessy a écrit :
> I'm OK if we merge this if it helps for an interim solution. But overall
> I think we need more discussion...

> I think we should meet to discuss this further. I'm not sure a long PR
> thread is the best way to hash out the design.
>

That's the sense of my previous comment.

PR #10, opened 2015-02-06
Jshaughn/data insert

jshaughn:
The main work for this PR is here, and has to do with changes to Data insertion and rules execution.  I'm looking now at some more unit tests for the engine.


tsegismont:
Looks good to me. Maybe at some point we can split RulesEngineTest in multiple tests classes.

PR #11, opened 2015-02-10
Add Travis config files

tsegismont:
These files are needed to activate Travis support

### DIFF
--- a/.travis.build.bus.sh
+++ b/.travis.build.bus.sh
@@ -19,8 +19,10 @@
 
 set -xe
 
+BUS_BUILD_LOG="${HOME}/bus.build.log"
+
 cd "${HOME}"
 
 git clone --depth=50 https://github.com/hawkular/hawkular-bus.git hawkular/hawkular-bus
 cd hawkular/hawkular-bus
-mvn clean install -DskipTests
+mvn clean install -DskipTests > "${BUS_BUILD_LOG}" 2>&1 || (cat "${BUS_BUILD_LOG}" && exit 1)
--- a/.travis.build.bus.sh
+++ b/.travis.build.bus.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+#
+# Copyright 2015 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


tsegismont:
The build fails because the bus snapshot artifacts are not in JBoss Snapshots repo

tsegismont:
From IRC

```
<tsegismont> Alerts cannot be built on Travis because the bus snapshot artifacts are not in JBoss Snapshots repo
<tsegismont> https://travis-ci.org/hawkular/hawkular-alerts/builds/50197352#L237
<lponce> tsegismont: yes, we depend on the work of the bus, mazz was working in an assembly, so once it's finished we can automate it
<tsegismont> What do you mean? The build needs org.hawkular.bus:hawkular-bus-mdb:jar:1.0.0.Alpha1-SNAPSHOT
<tsegismont> and probably other artifacts
<lponce> tsegismont: mvn clean install, only executes local junit test, but not integration tests, those ones are separated into a different rest profile
<lponce> tsegismont: I want to put it together in an automated way, to be able to deploy and run integration tests also, but I need some work in bus to do that, like a similar plugin for wildfly but with the bus
<tsegismont> Ok. But the problem here is just the compile phase
* jdoyle s'appelle maintenant jdoyle_away
<lponce> tsegismont: yeap, you are right, there is some artifacts from the bus project that are missing
<lponce> but how it can be addressed that ? because we depend of artifacts that are in SNAPSHOT state
<lponce> so I don't see as a good rule to perform a formal release for those
<lponce> travis can download/execute a mvn clean install for bus to populate its internal repo ?
<lponce> as similar we do in a local environment ?
<lponce> I'm not familiar with travis, may be I'm asking something trivial or impossible to do, I don't know :)
* jdoyle_away s'appelle maintenant jdoyle
<tsegismont> Ideally, all dependency should be in an external repo
<tsegismont> It should be possible to clone and build the bus project in pre-install phase
<tsegismont> But that means any time we want to test a PR on alerts we're also testing bus master branch
<lponce> if we are talking about adding 3 more minutes, I don't think is bad at all, as we can test is some last commit in bus can broke its dependencies
<lponce> but anyway, I don't know if that is acceptable
<lponce> for it can be if there is not an easier workaround
<tsegismont> I'll add it like this for now. but we'll need to remove this later when bus deps stabilize
<lponce> +1
```

tsegismont:
So we will build Hawkular bus before testing alerts. Now Travis builds passes but before merging I'd like to make the bus build quiet (if not failing).

tsegismont:
Ready for merging

lucasponce:
Reviewed. It looks OK. Thanks to add alerts into travis world :-).

PR #17, opened 2015-02-17
[HWKALERTS-2] Initial support for Trigger Safety Mode

jshaughn:
A Trigger can now be in FIRE mode or SAFETY mode. A Condition set and
Dampening can be defined for both modes on the same Trigger. Based on the
Trigger's current mode the relevant conditions and dampening will be used
for evaluations.
- introduce trigger mode across the various constructor/hashcode/equals/toStrings
- add safetyEnabled to indicate that safety conditions are set and enabled
- make ConditionEval.getLog() impls more verbose w/ timestamp info
- update the DB and File def service impls as needed
- add test code

ToDo: Currently satisfying the safety dampening only toggles the Trigger
      back to fire mode.  We need to discuss whether we should be generating
      an alert and/or performing actions.

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/Condition.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/Condition.java
@@ -16,6 +16,8 @@
  */
 package org.hawkular.alerts.api.model.condition;
 
+import org.hawkular.alerts.api.model.trigger.Trigger.Mode;
+
 /**
  * A base class for condition definition.
  *
@@ -30,6 +32,11 @@ public abstract class Condition {
     protected String triggerId;
 
     /**
+     * The owning trigger's mode when this condition is active
+     */
+    protected Mode triggerMode;
+
+    /**
      * Number of conditions associated with a particular trigger.
      * i.e. 2 [ conditions ]
      */
@@ -42,15 +49,16 @@ public abstract class Condition {
     protected int conditionSetIndex;
 
     /**
-     * A composed key for conditionId
+     * A composed key for the condition
      */
-    protected String conditionId;
+    protected String id;
 
-    public Condition(String triggerId, int conditionSetSize, int conditionSetIndex) {
+    public Condition(String triggerId, Mode triggerMode, int conditionSetSize, int conditionSetIndex) {
         this.triggerId = triggerId;
+        this.triggerMode = triggerMode;
         this.conditionSetSize = conditionSetSize;
         this.conditionSetIndex = conditionSetIndex;
-        this.conditionId = triggerId + "-" + conditionSetSize + "-" + conditionSetIndex;
+        updateId();
     }
 
     public int getConditionSetIndex() {
@@ -59,6 +67,7 @@ public abstract class Condition {
 
     public void setConditionSetIndex(int conditionSetIndex) {
         this.conditionSetIndex = conditionSetIndex;
+        updateId();
     }
 
     public int getConditionSetSize() {
@@ -67,6 +76,7 @@ public abstract class Condition {
 
     public void setConditionSetSize(int conditionSetSize) {
         this.conditionSetSize = conditionSetSize;
+        updateId();
     }
 
     public String getTriggerId() {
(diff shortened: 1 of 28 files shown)


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/Condition.java
```diff
@@ -75,43 +85,59 @@ public String getTriggerId() {
...
-        if (conditionSetSize != condition.conditionSetSize) return false;
-        if (triggerId != null ? !triggerId.equals(condition.triggerId) : condition.triggerId != null) return false;
+    public String getConditionId() {
+        return conditionId;
+    }
 
-        return true;
+    private void updateId() {
+        StringBuilder sb = new StringBuilder(triggerId);
+        sb.append("-").append(triggerMode.ordinal());
```
lucasponce: I wonder if triggerMode.ordinal() needs to be in the conditionId.
My first thought is not, as the triggerMode should not affect into the relationship between Trigger <-> Condition.
I mean, a condition is linked with a triggerId, and conditionSetSize + conditionSetIndex is important, but I think we won't have a collision with the triggerMode, so IMO I would put it out of the conditionId.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DbDefinitionsServiceImpl.java
```diff
@@ -196,7 +197,10 @@ private void initFiles(String folder) {
                 initNotifiers(initFolder);
             }
         } catch (Exception e) {
-            msgLog.errorDatabaseException("Error initializing files. Msg: " + e.getMessage());
+            if (log.isDebugEnabled()) {
```
lucasponce: A small (a non-important comment). debugf method invokes the isDebugEnabled() so you can save the check outside in the code.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/MemDefinitionsServiceImpl.java
```diff
@@ -41,6 +41,7 @@
 import org.hawkular.alerts.api.model.condition.ThresholdRangeCondition;
 import org.hawkular.alerts.api.model.dampening.Dampening;
 import org.hawkular.alerts.api.model.trigger.Trigger;
+import org.hawkular.alerts.api.model.trigger.Trigger.Mode;
```
lucasponce: Side comment: I left here the MemDefinitionsServiceImpl.java just to have an "example" of a different implementation. But I am ok if we dropped as DbDefinitionsServiceImpl is the important one.
Both are in "draft/demo/poc" status, so perhaps just one can be enough and we don't need to maintain it.

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -378,7 +379,37 @@ rule AlertOnSatisfiedDampening
...
+
         $d.reset();
         update( $d )
+
+        if ($t.isSafetyEnabled()) {
+            if (log != null && log.isDebugEnabled()) {
+                log.debug("Setting Trigger to Safety Mode! " + $t);
+            }
+            $t.setMode(Mode.SAFETY);
+            update( $t )
```
lucasponce: At this point we need to re-feed the DefinitionsService in some way.
When the Trigger is modified, from the API, the UI should take account of this change, so I guess we need to implement this flow, as the safety mode can be set up from the UI, right ?

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -378,7 +379,37 @@ rule AlertOnSatisfiedDampening
...
+    then
+        if (log != null && log.isDebugEnabled()) {
+            log.debug("SetFiringModeOnSatisfiedDampening! " + $d.log());
+        }
+
+        $d.reset();
+        update( $d )
+
+        $t.setMode(Mode.FIRE);
+        update( $t )
```
lucasponce: Same idea like the previous one: as we can modify the trigger from the UI, if the engine changes the status this should be updated via DefinitionsService to have a consistency between UI and persistence.
If not we can have the risk to have some "detached objects" in the session engine from the DefinitinosService.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/Condition.java
lucasponce: I think that instead to save an ordinal (that is hard to remind) it can be also good to store the literal of the triggerMode, I guess that at this time it's ok and we can gain for clarity.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DbDefinitionsServiceImpl.java
jshaughn: But I am not using log.debugf.  I was just printing the stack trace.  I'm not sure if we want to do this here but more than once I found it valuable.  If you have a suggestion for how you want this code snippet to look let me know, I will change it.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/MemDefinitionsServiceImpl.java
jshaughn: That is fine.  We can drop it after this PR is merged, it will reduce any further work with refactoring.

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
jshaughn: This is an interesting point.  I agree that the UI may want to show the current mode although I'm wondering whether Trigger.mode should even be persisted.  It may be best as a transient field, it's not really part of the Trigger definition but rather a runtime value.  If the server is restarted I think it may be OK if all Triggers go back to Firing mode.  We should discuss this behavioral point further and probably get some feedback from others.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/Condition.java
jshaughn: I am OK to change it to triggerMode.name(),  I will make that change.

PR #19, opened 2015-02-25
Jshaughn/api

jshaughn:


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -37,31 +37,132 @@
...
+     * calls. The updated </code>Trigger</code> will be persisted.  If enabled the </code>Trigger</code>
+     * will be [re-]inserted into the Alerts engine and any prior dampening will be reset.
+     * @param triggerId
+     * @throws Exception If the </code>Trigger</code> does not exist.
+     */
+    Trigger updateTrigger(Trigger trigger) throws Exception;
+
     Trigger getTrigger(String triggerId) throws Exception;
 
+    Collection<Trigger> getAllTriggers() throws Exception;
```
lucasponce: Disclaimer: I'm using this PR not only to review but also for brainstorming :-).
I am wondering if it is time to introduce a tags/groups feature in the model.
We have requeriments as https://issues.jboss.org/browse/HWKALERTS-12 where we should handle "groups/profiles" in some way.
Without enter in a very low level detail yet, I guess that this can help us to prototype some kind of tagging mechanism.
My main doubt/fear with grouping is that at the end we should not depend of the inventory model, I mean, our tagging/group solution should be valid with or without inventory in the equation.
Having said that, looking at the DefinitionsService I ask myself if we should have also a Collection<Trigger> getTriggersByTag(Set<Tag> tags) or similar to have a starting point.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -37,31 +37,132 @@
...
+        CRUD interface for Dampening
+     */
+
+    Dampening addDampening(Dampening dampening) throws Exception;
+
+    void removeDampening(String dampeningId) throws Exception;
+
+    Dampening updateDampening(Dampening dampening) throws Exception;
+
+    Dampening getDampening(String dampeningId) throws Exception;
```
lucasponce: With the Dampening class I have also a main doubt: does it need to be part of the Trigger definition itself ?
I mean, from a public API perspective, all interesting info is in 1 to 1 relation with the Trigger class.
Dampening class also has transient info that is only valid for the internal inference inside the engine, but it should not visible from an API perspective IMO.
So, from a pure API perspective I'm wondefing if we should move public data of Dampening object into the Trigger and keep Dampening object but not in public api model, but in a private model that is interesting only for the engine.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -37,31 +37,132 @@
...
+     *   conditionSetSize
+     *   conditionSetIndex
+     * </pre>
+     * @param triggerId
+     * @param triggerMode
+     * @param condition Not null
+     * @return The updated, persisted condition set
+     * @throws Exception
+     */
+    Collection<Condition> addCondition(String triggerId, Trigger.Mode triggerMode, Condition condition)
```
lucasponce: +1
I think yesterday I was misunderstanding the chat.
So +1 to be able to add a condition with the triggerId and the new Condition object and it's a good idea to have the updated persisted condition set.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -37,31 +37,132 @@
...
 
+    /**
+     * @param triggerId
+     * @param triggerMode Return only conditions for the given trigger mode. Return all if null.
+     * @return The existing conditions for the trigger. Not null.
+     * @throws Exception
+     */
+    Collection<Condition> getTriggerConditions(String triggerId, Trigger.Mode triggerMode) throws Exception;
+
+    Collection<Condition> getAllConditions() throws Exception;
```
lucasponce: Also, same thought for Triggers in relation to tagging/groups is valid here with conditions.
By default, a Trigger is not attached to a resource (what is good IMO) as we can have flexibility to define triggers with multiple resources/metrics. Also I'm thinking if the tagging/group mechanism can also be applied to conditions, so, we can get all conditions of a specific tag, with no matters in which triggers depends on.
So, in theory this is interesting for operations when we may want to apply a template/conditions to a groups of resources (at least without enter in low level details it can be a start point to refine later).

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -74,9 +175,13 @@
         i.e. querying for properties to fill for a specific notifier type.
      */
     void addNotifierType(String notifierType, Set<String> properties) throws Exception;
```
lucasponce: I am +1 to change the Notifier / Notifications terminology.
Without enter in a full discussion, I would start by the name.
I like the concept of Action instead of Notifier as you mentioned in some discussion, and I would propose ActionPlugin as NotifierType.
We can start with these changes (and propagating into the whole hawkular-alerts) maintaining current structure, and think/change structure later.
What do you think ?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -37,31 +37,132 @@
...
+     * will be [re-]inserted into the Alerts engine and any prior dampening will be reset.
+     * @param triggerId
+     * @throws Exception If the </code>Trigger</code> does not exist.
+     */
+    Trigger updateTrigger(Trigger trigger) throws Exception;
+
     Trigger getTrigger(String triggerId) throws Exception;
 
+    Collection<Trigger> getAllTriggers() throws Exception;
+
```
lucasponce: Also, as a general note for DefinitionsService: we should introduce also pagination helpers.
So, perhaps we should include in the API methods to getCountTrigger() or similar and getTriggers(from, to).
Also, we may need some kind of filtering/ordering (I don't know, but thinking loud about it).

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
```diff
@@ -30,18 +30,25 @@
 public interface AlertsService {
 
     void sendData(Data data);
+
     void sendData(Collection<Data> data);
 
     Collection<Alert> checkAlerts();
```
lucasponce: I think is also interesting to start introducing more rich methods for getting the alerts.
From the UI we can have the use case to get the alerts for an specific trigger, and also we should combine the tagging/grouping mechanism of the definitions into the alerts.
Also, filtering by time interval will be something to add.
An, in addition of all below, we also should add getCount() methods to be able to deal with pagination as well.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
jshaughn: We may want a more general getTriggersByCriteria(TriggerCriteria).  Just a catch-all filtering object that keeps the API the same while allowing an increasing number of filters.  As for Tags, I do think we need these although we should have explicit use cases for which they are useful.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
jshaughn: The original Trigger object was actually a heavy object, which we normalized given the original (current) CRUD API.  So this change was supporting that normalized approach.  Actually, Triggers encompass not only dampening but also the conditions.  I am OK with re-designing the model and API such that Trigger comprises its dampening and conditions.  What we need to avoid is something in the middle, which is what we had before this proposed change.

Dampening and Condition will still need to be objects understood by the calling code, we are talking only about their access, and possible their storage.   I think Dampening and Condition should be treated the same way in the API.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
jshaughn: We need to discuss tagging further as it relates to templates.  We need to just discuss templates :)

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
jshaughn: I am +1 to the "Action" terminology.  It can be introduced gradually.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
```diff
@@ -25,13 +25,16 @@
...
-        FIRING, SAFETY
+        FIRE, SAFETY
     };
 
     private String id;
     private boolean enabled;
+    private boolean safetyEnabled;
     private Mode mode;
 
+    private transient Match match;
```
lucasponce: I have an open question and I don't have a good answer for that.
I wonder if the public api.model would be more DTO oriented rather than have hidden logic.
I mean, even if we mark some properties as transient, those are there and they are part of the REST interface.
So I wonder if we should have a single model with everything (public properties or private properties used by the engine but not needed from the API) or may be to split model, and to have a plubic model very simple and POJO oriented that will be used for integration, and have private/hidden properties in a private model used by the engine.
Both options have pros/cons, so I don't have an answer but I bring it to the review.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/AvailabilityCondition.java
```diff
@@ -16,8 +16,11 @@
  */
 package org.hawkular.alerts.api.model.condition;
 
+import static org.hawkular.alerts.api.model.trigger.Trigger.Mode.FIRE;
+
 import org.hawkular.alerts.api.log.MsgLogger;
```
lucasponce: Another thing that I have on my todo list is to review the logger dependency in the API.
In the conditions we have the match() logic and we have to log a possible error if operator is not correct.
Thinking this model can be shared, I think this is a reason to have a more DTO oriented approach.
We can have all match() methods as a part of the engine implementation, then we can also simplify the model and get rid of the logging dependencies for the hawkular-alerts-api artifact what is good.

lucasponce:
Today I was reviewing this branch. I have updated the hawkular-bus and in a clean environment I get an error like: 

Caused by: java.lang.IllegalStateException: JBAS014232: @PostConstruct method of EJB singleton DbDefinitionsServiceImpl of type org.hawkular.alerts.engine.impl.DbDefinitionsServiceImpl has been recursively invoked

Probably as now in the DbDefinitionsServiceImpl has a reference to AlertsService, so the initialization tasks of the database (the population of demo data) should be moved out of the @PostConstruct method. Perhaps we can check in the @PostConstruct if the database is empty or not, and use a flag to populate once in a call of a regular method, I guess this can solve the issue.

Let me know if you are working on it, or if you want me to take a look.

lucasponce:
I think I have an easy fix for this one, I will prepare a PR for your branch.

jshaughn:
This PR is not going to be merged.  Instead we have copied this branch to the "wintermute" branch.  Work will continue there, outside of a PR.  When wintermute is merged we'll remove close this PR and delete the jshaughn/api branch.  The wintermute branch will act as our "next version" branch, and will also have an analogously named branch in the UI repos.

PR #21, opened 2015-03-05
HWKALERTS-24 Refactoring of notifications

lucasponce:
This PR has:
- A major renaming of Notifier/NotifierType/Notification per "Action" and "ActionPlugin".
- A refactor of plugins architecture:
* A Plugin registers into the alerts engine at deployment time.
* Register/Deregister callback from engine to plugins is deleted.
* An action message will have the properties.


hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/init/AlertEngineRegister.java
```diff
@@ -38,11 +38,11 @@
     private final Logger log = Logger.getLogger(AlertEngineRegister.class);
 
     @EJB
-    NotificationsService notifications;
+    ActionsService notifications;
```
jshaughn: another place where the var name is now dated.

jshaughn:
Lucas, I am fine with this PR. Only the minor comments. It can be merged.

lucasponce:
Ok thanks. I've updated these minor changes and merged it.

PR #22, opened 2015-03-10
Add Json anotations into API model

lucasponce:
As a previous step to review REST end points, I have added Json annotations into the API model.
This can also help in the documentation between interaction from Java vs JSON model.
So, with this approach we can avoid to maintain a parallel model inside the engine.
REST endpoints are not yet reviewed, those will come in future commits.

### DIFF
--- a/hawkular-alerts-api/pom.xml
+++ b/hawkular-alerts-api/pom.xml
@@ -54,6 +54,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-jackson2-provider</artifactId>
+      <scope>provided</scope>
+    </dependency>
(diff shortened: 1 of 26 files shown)


jmazzitelli:
Does everyone use Jackson annotations? Or RESTEasy? I'm not sure - we should probably make sure everyone is consistent in using the same annotations.

lucasponce:
True.

I've took hawkular-metrics as an example:

https://github.com/hawkular/hawkular-metrics/blob/master/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/BucketedOutput.java#L24

lucasponce:
Thx for your review Jay, tomorrow morning I will fix those changes and update the PR for merging.

PR #25, opened 2015-03-16
Swagger docs auto push

Jiri-Kremser:
Part of this PR is also upgrading swagger to 2.10 to alight with other projects.
The `responseClass` annotation attribute was removed in 2.10, hence the `responseContainer` and `response` attributes. This is the suggested way now, afaik.

Travis build won't invoke the build of the website immediately (=> tons of builds), but once the website is built it takes this up-to-date adoc file by this script (and adds a jBake header). Basically the poll model.

the "other side" of the mechanism:
https://github.com/hawkular/hawkular.github.io/blob/pages/.travis.swagger.sh

It uses GitHub API for pushing/reading the files, because githum mvn plugin is quite heavy-weight and we don't need to clone/push whole repo (it's expensive in Travis and it prolongs the build time).


Jiri-Kremser:
Ouch, the branch must be named differently, sorry for the confusion, I'll create another one. (There can't be 2 different PRs with the same (from) branch)

PR #26, opened 2015-03-16
Swagger docs auto push

Jiri-Kremser:
Part of this PR is also upgrading swagger to 2.10 to alight with other projects.
The `responseClass` annotation attribute was removed in 2.10, hence the `responseContainer` and `response` attributes. This is the suggested way now, afaik.

Travis build won't invoke the build of the website immediately (=> tons of builds), but once the website is built it takes this up-to-date adoc file by this script (and adds a jBake header). Basically the poll model.

the "other side" of the mechanism:
https://github.com/hawkular/hawkular.github.io/blob/pages/.travis.swagger.sh

It uses GitHub API for pushing/reading the files, because githum mvn plugin is quite heavy-weight and we don't need to clone/push whole repo (it's expensive in Travis and it prolongs the build time).


Jiri-Kremser:
I have no idea, why the build is failing because of missing licence header in the bash script. It's the same script as in the metrics and there it is passing. Upgrading the hawkular-parent to the latest doesn't help :rage3:

lucasponce:
Jay and me have to do a refactor of REST endpoints between today and tomorrow max, so I will focus on that for the MVP and after that I will try to merge your PR, to see what can be happening.

Jiri-Kremser:
@ppalaga any idea why the build fails here and passes in hawkular-metrics? Something with licence header check.

ppalaga:
`.travis.swagger.sh` is new. Its copyright should be 2015 rather than 2014 - 2015. 
`mvn license:format` is always worth trying when you are in doubt.

Jiri-Kremser:
@ppalaga `mvn license:format` worked as a charm, thanks!

PR #27, opened 2015-03-18
Instruct license plugin to ignore *.data files in hawkular-alerts-engine

ppalaga:


ppalaga:
plz ignore, sent against wrong branch

PR #33, opened 2015-04-01
Action Email fixes

tsegismont:
Action Email fixes

EmailListener saved the CC recipient in the TO field, and printed a dash in the subject even when the description property was not set.
Also, the EmailListenerTest was wrongly moved during a prior refactoring, so it was not executed anymore (and thus was not updated when the ActionMessage properties field was introduced).


tsegismont:
@lucasponce , @jshaughn : can one of you please review? Thanks

lucasponce:
It looks good, thanks !

PR #34, opened 2015-04-01
HWKALERTS-31 Create a Twilio SMS notification sender

tsegismont:
The plugin expects to find a few system properties to configure the Twilio client:
service id, token, from number

If the client can't be initialized, an error message is logged and when messages arrive they're just dropped.

The PR has a unit test and the integration was tested with a free Twilio account


tsegismont:
@jshaughn Lucas is on PTO, it's all in you hands :smile: :gift:

PR #40, opened 2015-04-29
HWKALERTS-34 Add a Cassandra implementation

lucasponce:
Initial branch with support for Cassandra persistence.
The "rest" profile has been extended also in hawkular-alerts-engine and hawkular-alerts-rest-tests projects to provide a server with Cassandra embedded as similar way it works in Kettle.
There is space for improvement, but we can have this start point here.

Please, help me to review & test it and check how it goes.
Thanks,


lucasponce:
I've reviewed and I've found a couple of minor things, I'll update the PR.

jsanda:
I have only briefly looked at the changes. I noticed that service classes are implemented as EJBs. If I remember correctly, the container will start a transaction for methods like `CassAlertsServiceImpl.addAlerts` since no transaction attribute is specified. Cassandra does not support transactions (in the ACID sense), so enlisting business methods in transactions does not make a whole lot of sense unless of course there are some transactional resources involved. To be honest, I see little if any benefit of using EJB with Cassandra backend. You can get the dependency injection by just using CDI managed beans.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -0,0 +1,849 @@
...
+            throw new IllegalArgumentException("Alerts must be not null");
+        }
+        if (session == null) {
+            throw new RuntimeException("Cassandra session is null");
+        }
+        if (insertAlert == null) {
+            throw new RuntimeException("insertAlert PreparedStatement is null");
+        }
+        try {
+            for (Alert a : alerts) {
```
jsanda: Since the driver is async, it is typically better to do these inserts in parallel something like,

```java
List<ResultSetFuture> futures = new ArrayList<>;
for (Alert a : alerts) {
    futures.add(session.executeAsync(insertAlert.bind(a.getAlertId(), toJson(a)));
    futures.add(session.executeAsync(insertAlertTrigger.bind(a.getAlertId(), a.getTriggerId())));
    futures.add(session.executeAsync(insertAlertCtime.bind(a.getAlertId(), a.getCtime())));
    futures.add(session.executeAsync(insertAlertStatus.bind(a.getAlertId(), a.getStatus().name())));
}
// If any of the futures fail, the future returned from Futures.allAsList will fail. This is a somewhat simplistic approach to error handling, particularly if you are inserting a lot of alerts, but it is a good starting point. If the returned future fails, you simply retry the inserts. 
return Futures.allAsList(futures);
```

Doing async, parallel inserts will be much faster and more scalable  than doing blocking inserts particularly as the number of inserts increases. I might also be inclined to further refactor my code to eliminate some of the boiler plate, creating some reusable functions in the process,

```java
Function<List<BoundStatement>, List<ResultSetFuture>> insertAsync = statements ->
                statements.stream().map(session::executeAsync).collect(toList());

Function<Alert, List<BoundStatement>> bindAlert = alert -> asList(
                insertAlert.bind(alert.getAlertId(), toJson(alert)),
                insertAlertTrigger.bind(alert.getAlertId(), alert.getTriggerId()),
                insertAlertCtime.bind(alert.getAlertId(), alert.getCtime()),
                insertAlertStatus.bind(alert.getAlertId(), alert.getStatus().name()));

return Futures.allAsList(alerts.stream()
                .map(bindAlert)
                .map(insertAsync)
                .flatMap(List::stream)
                .collect(toList()));
```

jshaughn:
Other than the few questions I asked everything looks good to me, nice job, a lot of work. I think we are close to a merge, just please respond to my comments and respond to jsanda's comments.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
jshaughn: OOh, all fancy with Java8, gotta love it.   Jsanda, thanks for looking at this stuff.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -0,0 +1,849 @@
...
+            if (triggerIds.size() > 0) {
+                filterByTriggers = true;
+                StringBuilder sAlertsTriggers = new StringBuilder("SELECT alertId FROM ").append(keyspace)
+                        .append(".alerts_triggers WHERE ");
+                if (triggerIds.size() == 1) {
+                    sAlertsTriggers.append("( triggerId = '");
+                    sAlertsTriggers.append(triggerIds.iterator().next());
+                    sAlertsTriggers.append("' )");
+                } else {
+                    sAlertsTriggers.append("( triggerId IN (");
```
jsanda: Be cautious with using an IN clause. It is generally considered an anti-pattern except in the case where are you only querying a single partition. The problem is that it puts a lot of pressure on the coordinator node. The coordinator has to wait for responses from every replica involved in the request before responding to the client. If any of those internal requests time out, the whole request fails from the client's perspective. It is generally better to issue multiple queries in parallel and merge results on the client side. Then you can retry those individual queries that fail rather than all of them.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -0,0 +1,849 @@
...
+                }
+                if (criteria.getEndTime() != null) {
+                    if (andNeeded) {
+                        sAlertsCtimes.append(" AND ");
+                    }
+                    sAlertsCtimes.append("( ctime <= ");
+                    sAlertsCtimes.append(criteria.getEndTime());
+                    sAlertsCtimes.append(" )");
+                }
+                sAlertsCtimes.append(" ALLOW FILTERING ");
```
jsanda: I noticed use of `ALLOW FILTERING` in a few places. That is probably a good indication that you want to revisit your schema design to better optimize for reads. Check out http://www.datastax.com/dev/blog/allow-filtering-explained-2 for some background.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -0,0 +1,849 @@
...
+                session.execute(insertAlertCtime.bind(a.getAlertId(), a.getCtime()));
+                session.execute(insertAlertStatus.bind(a.getAlertId(), a.getStatus().name()));
+            }
+        } catch (Exception e) {
+            msgLog.errorDatabaseException(e.getMessage());
+            throw e;
+        }
+    }
+
+    public List<Alert> getAlerts(AlertsCriteria criteria) throws Exception {
```
jsanda: This is pretty big method with a lot going on. I may have missed some things, but it basically looks like it is doing,

* Fetch alert ids from alert_triggers table if filtering by trigger ids
* Fetch alert ids from alerts_ctimes table if filtering by ctime
* Fetch alert ids from alerts_statuses table if filtering by status
* Merge all alert ids
* Fetch alerts from alerts table

Let me first share a few, general observations. With data modeling/schema design, you want to design in such a way that optimizes for your reads. In the ideal scenario when fetching a single entity or even multiple entities,  you can satisfy the request by reading a single partition. Accomplishing this usually requires denormalizing your data model and performing additional filtering/merging on the client side. Keep in mind, I am not suggesting that doing this is easy :)

Here is an idea for an alternative approach. In each of the filter tables, alerts_triggers, alerts_ctimes, alerts_statuses, store the payload in addition to the alert id. Then for each filter that is used, you can query against each of the filter tables in parallel, merging the results on the client side using AND or OR semantics as appropriate or as dictated by the AlertCriteria. 

If an alert payload is large and/or there is unbounded growth of them, then fewer filter tables with more client-side filtering might be better. For example, maybe do something like,

```
CREATE TABLE alerts_triggers (
    triggerId text,
    ctime bigint,
    alertId text,
    status text,
    PRIMARY KEY (triggerId, ctime, alertId)
);
```
Then you can do,

```
SELECT alertId 
FROM alerts_triggers 
WHERE triggerId = ? AND ctime >= ? AND ctime <= ?
```

You would want to execute these in parallel, and then you could filter by status on the client side. Lastly, you should be using prepared statements.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
jshaughn: This is good to know.  I suppose "criteria-style" querying is not a great an idea against cassandra as it was against an RDB.  One problem we'd have trying to convert to multiple async queries would be the potential for a combinatorial issue when dealing with multiple "IN clause" criteria.  I'm not sure what the solution is here.  We could leave as-is and see how it goes, or we could be proactive and try to make a change, possibly changing the API to allow for more restricted querying.  But I can certainly imagine the need for clients to want to query for alerts for say, for multiple triggers (all the triggers for a single resource, for example).  We'll need to find the sweet spot between function and efficiency/practicality.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
jshaughn: Although I'm not going to say scale is not an issue, I would mention that alert persistence should be far from metric persistence in scale.  What I mean is, Lucas, I think we should get the cassandra impl into master so it can start to get some use, and not delay for optimization work. Especially at this time when we don't yet need to worry about upgrades.

jsanda: I absolutely agree with merging changes sooner rather than later. The best way to learn is to have people beat on it, and that won't happen until the changes get merged. With that said, a lot of the things I have mentioned are not simply matters of scale. They are recommended/best practices and anti-patterns. Moving to Cassandra is a big paradigm shift, and Lucas is off to a really nice start; however, you want to cognizant of things you should and should not do regardless of scale.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
jsanda: Cassandra is neither designed or intended for ad-hoc queries like you have with SQL. That does not however rule out criteria-style querying. Schema design though is more important as well performing work client side such as filtering and joining results.

@jshaughn I don't quite understand the combinatorial issue. Maybe we can discuss further on the mailing list or on IRC.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
Yes, parallel async insert is a good point.
But the business method would be sync, I mean, I think that propragation of the async cappabilities should be (at the moment) in the implementation detail.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
lucasponce: I see your point and I started to model in that way.
But there is a tradeoff, as I commented in one of Jay's comments, we have a requeriment of query information in the domain.
After investigate/read/learn, we have a tradeoff between the logic in the client, the design of the model in the persistence layer and the logic (queries) we send to the persistence layer. And everyone is strongly dependent of the context itself.
So, our main goal here is to move from a h2 database to a cassandra persistence, as first step.
Also, we are trying to put all information related to a same trigger into same partition, to avoid that we need to hit several partitions.
Also, the alerts domain is different of others in metrics/inventory as we are going to have a discrete number of definitions, and a very small fraction of alerts (comparing, for example with metrics).
So, I have identified some areas for improvement into the model, but also I want to define perf tests to get measures and to validate perf progress in overall (i.e. not only focus on cassandra, but also analyze the global logic, sometimes the improvements are not in the obvious place).
Also, we sill have query business requeriments that we want to maintain, so everything will help as an input in the improvement process.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
lucasponce: Thanks for the comment.
I guess I've commented the general idea in previous comments.
We need to study the global context, not only the cassandra model.
Our main goal is to prepare a start point to study which alternatives are better (to put a more denormalized and repeated model vs add more filtering/joining logic into the client).
As this is very dependent of the domain model I guess our approach can be to jump into Cassandra and make iterative improvements in small steps based as well on perf tests.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
lucasponce: @jsanda @jshaughn thanks for your comments. They are appreciated.
I think best approach is to merge the branch to have a "Cassandra start point" in alerts and to make small steps with improvements and enhancenments into me model but also based on tests in order to make our model solid.
I had some previous notes and also @jsanda has suggested new ones that are interesting, so thanks for you review.

lucasponce:
I think is interesting to merge the PR as a preliminar start point.
There are small aspect where we can work in following small commits.

PR #42, opened 2015-04-30
Moving the mustache template (for the rest docs) to build tools 

Jiri-Kremser:
To enable generation of docs when offline. Currently there is a ling to github, this won't work without connection to internet. I've moved the template file into build tools repo. Parent module depends on it, and everything depends on parent module, so it should be cached in local mvn repo for all hawkular modules.


jshaughn:
I assume this is fine, I think only Jiri knows this anyway :)

Jiri-Kremser:
thank for merge,
Thomas S and Heiko know too :)
It was discussed here https://github.com/hawkular/hawkular-metrics/pull/190

PR #43, opened 2015-05-11
HWKALERTS-44 Add a standalone factory

lucasponce:
Small commit to add a helper factory for standalone cases.
It is mainly used for tests/demo purposes but I think it is more clear rather than invoking directly the implementations.
I have thought to do it in a pure SPI style but inter-dependencies have to be resolved manually in anycase so I decided to bring a small factory for it.


pilhuhn:
Factory of *what* ?

lucasponce:
Factory of Alerts/Definitions/Actions services.

PR #47, opened 2015-05-20
Upgrade to hawkular-parent 13

ppalaga:


ppalaga:
@jshaughn please review and eventually merge

PR #49, opened 2015-05-27
Cleanup pom dependencies

theute:


theute:
https://github.com/hawkular/hawkular-alerts/pull/48 should be applied first

PR #51, opened 2015-05-27
[HAWKULAR-138] shut down C* driver session

jsanda:
We need to call ```Session.close()``` on server shutdown; otherwise, while the embedded Cassandra instance has shutdown, the driver will continue trying to reconnect, delaying the Hawkular server shutdown.


lucasponce:
Good catch.
Thx for the PR John.
I'm going to consolidate it into alerts and definitions services implementations with some validation.

lucasponce:
I've consolidated this change on wintermute:
https://github.com/lucasponce/hawkular-alerts/commit/8a25b92181a97f0be34b061a4434a0eaeedd4428
As after the HAWKULAR-40 merge, we are going to rebase master with wintermute to maintain only a single branch for development.
Thx !

PR #53, opened 2015-05-30
HAWKULAR-286 - Accounts deployed inside EAR for alerts.

jpkrohling:


jpkrohling:
@vrockai , can you check if it solves your problem?

lucasponce:
Tested in a clean invironment with -Prest profile and it passes.

PR #55, opened 2015-06-01
Jshaughn/severity

jshaughn:
HWKALERTS-53 :  Add Severity to Alerting


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -205,6 +209,10 @@ private void initPreparedStatements() throws Exception {
             insertAlertStatus = session.prepare("INSERT INTO " + keyspace + ".alerts_statuses " +
                     "(tenantId, alertId, status) VALUES (?, ?, ?) ");
         }
+        if (insertAlertSeverity == null) {
+            insertAlertSeverity = session.prepare("INSERT INTO " + keyspace + ".alerts_severities " +
+                    "(alertId, severity) VALUES (?, ?) ");
```
lucasponce: Missing insert tenantId.
TenantId is used as partition key, to distribute data across nodes. Data needs to have affinity, so with tenantId we can have all data related to a tenant on a same node, avoiding that a query should be spread across multiple nodes and executing a map/reduce to get result.

jshaughn:
Lucas, I've fixed the query issues you pointed out.  Also, I've now merged this PR on master (the refactor with CassStatement), and added a bit more alerts query testing.  I think this PR is ready to merge, please perform any further review you deem necessary and merge if you see fit.

PR #56, opened 2015-06-02
Hwkalerts 54

jshaughn:
Refactor such that AlertsService and Impl does not handle both client requests and engine running concerns.  Allow client requests (like getAlerts) to be serviced by a pool of beans, while maintaining the singleton for the engine.

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
@@ -32,27 +32,12 @@ import org.hawkular.alerts.api.model.data.Data;
  */
 public interface AlertsService {
 
-    void sendData(Data data);
-
-    void sendData(Collection<Data> data);
-
     /**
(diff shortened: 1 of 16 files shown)


lucasponce:
Hi Jay,
I'm reviewing this PR and I've seen a couple of things with the session that need to be fixed.
I've checkout your branch and I'm preparing a couple of commits to pile on that.

lucasponce:
In particular I'm working in these commits:
- Makes a share session for keyspace, instead to have two as we have.
- The current PR needs to change that, if not a multiple session creation can lead into performance issues.
- If we separate all statements into CassStatement, I think we should refactor definitions as well, I think it's confusing to have splitted this.

I hope to have ready this fixes during the day.

jshaughn:
OK, I was admittedly unsure about the Cassandra session handling and keyspace, so I'm glad you picked up on that.  I put a TODO in the code mentioning that we should probably consolidate all PreparedStatements into CassStatement.

lucasponce:
+1
Thx for the changes !

PR #57, opened 2015-06-04
HWKALERTS-56 : Need a way to fetch Alerts without the evaluations

jshaughn:
Add a "thin" query param to Alerts GET endpoint.  It currently will remove
the evalSets and resolvedEvalSets from the Alerts returned from the fetch.
This can be a good option for an initial list of alerts, where a more
robust pull of alert details can be deferred.

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/Alert.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/Alert.java
@@ -16,6 +16,10 @@
  */
 package org.hawkular.alerts.api.model.condition;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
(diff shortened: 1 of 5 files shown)


lucasponce:
I'm +1 with the PR.
The only aspect I would improve is to separate the test in a new one, to not overhead the LifeCycle tests.
I prefer even a simple one (like the manual resolution), but explicit to focus in thin vs non-thin.

I'm afraid that for us LifeCycle class is good, but for third eyes it could be kinda cryptic.

So, I let you decide if you want to re-org the test with a new one or maintain the change.

jshaughn:
The reason I keep merging sub-tests into the existing tests is just because we don't get alerts by default, we have to generate them.  And the async nature of doing that increases the testing run time.  I agree it makes it harder to see what is actually tested.  Perhaps JUnit allows for some test ordering/dependency, such that I can use alerts generated by earlier tests just to test updates and variations of fetches... I'll take a look.

lucasponce:
Perhaps I wrote it bad. I think LifeCycle class is the proper way, but separate method.

PR #58, opened 2015-06-09
Support pagination for Alerts

lucasponce:
- Introduced pagination in Java API for Alerts.
- Reused same API as inventory.
- Introduced pagination on REST endpoints for Alerts.
- Implementation based on server-side ordering as a first approach.
- Pagination not yet implemented on definitions (as priority was alerts).
- Deleting from master DB* implementation, created a h2-engine branch for it.


jshaughn:
Good Stuff, Lu[ck]as :)

PR #63, opened 2015-06-24
Hwkalerts 50

jshaughn:
This is a first version of a higher-level integration component between hawkular-alerts and hawkular-metrics.  It is currently packaged in the alerts ear but should be extracted into a separate, add-on deployment.

This work also incorporates work to initially satisfy the external alerting hooks requested in HWKALERTS-52.,


hawkular-alerts-metrics/src/main/java/org/hawkular/alerts/external/metrics/CassCluster.java
```diff
@@ -0,0 +1,87 @@
...
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.ProtocolVersion;
+import com.datastax.driver.core.Session;
+
+/**
+ * Cassandra cluster representation and session factory.
+ *
+ * @author Lucas Ponce
+ */
+public class CassCluster {
```
lucasponce: This class here triggered an alarm.
Do you need to access directly Cassandra ? from hawkular-alerts-metrics ?
I haven't finished the full review and perhaps I can find it.

hawkular-alerts-metrics/src/main/java/org/hawkular/alerts/external/metrics/MetricsServiceProducer.java
```diff
@@ -0,0 +1,48 @@
...
+ * @author john sanda
+ */
+public class MetricsServiceProducer {
+
+    private MetricsServiceImpl metricsService;
+
+    @Produces
+    public MetricsService getMetricsService() throws Exception {
+        if (metricsService == null) {
+            metricsService = new MetricsServiceImpl();
```
lucasponce: Ah, interesing, I guess that my next question should go to Metrics, but my first view is that the MetricsService should be "callable" as metrics component should offer a pool of components (EJB3 Session) that you can access too.
This Producer sounds weird here.
Because it is not using a service, it is just creating a new one, with all side effects it may have.
Anyway, I don't think this is an issue in the PR, is just an architecture issue about how Metrics offers its service to other scenarios.

lucasponce:
I did a brief revision and the logic of the feature is great.
I have a couple of doubts about what is the preferred order to merge.
As we have a HWKALERTS-17 where we can change the .ear per a .war, I think it could be worth to give a try, perhaps, then we can co-locate hawkular-alerts-metrics without need to use the .ear.
I mean, we are going to have this issue with the re-packaging effort, so next week we'll have to deal with it.
Perhaps an easy merge of HWKALERTS-17 and a next rebase of this branch can be interesting.

The weird thing is the instance of the MetricsService, I don't see why metrics component doesn't offer an EJB that you can simply connect IMHO.

jsanda:
> The weird thing is the instance of the MetricsService, I don't see why metrics component doesn't offer an EJB that you can simply connect IMHO.

JEE components like JDBC connections and EJBs are traditionally single threaded and synchronous. If you want to handle multiple concurrent requests to inserts records into the db, the container carries out those requests using multiple threads. Each thread might execute a different EJB instance and use a different JDBC connection from a pool. The DataStax driver is async, which means among other things that we can efficiently serve multiple requests concurrently from a single thread. With respect to MetricsService specifically, there simply is no need for EJB. Cassandra does not support transactions, so container or bean managed transactions are of no use. We can support concurrent access much more efficiently by simply using the async requests that the driver offers. 

If anything, now that h-alerts is using Cassandra I wonder why EJB is still being used :)

lucasponce:
Well, the EJB offers to you out of the box the pooling.
True is that JDBC os single thread, but from the thread of the EJB you can span more threads to offer an async behaviour, to handle the pooling of the logic that is not directly handled by the driver.
But I think that a cost to need to use a custom instance of the component.
And when we have different packages (.war) in this case, I wonder why not use a standard way to access to the component.
EJB is not only about transactions :-), but security, and pooling.

hawkular-alerts-metrics/src/main/java/org/hawkular/alerts/external/metrics/MetricsServiceProducer.java
jshaughn: I'm open to alternatives but MetricsService is not an EJB, just a bean.  To use the @Inject this seemed like the cleanest way, and was recommended by jsanda.

hawkular-alerts-metrics/src/main/java/org/hawkular/alerts/external/metrics/CassCluster.java
jshaughn: I admit that it seems odd but I need a cassandra session to startup MetricsService.   Perhaps we could access the existing class with a dependency on the alerts engine?  I don't love the duplication.  Also, I think jsanda is working on an actual MetricsClient that may replace the need to do this at all.

jshaughn: Actually, that is a Rest Client he's working on, so that does not apply.

jsanda:
I agree that EJB is not only about transactions. It does offer security, pooling, some dependency injection, probably more features I am not thinking of. We do not need pooling. And by *we*, I am referring to anyone using the DataStax driver. While the driver is aysnc, you do not want to block its I/O threads with any long running work. It could cause other requests to time out. When the response returns and you have some work to do, it is better to do that work on a separate thread pool. EJB does not help here. There is nothing with respect to the pooling for EJB that is needed or even useful.

And yes EJB does provide a fine-grained security model. But are you using any of the security features?

jshaughn:
Thanks for the review!  We can discuss offline the merge strategy..

PR #64, opened 2015-06-29
HWKALERTS-57

lucasponce:
Preliminar work for:
- Updating action architecture to pass alert and definitions into the plugin.
- Support for plain and html templating in email plugin.

There are some tasks that are not fully finished as it depends of other JIRAs:
- We need to unify jackson/gson libraries with jackson to simplify the json serialization/deserialization and use just one annotation (I have a JIRA for that).
- It doesn't send email on alert resolution yet, as it can create a "spam" sent of emails (thinking on "Resolve All" button), so that it will be added when UI uses the AUTO-RESOLVE mode, in order to not add noise from plugins.
- Integration profile -Drest is not working in master, as we have the WF9 on the way, when WF9 is out, I will update the profile again (the HWKALERTS-57 was tested with Alpha1 and no major issue found).

I've rebased with master to make the merge more easy, I would suggest to merge it and we can add fix for pending topics later.


jshaughn:
> It doesn't send email on alert resolution yet, as it can create a "spam" sent of emails

If we add actions to any lifecycle event I would expect that they would be configurable just like they are for the "Open" event (i.e. alert  create time).  This is probably your understanding as well, but I wanted to make sure because the sentence above made it seem like it may be fixed to an e-mail notification or something like that.

lucasponce:
Correct. At the moment, we send actions on "open" alert, but not on "ack" and "resolve".
I introduced in the PR but finally I thought is better to define first the scenario and configuration parts.
With the current UI, a "Resolve All" would send a lot of emails with no useful info.

But your point is right, it is just a matter of definition about configuration for sending actions per "ack" and "resolve" mode (perhaps we may want to send on "resolve" but not "ack"), so, anyway, the PR only sends on open to avoid introducing new behaviours that may lead into confusion.

hawkular-actions-plugins/hawkular-actions-email/src/main/java/org/hawkular/actions/email/registration/RegistrationInit.java
```diff
@@ -58,9 +59,13 @@ public void init() {
...
+             */
+
             ActionPluginMessage apMsg = new ActionPluginMessage();
-            apMsg.setOp("init");
-            apMsg.setActionPlugin("email");
+            apMsg.setOp(EmailPlugin.INIT_PLUGIN);
+            apMsg.setActionPlugin(EmailPlugin.PLUGIN_NAME);
             Set<String> properties = new HashSet<String>();
             properties.add("to");
             properties.add("cc");
```
jshaughn: Do you want to EmailPlugin constants here, instead of literal Strings?

hawkular-actions-plugins/hawkular-actions-email/src/main/java/org/hawkular/actions/email/template/EmailTemplate.java
```diff
@@ -0,0 +1,259 @@
...
+
+            processed.put("plain", writerPlain.toString());
+            processed.put("html", writerHtml.toString());
+        }
+        return processed;
+    }
+
+    private void subjects(Map<String, Object> props, Alert alert) {
+        if (props != null && alert != null && alert.getTrigger() != null && getFirstCondition(alert) != null) {
+            Condition.Type type = getFirstConditionType(alert);
```
jshaughn: I think perhaps this sort of customization should be performed only for single-condition triggers.  If the trigger has multiple conditions we should default to a generic message.  This is especially true if the Trigger's match type is "ANY", but I think it's better in general, because we just don't know what the trigger's purpose is when there are multiple conditions.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/Alert.java
```diff
@@ -89,6 +91,22 @@
...
+
+    /*
+     * This is the dampening attached to a trigger when the alert was fired.
+     * As a trigger, the dampening can change during time, but an alert should be attached with a specific instance.
+     */
+    @JsonInclude(Include.NON_EMPTY)
+    @Thin
+    private Dampening dampening;
+
     @JsonInclude(Include.NON_EMPTY)
```
jshaughn: I'm not totally convinced we need to drag around this information on an Alert, it makes a heavy object even heavier and it could increase the complexity of converting from json back to pojo if there are changes to the objects.  But I guess I'm not so opposed as to call for the removal, I'm just not sure it's a good idea.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ThresholdRangeConditionEval.java
```diff
@@ -45,6 +45,9 @@ public ThresholdRangeConditionEval(ThresholdRangeCondition condition, NumericDat
         super(condition.match(data.getValue()), data.getTimestamp());
         this.condition = condition;
         this.value = data.getValue();
+        if (this.condition != null) {
+            this.type = this.condition.getType();
+        }
     }
```
jshaughn: Hopefully keeping condition type as a member variable will be worth the trouble.  I guess it depends how much people like the targeted message templates.

Note that ExternalCondition will need this change when HWKALERTS-50 work is merged in.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -246,7 +321,7 @@
      */
 
     /**
-     * Create a new Action.
+     * Create a new Action
      *
      * @param tenantId Tenant where actions are stored
      * @param actionPlugin Action plugin where this action is stored
```
jshaughn: +1 on adding jdoc!

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
```diff
@@ -62,6 +62,7 @@
     public static final String INSERT_CONDITION_THRESHOLD;
     public static final String INSERT_CONDITION_THRESHOLD_RANGE;
     public static final String INSERT_DAMPENING;
+    public static final String INSERT_DEFAULT_PROPERTIES_ACTION_PLUGIN;
```
jshaughn: Would have preferred INSERT_ACTION_PLUGIN_DEFAULT_PROPERTIES per existing naming convention

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
```diff
@@ -88,6 +89,7 @@
     public static final String SELECT_DAMPENING_ID;
     public static final String SELECT_DAMPENINGS_ALL;
     public static final String SELECT_DAMPENINGS_BY_TENANT;
+    public static final String SELECT_DEFAULT_PROPERTIES_ACTION_PLUGIN;
```
jshaughn: Would have preferred SELECT_ACTION_PLUGIN_DEFAULT_PROPERTIES

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
```diff
@@ -108,6 +110,7 @@
     public static final String UPDATE_ACTION_PLUGIN;
     public static final String UPDATE_ALERT;
     public static final String UPDATE_DAMPENING_ID;
+    public static final String UPDATE_DEFAULT_PROPERTIES_ACTION_PLUGIN;
```
jshaughn: Would have preferred UPDATE_ACTION_PLUGIN_DEFAULT_PROPERTIES

jshaughn:
Lucas,  Overall this is good and I could have merged it as-is.  I will give a you a chance to review the comments and make any changes you see fit.  I will merge it tomorrow after confirming that you have no more commits to put in.

hawkular-actions-plugins/hawkular-actions-email/src/main/java/org/hawkular/actions/email/registration/RegistrationInit.java
lucasponce: I was reusing same literal in several places, so I thought a constant can help to understand the meaning of the constant as well.

hawkular-actions-plugins/hawkular-actions-email/src/main/java/org/hawkular/actions/email/template/EmailTemplate.java
lucasponce: Yes. This logic is designed for hawkular instead of a general alert.
Perhaps the best approach could be to have a plugin for hawkular specific scenarios and other plugin for general use cases.
I suggest to merge as it is, as our primary goal is hawkular integration, but within re-packaging tasks I'm going to create a more generic email plugin covering all possibilities (targetting more general use cases).
In any case, the logic is inside the plugin.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/Alert.java
lucasponce: Well, the point is that when an alert is created we have a reference of the triggerId but the trigger can be changed. So, in other words, we can have alerts with no info about trigger details or dampening data that originated (we have the conditions, but we miss this info).
Also, another point to maintain that, is that an action should now the alert and the context that generated in one single call, if after receive an alert, the plugin should call to DefinitionsService to get the Trigger and the Dampening in two calls, the overhead can be even higher.
I agree that the object is getting more fat than expected, perhaps after some feedback from the plugins usecase we can just add the significant info that an action should know to have a good knowledge about context when an alert was generated.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ThresholdRangeConditionEval.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
lucasponce: +1

lucasponce:
Thx for the review, I'm going to send a last commit with a couple of small changes and the PR will be ready to merge.

hawkular-actions-plugins/hawkular-actions-email/src/main/java/org/hawkular/actions/email/registration/RegistrationInit.java
jshaughn: I don't understand this comment :)  I meant do you want to replace the "to" and "cc" literals with the constants you have already defined?

lucasponce: Ops, I thought you were commented the marked lines :-).
Yes, it seems I missed to add the new list of properties supported by the plugin.
Good catch. I will fix it and I will send a new commit with that.
Thx.

jshaughn:
Looks good!  Merging...

PR #65, opened 2015-06-30
HWKALERTS-63 Fix for bus connection not properly closed

lucasponce:


jshaughn:
The explicit close makes sense.  When I was looking at this it seemed like we should be doing the explicit close. Adding this does not exactly explain why the session was being closed but I expect that maybe in the bowels of WF9 they are autoclosing the session after use. Perhaps the bus could be setting an option to control that behavior. It's unclear.

Anyway, this explicit close is good and I'll merge.

PR #67, opened 2015-07-07
HWKALERTS-47 Add Rest query for triggers by tag

jshaughn:
- Add /alerts/triggers/tag rest endpoint
- Add SLSB support for triggers by tag query
- Store "" category instead of nulls
- Fix query SELECT_TAGS_TRIGGERS_ALL_BY_NAME

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
@@ -312,9 +312,10 @@ public class CassStatement {
 
         SELECT_TAGS_TRIGGERS_ALL_BY_NAME = "SELECT tenantId, triggers "
                 + "FROM " + keyspace + ".tags_triggers "
-                + "WHERE name = ? ";
+                + "WHERE name = ? "
+                + "ALLOW FILTERING";
 
-        SELECT_TAGS_TRIGGERS_BY_CATEGORY = "SELECT triggers FROM " + keyspace + ""
+        SELECT_TAGS_TRIGGERS_BY_CATEGORY = "SELECT triggers FROM " + keyspace
                 + ".tags_triggers WHERE tenantId = ? AND category = ? ";
 
         SELECT_TAGS_TRIGGERS_BY_CATEGORY_AND_NAME = "SELECT triggers "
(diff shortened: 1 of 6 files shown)


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
```diff
@@ -312,9 +312,10 @@
 
         SELECT_TAGS_TRIGGERS_ALL_BY_NAME = "SELECT tenantId, triggers "
                 + "FROM " + keyspace + ".tags_triggers "
-                + "WHERE name = ? ";
+                + "WHERE name = ? "
+                + "ALLOW FILTERING";
```
lucasponce: This "ALLOW FILTERING" may lead that something is "forcing" a non performance query.
It seems you need the tenantId.
Instead to force the driver for this query, this it can be done on client side, having a first query with tenants, and running multiple queries concurrently with the tenantId.
It can be interesting to think about it.
In my first version I introduced some ALLOW FILTERING or IN operators, but I could remove them in the 2nd version.

lucasponce:
So, the PR looks OK.
It's not a blocker, but I would try to merge without that ALLOW FILTERING in the queries.
My strategy should be to have a first SELECT_TENANTS query and prepare a concurrent bunch of queries of all tags per tenant basis.
Then we can handle the query per partition basis and be gentle with the C* driver.
But I've said it's not a blocker, if you can do it in the PR perfect, or we can merge and do it later.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassStatement.java
jshaughn: Right, adding tenantId would remove the need for "ALLOW FILTERING" but we don't want the tenantId in the where clause as this is an "ALL" query, specifically for cross-tenant fetch.  It is not expected to be very performant.  You are right that if we had a way to get every tenant we could then run multiple queries in parallel, which may be more performant.   Do you want to change the approach for this PR?

jshaughn:
I will change it if you can show me how to get the tenants, do we have an example?

lucasponce:
We can use triggers table, as in alerts is our first entry point for tenants as well.
Can you try a  "select distinct tenantId from triggers" ?
It should be a fast query.
I have some doubts about the version of the driver we are using, but let me know if it works.

lucasponce:
[cqlsh 5.0.1 | Cassandra 2.1.2-SNAPSHOT | CQL spec 3.2.0 | Native protocol v3]
So the driver version we are using support "select distinct" queries on partition keys.
I think it should work.

lucasponce:
Thx for the fix Jay.
Can you rebase with master to merge it ?

PR #69, opened 2015-07-13
Hwkalerts 37

lucasponce:
- Sending actions on ack and resolved states on alerts.
- Add support into email plugin to deal with multiple states.
- Improve email plugin to support hawkular 1 condition use cases and generic multiconditions alerts.
- Add improvements to templates mechanisms.


jshaughn:
As I've mentioned offline, I'm OK with this change although I think we should keep in mind the alternative of explicit assignment of actions to lifecycle events, as opposed to invoking the same action repeatedly and building more logic into the plugin.  I feel that the explicit assignment may make it easier to manage and understand what will happen at what time, and may allow for more flexibility.  But we can wait for more feedback from the user community.

PR #72, opened 2015-07-16
HWKALERTS-67 Migrate gson dependencies with jackson libraries

lucasponce:
This is the migration from gson to jackson in alerts.
It cannot be tested with -Prest profile as it needs 0.2.1-SNAPSHOT as well as the change in hawkular with the alerts deployment, using hawkular-alerts-rest.war instead of hawkular-alerts-ear.ear package.
I have tested manually with a custom hawkular built, just updating these dependencies.
Apparently, bus and alerts send messages and receive with the new json library Exceptions.
But I didn't perform extensive tests, so not sure if some cornercase was left behind.
I'm now to look into hawkular project as there are some "publisher" that creates alerts message using old gson.
Perhaps some additional fix can be needed as soon as we can test also the hawkular/modules with jackson.
But most of the work is done and I think this it can be ready for reviewing.


jshaughn:
The PR looks good. For completeness you may want to add Data.context and CompareConditionEval.context2 to the equals/hashcode methods for those classes, as you did already for CoditionEval.context.

PR #73, opened 2015-07-16
Upgrade to hawkular-parent 18

ppalaga:


ppalaga:
@jshaughn could you please review?

PR #74, opened 2015-07-21
HWKALERTS-69 Rename hawkular-actions into hawkular-alerts-actions

lucasponce:


jshaughn:
Looks OK.  The artifact name change will need to get updated in Hawkular Dist Build.

PR #75, opened 2015-07-27
HWKALERTS-70 Add ability to delete alerts

jshaughn:
- new endpoints
  - delete: /alerts/<id>
  - put: /alerts/delete <criteria>
  - get: /alerts/alert/<id>


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -264,7 +295,7 @@ public void addAlerts(Collection<Alert> alerts) throws Exception {
                     We have a filter, so we are going to perform several queries with alertsIds filtering
                  */
                 PreparedStatement selectAlertsByTenantAndAlert = CassStatement.get(session,
```
lucasponce: Minimal change. I should change the PreparedStatement to selectAlert to be equals as CassStatement used.
This doesn't required any change.
I leave a note to change it in future reviews.

PR #77, opened 2015-07-29
HAWKULAR-432 Refresh C* session and increase timeout

lucasponce:
With this PR we refresh the cached C* session on Definitions and Alerts services.
Then for slow C* startup it will try to connect in each access avoiding long error logs.
Also number of attempts and timeout was also increased.


pilhuhn:
Just a question : would it make sense in a future iteration to move the "if session=null -> throw exception into the getSession method to unclutter the code a bit?

lucasponce:
Yes it makes sense to change the check inside getSession().
Original logic was to get session at initialization only.
I will change that for next version.
Thx.

PR #78, opened 2015-07-31
HWKALERTS-68

lucasponce:
The main goals of this PR are:
- Refactoring for Actions architecture, no bus dependencies for plugins development.
- Add a flexible way to detect plugins and glue-them with bus or standalone profiles under the hood.
- Create a standalone profile where alerts can be deployed in a clean wildfly server without bus dependencies.
- Adapt tests for bus and standalone profiles.
- Re-org code and artifacts to adapt to bus and non-bus profiles.
- Update rest profile with Alpha3 for local integration tests.


hawkular-alerts-actions-api/src/main/java/org/hawkular/alerts/actions/api/ActionPlugin.java
```diff
@@ -0,0 +1,35 @@
...
+package org.hawkular.alerts.actions.api;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Define an alerts actions plugin implementation
+ * Plugin must have a valid name that will be used at registration phase
```
jshaughn: So, what is a valid name?

lucasponce: Good question.
At the moment there is not any strict validation, the plugin must be just a string.
But I think that in the future we can add some regexp for valid (i.e. not starting with number, special characters, etc).
So, no restriction on this version now.

hawkular-alerts-actions-api/src/main/java/org/hawkular/alerts/actions/api/PluginOpMessage.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * A message sent to the alerts engine from the plugin
+ * It defines a code of operation and payloads
+ *
+ * It used for plugin registration but additional operations should be supported
+ *
+ * @author Lucas Ponce
+ */
+public interface PluginOpMessage {
```
jshaughn: Personal preference, less abbreviation: PluginOperationMessage.  Having said that. "Operation" is a term that may be getting overused.  Perhaps PluginMessageType.

lucasponce: I prefer PluginOperationMessage. PluginMessageType is not meaningful here - I think - as this is an operation message sent by the plugin.
The type is an internal field to define the operation.

hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/EmailTemplate.java
```diff
@@ -268,25 +264,27 @@ private void subjects(Map<String, Object> props, Alert alert) {
...
-            return "not up";
-        } else if (cond.getOperator().equals(AvailabilityCondition.Operator.DOWN)) {
-            return "down";
+        switch (cond.getOperator()) {
+            case UP:
+                return "up";
+            case NOT_UP:
+                return "not up";
+            case DOWN:
+                return "down";
```
jshaughn: I like the change to switch statement but we should probably move these text translations into the enum and just call cond.getOperator.asText() or something like that.

hawkular-alerts-actions-standalone/src/main/java/org/hawkular/alerts/actions/standalone/StandaloneActionPluginListener.java
```diff
@@ -0,0 +1,108 @@
...
+
+import org.hawkular.alerts.actions.api.ActionPluginListener;
+import org.hawkular.alerts.actions.api.PluginMessage;
+import org.hawkular.alerts.api.model.action.Action;
+import org.hawkular.alerts.api.services.ActionListener;
+import org.hawkular.alerts.api.services.DefinitionsService;
+import org.jboss.logging.Logger;
+
+/**
+ * Main bus listener for plugins implementation.
```
jshaughn: I think this class jdoc is wrong.

hawkular-alerts-actions-standalone/src/main/java/org/hawkular/alerts/actions/standalone/StandaloneActionPluginRegister.java
```diff
@@ -0,0 +1,80 @@
...
+
+import javax.naming.InitialContext;
+import javax.naming.NamingException;
+import org.hawkular.alerts.actions.api.ActionPluginListener;
+import org.hawkular.alerts.api.services.ActionListener;
+import org.hawkular.alerts.api.services.ActionsService;
+import org.hawkular.alerts.api.services.DefinitionsService;
+
+/**
+ * Main bus register for plugins implementations
```
jshaughn: And this as well.

hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/EmailTemplate.java
lucasponce: Probably, but the context here is to build a text that is show in the email.
I think that for future the best place is to play with i18n as this text for sure will be localized/translated at some point.

hawkular-alerts-actions-standalone/src/main/java/org/hawkular/alerts/actions/standalone/StandaloneActionPluginListener.java
lucasponce: +1
Generated from bus one and missing the good description.
Good catch.

hawkular-alerts-actions-standalone/src/main/java/org/hawkular/alerts/actions/standalone/StandaloneActionPluginRegister.java
lucasponce: +1
Generated from bus one and missing the good description.
Good catch.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -42,6 +44,7 @@
  * @author Lucas Ponce
  */
 @Singleton
+@TransactionAttribute(value= TransactionAttributeType.NOT_SUPPORTED)
```
jshaughn: Do we want NOT_SUPPORTED on the non Cass*Impl SLSB?

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/AbstractITestBase.groovy
```diff
@@ -46,5 +45,11 @@ class AbstractITestBase {
             .encodeToString("$testUser:$testPasword".getBytes("utf-8"))
          */
         client.defaultRequestHeaders.Authorization = "Basic amRvZTpwYXNzd29yZA=="
+
+        /*
+            This is used for non-bus scenarios.
+            In Bus scenarios this property is skipped
```
jshaughn: skipped or supplied by Hawkular Accounts?

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/LifecycleITest.groovy
```diff
@@ -105,18 +106,16 @@ class LifecycleITest extends AbstractITestBase {
...
+                      ",\"value\":\"DOWN\",\"type\":\"AVAILABILITY\"}]}";
+        resp = client.post(path: "data", body: jsonData);
+        assertEquals(200, resp.status)
 
         // The alert processing happens async, so give it a little time before failing...
         for ( int i=0; i < 10; ++i ) {
-            // println "SLEEP!" ;
-            Thread.sleep(500);
+            println "SLEEP!" ;
+            Thread.sleep(1000);
```
jshaughn: Why id you change this? It will likely just slow down the tests.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
lucasponce: As the C*Impl SLSB has not tx support and the rules engine is not invoked in a different context, I think it has sense to move it as well to NOT_SUPPORTED.
In any case, calling it from a NOT_SUPPORTED just will create a new tx id, but as drools library is not XA, it won't have any significant effect.

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/AbstractITestBase.groovy
lucasponce: Indeed it's skipped.
Hawkular Accounts injects a Persona object and it uses a Hawkular-Persona header.
What hawkular-alert-rest does is to detect if we have accounts or not.
If we have accounts, then it fills the internal tenantId with Persona.getId().
If not present, then it check the Hawkular-Tenant header and fills the tenantId.
It this case, the -rest.war is not coupled with hawkular-accounts and it can be secured with basic, keycloak or other mechanisms.

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/LifecycleITest.groovy
lucasponce: When I executed the mvn clean install -Prest or -Pstandalone profile I noticed that some times (1 time in 9 in a row) tests are failing due some bad result.
I tend to think that in some cases there is some lag somewhere (probably due embedded C*).
Increasing the sleep helps to avoid this issue.

hawkular-alerts-actions-api/src/main/java/org/hawkular/alerts/actions/api/ActionPlugin.java
lucasponce: +1

hawkular-alerts-actions-api/src/main/java/org/hawkular/alerts/actions/api/PluginOpMessage.java
lucasponce: +1

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/LifecycleITest.groovy
jshaughn: increasing the iterations would have done the same but not slowed down the tests.  but honestly, I think the tests should have already had plenty of time.  we should perhaps not supress the prior failures but looks to see if we have perf degradation on repeated runs.  For this PR it is OK, but we may need to investigate.

PR #79, opened 2015-08-03
Order imports to comply with the new Checkstyle rule

ppalaga:


jshaughn:
Do you know the root cause of the ordering issue?  Is there an updated version of the eclipse/IJ formatter? Thes mostly look like file from Lucas' most recent PR, so I wonder if he is out of date in his IDE.

ppalaga:
@jshaughn maybe it is related to https://youtrack.jetbrains.com/issue/IDEA-142468 which is about broken import of the import order prefs to Idea. The ordering itself works properly Idea, if setup manually.

PR #80, opened 2015-08-04
Hwkalerts 74

jshaughn:
Change approach to support dampening for ANY-Match triggers.  The change moves the current condition eval set from working memory (as facts in the rule engine) to the trigger's Dampening record itself.  This actually ended up having some addition benefits other than allowing ANY-Match dampening:
- It simplified the rules and removed the condition limit for ALL-Match triggers (was 4 conditions).
- After repeated runs of the -Pperf engine unit tests I see engine execution speed reduced by ~37%!

After rebasing on master this PR branch had some work done to resolve issues unrelated to the dampening issue that can also be reviewed:
- The File action plugin needed to be adapted to work on windows
- The CassCluster.getSession() method was made synchronized.


lucasponce:
Thanks for the PR.
I liked the way that ConditionEval are treated and it's more clear now.
Also, +1 with perf improvements, the simplification of the Dampening is cool in that way.

PR #85, opened 2015-08-27
Hwkalerts 12

jshaughn:


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/MemberCondition.java
```diff
@@ -0,0 +1,75 @@
...
+        return dataIdMemberMap;
+    }
+
+    /**
+     * The dataIdMap is a map of the dataId tokens in the group condition to the actual dataIds to be used for the
+     * current member triggers. This map will usually have 1 entry but because a condition could have multiple
+     * dataIds (e.g CompareCondition), it may have multiple entries.  The inner map maps member triggerIds to
+     * the dataId to be used for that member trigger for the given token.  It should have 1 entry for each
+     * member trigger.
+     * @param dataIdMap the dataID mappings to be used for the existing member triggers.
```
lucasponce: Javadoc dataIdMap doesnt match with parameter argument.
This may cause a problem in the release phase, so perhaps it's better to fix here.
Description is good but I need to read it twice to understand it, I propose to add a example as well before the @param description, that can help and speed up the use.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/MemberTrigger.java
```diff
@@ -0,0 +1,102 @@
...
+ * @author jay shaughnessy
+ * @author lucas ponce
+ */
+public class MemberTrigger {
+
+    private String groupId;
+    private String memberId;
+    private String memberName;
+    private Map<String, String> memberContext;
+    private Map<String, String> dataIdMap;
```
lucasponce: I guess MemberTrigger.dataIdMap should be propagated at some point in MemberCondition.dataIdMemberMap, right ?
I imagine that there is an example in the rest of the code that I can study, but the different of types with the inner types makes me wonder if I'm understanding correctly the tokens/substitution algorithm.
Basically I understand that the group has a list of tokens and a list of members that need to be substituted at some point.
The inner map in the MemberCondition.dataIdMemberMap should be also in the upper MemberTrigger.dataIdMap ?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
```diff
@@ -28,69 +34,144 @@
  * @author Jay Shaughnessy
  * @author Lucas Ponce
  */
-public class Trigger extends TriggerTemplate {
+public class Trigger {
```
lucasponce: +1
Unifying Trigger and TriggerTemplate is good as it wasn't worked as a real template.
This makes the domain model more clean.
Thanks.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
```diff
@@ -28,69 +34,144 @@
...
+    /** Resolve all unresolved alerts when auto-resolve condition-set is satisfied */
+    @JsonInclude
+    private boolean autoResolveAlerts;
+
+    @JsonInclude
+    private Match autoResolveMatch;
+
+    /** Is a member trigger of the specified group triggerId */
+    @JsonInclude(Include.NON_EMPTY)
+    private String memberOf;
```
lucasponce: So, a null value will identify a "regular" Trigger, a Trigger who is not a member of any Group, and it is not a Group Trigger itself.
I guess that this flag could be incompatible with Trigger.group field, right ?
Then we have three states (Group, Member, and normal that is not group nor member).

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -38,39 +39,108 @@
...
+     * @param tenantId Tenant where trigger is created
+     * @param groupTrigger New trigger definition to be added
+     * @throws Exception If the <code>Trigger</code> already exists.
+     * @see {@link #addTrigger(String, Trigger)} for adding a non-group trigger.
+     */
+    void addGroupTrigger(String tenantId, Trigger groupTrigger) throws Exception;
+
+    /**
+     * Generate a member trigger for the specified group trigger. The dataIdMap replaces the tokens in the
+     * group trigger's conditions with actual dataIds. The member trigger gets the enabled state of the group.
```
lucasponce: As an enhancement, I think that an example of tokens of dataIdMap javadoc can be convenient.
I remember the learning cuve that UI team had first time with the API and a couple of examples can help here.
Nothing complex, just to add a couple of lines of what is expected for dataIdMap a token is also here in the API.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -38,39 +39,108 @@
...
+     * @see {@link #addGroupTrigger(String, Trigger)} for adding a group trigger.
+     */
+    Trigger addMemberTrigger(String tenantId, String groupId, String memberId, String memberName,
+            Map<String, String> memberContext, Map<String, String> dataIdMap) throws Exception;
+
+    /**
      * The <code>Trigger</code> will be removed from the Alerts engine, as needed, and will no longer be persisted.
      * @param tenantId Tenant where trigger is stored
      * @param triggerId Trigger to be removed
+     * @throws NotFoundException if trigger is not found
```
lucasponce: I see that NotFoundException has been introduced.
I have a couple of suggestions:
- If we introduce domain specific Exceptions, those should go into the API, not in the implementation. If some user wants to catch this one but not generic, it needs to explicitly declare implementation classes which should be hidden and only used on dependencies.
- Also, I see that NotFoundException has sense to be in the engine but not in the API as it has an @ApplicationException which is an implementation detail, and this annotation shouldn't go into the API.
- If we introduce a NotFoundException then it has also sense to introduce DuplicateException as the opposite when adding a new Trigger or other entity that exists.
- To make a more friendly use of the API, these domain Exceptions can be extended in the API as vanilla classes, and in the implementation extend a hidden one where the @ApplicationException annotation are set. In that way, we may have annotations in the API without explicit dependencies that are extended in the engine with specific behaviour.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -38,39 +39,108 @@
...
      */
     void removeTrigger(String tenantId, String triggerId) throws Exception;
 
     /**
-     * Update the <code>Trigger</code>. <code>Conditions</code> and <code>Actions</code> are manipulated in separate
-     * calls. The updated <code>Trigger</code> will be persisted.  If enabled the <code>Trigger</code>
-     * will be [re-]inserted into the Alerts engine and any prior dampening will be reset.
+     * The group <code>Trigger</code> will be removed from the Alerts engine, as needed, and will no longer be
+     * persisted. The member triggers will be removed as well, depending on the settings for
+     * <code>leaveMember</code> and <code>leaveOrphans</code>. Note that any member triggers not removed will
```
lucasponce: Javadoc needs to be updated. It maintains leave* for parameters in the signature.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -110,26 +189,110 @@
...
      * @param dataIdMap Tokens to be replaced in the new trigger
-     * @return a copy of original trigger
-     * @throws Exception on any problem
+     * @return the member trigger
+     * @throws NotFoundException if trigger is not found
+     * @throws Exception
+     * @see {@link #orphanMemberTrigger(String, String)} for setting a member to be an orphan.
      */
-    Trigger copyTrigger(String tenantId, String triggerId, Map<String, String> dataIdMap) throws Exception;
+    Trigger unorphanMemberTrigger(String tenantId, String memberId, Map<String, String> memberContext,
```
lucasponce: If the Group was deleted, it has sense to make an "orphan" trigger a normal trigger ?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
```diff
@@ -265,10 +273,10 @@ private void initConditions(File initFolder) throws Exception {
                         log.debugf("Init registration - Inserting [%s]", newCondition);
                     }
                     if (type != null && !type.isEmpty() && type.equals("compare")) {
-                        String dataId = (String)c.get("dataId");
-                        String operator = (String)c.get("operator");
-                        Double data2Multiplier = (Double)c.get("data2Multiplier");
-                        String data2Id = (String)c.get("data2Id");
+                        String dataId = (String) c.get("dataId");
```
lucasponce: I've updated my IJ setting to add this space automatically :-).

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
```diff
@@ -533,6 +698,64 @@ public Trigger updateTrigger(String tenantId, Trigger trigger) throws Exception
...
+    @Override
+    public Trigger orphanMemberTrigger(String tenantId, String memberId) throws Exception {
+        if (isEmpty(tenantId)) {
+            throw new IllegalArgumentException("TenantId must be not null");
+        }
+        if (isEmpty(memberId)) {
+            throw new IllegalArgumentException("TriggerId must be not null");
+        }
+
+        checkTenantId(tenantId, memberId);
```
lucasponce: In this case checkTenantId is not needed. This helper method is used to ensure that Trigger/Dampening/Tag objects have tenantId correctly. When arguments is different of that, it can be omitted.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
```diff
@@ -533,6 +698,64 @@ public Trigger updateTrigger(String tenantId, Trigger trigger) throws Exception
...
+    public Trigger unorphanMemberTrigger(String tenantId, String memberId, Map<String, String> memberContext,
+            Map<String, String> dataIdMap) throws Exception {
+        if (isEmpty(tenantId)) {
+            throw new IllegalArgumentException("TenantId must be not null");
+        }
+        if (isEmpty(memberId)) {
+            throw new IllegalArgumentException("TriggerId must be not null");
+        }
+
+        checkTenantId(tenantId, memberId);
```
lucasponce: Same here.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
```diff
@@ -715,6 +938,40 @@ public Trigger getTrigger(String tenantId, String triggerId) throws Exception {
         }
     }
 
+    // This impl is not efficient. It actually pulls all triggers for the tenant and then filters out the
+    // triggers that are not the requested members.  To make this more efficient we'd likely need to
+    // create a new member_triggers table which duplicated the triggers table, and where we would maintain
+    // a replica entry for each member trigger.  We'd need a primary key like ((tenantId,groupId),memberId) and
+    // then we could query for member triggers directly.  Because this method is expected to be called
+    // rarely, and because the trigger population is not expected to be huge (likely maxing out in the thousands),
+    // we may just get away with this dumb impl.
```
lucasponce: This method should be used rarely, so I'm not fan to do early improvements until, too.
Also, after reading until here, I'm wondering that as group Triggers are not going to be mixed with normal Triggers, perhaps group triggers can go in a different table.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
```diff
@@ -782,81 +1047,155 @@ public Trigger copyTrigger(String tenantId, String triggerId, Map<String, String
...
+                        dataIdMap.get(groupCondition.getDataId()),
+                        ((CompareCondition) groupCondition).getOperator(),
+                        ((CompareCondition) groupCondition).getData2Multiplier(),
+                        dataIdMap.get(((CompareCondition) groupCondition).getData2Id()));
+                break;
+            case EXTERNAL:
+                newCondition = new ExternalCondition(member.getId(), groupCondition.getTriggerMode(),
+                        groupCondition.getConditionSetSize(), groupCondition.getConditionSetIndex(),
+                        dataIdMap.get(groupCondition.getDataId()),
+                        ((ExternalCondition) groupCondition).getSystemId(),
```
lucasponce: Does it have sense to have tokens for ExternalCondition.systemId fields ?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
```diff
@@ -782,81 +1047,155 @@ public Trigger copyTrigger(String tenantId, String triggerId, Map<String, String
...
+                        ((ExternalCondition) groupCondition).getExpression());
```
lucasponce: Also for expressions, too.
As I think Expressions might be related to a specific metric, right ?
So, expanding that metric, might affect a multiple metrics.
I'm trying to imagine if we have a ExternalCondition for avg(ResponseTime) that wants to be applied into multiple machines too.
Perhaps it's a cornercase, but I put this note here in case it can be interesting to thing about this.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/DefinitionsTest.java
```diff
@@ -72,51 +78,538 @@ public void test000InitScheme() throws Exception {
     }
 
     @Test
-    public void test001CopyTrigger() throws Exception {
-        Trigger t = definitionsService.getTrigger(TEST_TENANT, "trigger-1");
-        assertTrue(t != null);
+    public void test0010GroupTrigger() throws Exception {
```
lucasponce: Number on tests were provided when we can't delete alerts so easily, so perhaps as an improvement for future we can validate if tests are isolated and we can stop using number for ordering.
It's just a note to remind it in future.

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/TriggersITest.groovy
```diff
@@ -87,6 +95,216 @@ class TriggersITest extends AbstractITestBase {
...
+
+        resp = client.get(path: "triggers/group-trigger")
+        assertEquals(200, resp.status)
+        groupTrigger = (Trigger)resp.data;
+        assertEquals( true, groupTrigger.isGroup() );
+
+        ThresholdCondition cond1 = new ThresholdCondition("group-trigger", "DataId1-Token",
+            ThresholdCondition.Operator.GT, 10.0);
+        Map<String, Map<String, String>> dataIdMemberMap = new HashMap<>();
+        MemberCondition memberCondition = new MemberCondition( cond1, dataIdMemberMap );
```
lucasponce: Some feedback about name of the class.
With MemberCondition my first thought was this class is member of a group. But its meaning is to define the members for a condition. I wonder if ConditionMembers sounds better or if perhaps was my translation (probably it's me, but just in case).

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/TriggersITest.groovy
```diff
@@ -87,6 +95,216 @@ class TriggersITest extends AbstractITestBase {
...
+
+        resp = client.post(path: "triggers/groups/group-trigger/conditions", body: memberCondition)
+        assertEquals(resp.toString(), 200, resp.status)
+        assertEquals(1, resp.data.size())
+
+        // create member 1
+        Map<String,String> dataIdMap = new HashMap<>(1);
+        dataIdMap.put("DataId1-Token", "DataId1-Child1");
+        MemberTrigger memberTrigger = new MemberTrigger("group-trigger", "member1", "member1", null, dataIdMap);
```
lucasponce: Same here, MemberTrigger sounds to me that is part of a group. But this class is used to map the members of the TriggerGroup, so I also wonders if TriggerMembers could sound more descriptive. I don't know.

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java
```diff
@@ -202,18 +305,107 @@ public Response updateTrigger(
...
+            @ApiResponse(code = 200, message = "Success, Trigger updated"),
+            @ApiResponse(code = 500, message = "Internal server error"),
+            @ApiResponse(code = 404, message = "Trigger doesn't exist/Invalid Parameters") })
+    public Response unorphanMemberTrigger(
+            @ApiParam(value = "Member Trigger id to be made an orphan.", required = true)//
+            @PathParam("memberId")//
+            final String memberId,
+            @ApiParam(required = true, name = "memberTrigger",
+                    value = "Only context and dataIdMap are used when changing back to a non-orphan.")//
+            final MemberTrigger memberTrigger) {
```
lucasponce: If only context and dataIdMap are needed, I tend to a) use just an class that wraps these two arguments, b) use a generic Map and check that context and dataIdMap are needed.
Thinking in non-java REST clients, they would prefer to not load the full Trigger + Map of members, I think.
I didn't check it, but perhaps RESTEasy can force to have a full MemberTrigger deserialized and that can add a big json on client side.
On the other hand, if a client is reusing a json, then this is a minor issue, I think.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/MemberCondition.java
jshaughn: +1. This is the most complicated thing in the entire API.  I will fix/enhance the javadoc now and include it in a commit of PR review changes...

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java
```diff
@@ -514,45 +839,84 @@ public Response createCondition(
             final String triggerId,
             @ApiParam(value = "Json representation of a condition. For examples of Condition types, See "
                     + "https://github.com/hawkular/hawkular-alerts/blob/master/hawkular-alerts-rest-tests/"
-                    + "src/test/groovy/org/hawkular/alerts/rest/ConditionsITest.groovy")
+                    + "src/test/groovy/org/hawkular/alerts/rest/ConditionsITest.groovy")//
             String jsonCondition) {
```
lucasponce: I think this is something I can review in the future. Not part of this PR itself but I guess that with an specific annotation on abstract Condition class, the work of deserialize in inner specific conditions can be done in a transparent way. Instead to do it explicitly here.
It works as it is now, but just a point to improve in the future.

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java
```diff
@@ -514,45 +839,84 @@ public Response createCondition(
...
+            log.debugf(e.getMessage(), e);
+            return ResponseUtil.internalError(e.getMessage());
+        }
+    }
+
+    @POST
+    @Path("/groups/{groupId}/conditions")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Create a new condition for a specific trigger. Add/Delete/Update of a condition "
```
lucasponce: ApiOperation doc should point that this endpoint is for groups.

lucasponce:
Good PR. It's long and I have some comments to review but not any blocker from my side.
I have some topics for future, like to see if groups need specific class and mapping in the backend, but I think is something we can work on future efforts.
Please, Jay, review my comments and let me know if you ant to add some additional commit or if you want to merge it as it is.
Thanks for the work.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/MemberTrigger.java
jshaughn: MemberTrigger.dataIdMap is not the same as MemberCondition.dataIdMemberMap.  I have updated the class-level and method level jdoc in both utility classes.  You can review after the next commit and let me know if it's clear.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
jshaughn: +1,TriggerTemplate was an original guess at what we may need for a template level but it was not needed and this is definitely more clean.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
jshaughn: You are correct:   if 'group' flag is true then 'memberOf' is irrelevant/null and 'orphan' is irrelevant/false.   if memberOf is set then this is a member trigger and 'group' is false.  A member trigger is a trigger evaluated like any other, but is managed via the group.  An orphan is member trigger that is no longer managed by, but is still assigned to the group.  I have added jdoc in Trigger class to help make this clear, on setGroup(), setMemberOf() and setOrphan().

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
jshaughn: I have added javadoc links to the new example in MemberTrigger class.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
jshaughn: I am not opposed to adding an analogous FoundException. These need to be @ApplicationException because we don't want then to be wrapped as EJBException when thrown from the EJB method, and then can be caught easily by the caller (in this case the Rest API code).  Because we now must fetch the Trigger in most, if not all, EJB methods (to perform proper checking for group/member/orphan/standard settings) , it was more efficient and cleaner to introduce the new Exception.

I will see if I can make your recommended changes.  If I have difficulty this can be post PR work.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
jshaughn: good catch.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
jshaughn: It is not possible to have an orphan trigger without a group trigger.  At group trigger removal time the orphan would have already been deleted or made a standard trigger. The same is true of the non-orphan members. See removeGroupTrigger() for more details.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
jshaughn: Great! Because our eclipse formatter setting has that set already.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
jshaughn: +1, I removed it in the two unnecessary places.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
jshaughn: I considered that and decided it was going to cause a lot of conditional code.  Perhaps we will find it is a good idea but it was not a clear enough need to merit the work.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
jshaughn: I don't think this one is necessary, it's more a label.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
jshaughn: Yes, good question. This is a tough one and there are a couple of ways we could go.

If you look at the expression used in the ExternalCondition in our hawkular-metrics integration (See Expression.java) you'll notice that we indeed have metricId in the expression:  "target:interval:func(metricId           op      threshold )[, period]".

But, like all conditions the ExternalCondition does have a dataId. And it's feasible that the dataId could be enough to allow the external alerter code handle things in its own code.  For example, consider a group external condition like this:

    External( dataId: $MetricIdToken$, expression: avg($MetricIdToken$ > 100.0),5 ).

When applied to Member1 it would look like:

    External( dataId: Member1MetricId, expression: avg($MetricIdToken$ > 100.0),5 ).

When processed by the external alerter it would be simple for the external alerter to replace $MetricIdToken$ in the expression with "Member1MetricId".  So, as long as the external alerter had support for groups, it could handle it.

On the other hand, it would also be pretty simple for us to do the same thing. We could apply the same token substitution and replace all occurrences of the same dataId token in the expression.

I'll actually add support for that now, it will probably be enough and will simplify the coding in the external alerter.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/DefinitionsTest.java
jshaughn: To be honest, I'm not even sure these tests are performed in order. I'm not sure the superclass' annotation is getting picked up. But +1 either way.

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/TriggersITest.groovy
jshaughn: I don't like my name, either.  I also don't like your new one :-)    So, I made the following changes, hopefully better as it basically matches the operation:

MemberCondition -> GroupCondition  (the param for TriggersHandler.createGroupCondition())
MemberTrigger -> GroupMember  (the param for TriggersHandler.createGroupMember())

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/TriggersITest.groovy
jshaughn: see above.

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java
jshaughn: Using naming analogous to the newly named DTO's above, added new UnorphanMemberInfo.

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java
jshaughn: At least for now I leveraged JacksonDeserializer.deserializeCondition to re-use code and eliminate the big code block below.

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java
jshaughn: fixed.

jshaughn:
Thanks, excellent review of this large PR!  I have responded to many of your comments with my latest commit.  If it satisfies all of your concerns I think you may be free to merge.

PR #86, opened 2015-09-02
HWKALERTS-83 Add setConditions to REST API

jshaughn:
- add PUT /alerts/triggers/{triggerId}/conditions/{triggerMode}
  - now manage the condition set as a whole. this should simplify managing
    conditions as well as simplify the API.
  - deprecate add/delete/updateCondition endpoints/services
  - remove use of deprecated services in hawkular-alerts code (other than
    to support the deprecated rest services)
- add PUT /alerts/triggers/groups/{groupId}/conditions/{triggerMode}
  - analogous enpoint for group triggers
  - remove add/delete/updateGroupCondition endpoints/services (these can
    be removed as they were never released)
  - rename GroupConditionInfo to GroupConditionsInfo, and update to
    handle a condition set as opposed to a single condition.
- fix bug in addGroupMember, it was not applying group tags.
- increase max-sleep time on LifecycleItests that generate higher number
  of alerts as part of the test.


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/GroupConditionsInfo.java
```diff
@@ -62,24 +82,35 @@ public void setCondition(Condition condition) {
...
      *            [key   - "Member2",
-     *             value = "Member1SystemLoad"]
+     *             value = "Member2SystemLoad"]
+     *           }
+     *  ],
+     *  [key   = "$HeapUsed$",
+     *   value = {[key   = "Member1",
+     *             value = "Member1HeapUsed"],
+     *            [key   - "Member2",
+     *             value = "Member2HeapUsed"]
```
lucasponce: Nice example. Thanks.

lucasponce:
Looks good to me. Merging...

PR #87, opened 2015-09-03
Add context map at Condition level

lucasponce:
I'm going to split the HWKALERTS-85 in several commits.
Then we can consolidate changes on master and work on parallel tasks without being blocked for some commit.
This one brings context to Condition level, needed to support multiple types of conditions at actions plugins level.
The context is optional information added from the data and definition and used (optionally) in the actions plugins.

### DIFF
--- a/hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/pom.xml
+++ b/hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/pom.xml
@@ -39,6 +39,13 @@
       <version>${version.org.freemarker}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.hawkular.alerts</groupId>
+      <artifactId>hawkular-alerts-actions-tests</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+
   </dependencies>
 
   <build>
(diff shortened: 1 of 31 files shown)


lucasponce:
Define triggers and alerts with demo data is costly.
And for actions plugins there are several scenarios that need to be tested.
To simplify the test phase, I've added a shared project under plugins with some demo data will be used in the plugins.

jshaughn:
Are you really sure you want to add tenantId to the constructors/factories for Conditions and Dampening?  I thought we were specifically not doing that because it may be more difficult for clients who maybe don't have the tenantId readily available.  Since Conditions and Dampening are under the Trigger umbrella, it seemed reasonable to just get the information from the Trigger.  I'm not sure this isn't just added complexity in the API.

jshaughn:
Lucas, I will be away until Tuesday, After considering my comments, if you are satisfied with the PR go ahead and commit. I don't want to block you.

lucasponce:
History of the tenantId on Conditions/Dampenings is the following:
- In last HA-12 tenantId was introduced in the Trigger() constructor, so at that point I think we should also add it into the Condition/Dampening, as we don't have a "strong" coupled with them.
- In this particular commit it was introduced as a helper, for me was something needed, but if it creates confusion I can revert it and discuss it properly for MS6. It is not mandatory for the goal of HA-85.
- For me was natural to have it, if we introduced in Trigger() just for consistency.

Well, as we have some time constraints I'm going to rever it with additional feedbacks and we can discuss it for MS6.

Thx for the review, I'm going to add an additional commit with feedback and merge it.

PR #89, opened 2015-09-07
HWKALERTS-85 Improve email plugin to support context

lucasponce:
Improve email plugin to support context on Trigger and Condition attached to alerts.
Previous version was focus on specific type of conditions and it doesn't cover multiple conditions.
With this improvement email plugin can manage better email notifications for multiple scenarios.


jshaughn:
Lucas, this PR looks good.  Address the comments as you see fit and then feel free to merge. -Jay

lucasponce:
Thanks for the review Jay.
I'm going to add an additional commit with feedback comments and merge after that.

PR #95, opened 2015-09-23
Initial introduction of Event into the model

jshaughn:
Based on our design talks so far, Alert extends Event...   This is for the Event branch, and already incorporates HWKALERTS-88 (un-thin Alert.trigger) rand HWKALERTS-89 (simplified Data).


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
```diff
@@ -0,0 +1,225 @@
...
+
+    @JsonInclude
+    protected String id;
+
+    @JsonInclude
+    protected long ctime;
+
+    // A description of the event, suitable for display
+    @JsonInclude
+    private String eventText;
```
lucasponce: Minor comment. Why not simply text ? Event.text looks more readable than Event.eventText.
Also Event.description looked fine but I see Event.text more generic.

lucasponce:
Hi Jay,
I looks good to me.
As I commented by email, for Events branch I would adopt a more agile approach, so in the first iterations we can concentrate in the hawkular-alerts-api.
Updating all API references can be time consuming, so for these steps where we are working on the model, I would leave them without update until we can consolidate a new API.
We can update them one last step.

PR #98, opened 2015-09-23
Add preliminar EventCondition support

lucasponce:
Preliminar work to add EventCondition support.
Key in this design is to support expressions.
Events on conditions are a different type of animal, IMO, as it has not a metric/data source attached.
So I think that expressions could be the key for it.
I started some unit tests but not completed.
Feel free to play with it or even modified if necessary.


jshaughn:
I didn't have enough time to look at this in depth but I'm not sure we're thinking the same way [yet]. You are going in the direction of an uber-condition and I was thinking about something very simple.  I'm too distracted at the moment to think, I'll try to come back later or we'll discuss more tomorrow.

lucasponce:
I think that with the introduction of Events on conditions arena, we have now two main blocks of conditions: for data and for events, each one has its own details of processing.
And about the uber-condition, this is just a first idea, but I think that it makes sense to have expressions to define conditions in a easier way, more natural thinking in client-area.
But anyway, it's a first point to talk about it.
I will update a couple of unit tests that I couldn't finish yesterday.

lucasponce:
Added name into EventCondition

jshaughn:
I have discussed with Lucas and we have agreed to merge the PR.  As the merge is only into a branch things are still subject to change.  But we are trying for relatively small PRs...

jshaughn:
Lucas, I've closed this PR.  I rebased your PR on upstream/Events and resolved conflicts.  I could not push to your repo so I created a new PR from my own repo, and merged from there.  So basically, your code is now in the upstream/Events branch.

PR #100, opened 2015-09-23
Playing with new Event impact on Trigger (API module mainly)

jshaughn:
This does not need to be merged unless you think it's fine.  It's here more for review and any comments...

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
@@ -16,13 +16,16 @@
  */
 package org.hawkular.alerts.api.model.event;
 
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 import org.hawkular.alerts.api.model.condition.ConditionEval;
 import org.hawkular.alerts.api.model.dampening.Dampening;
+import org.hawkular.alerts.api.model.trigger.Tag;
 import org.hawkular.alerts.api.model.trigger.Trigger;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
@@ -48,13 +51,16 @@ public class Event {
 
     // A description of the event, suitable for display
     @JsonInclude
-    private String eventText;
+    private String text;
 
     @JsonInclude(Include.NON_EMPTY)
     private Map<String, String> context;
 
+    // TODO: WE NEED TO POTENTIALLY CONSOLIDATE OUR TAG APPROACH.  I THINK MAYBE WE SHOULD MOVE THE CURRENT Tag TO
+    // A NAME-VALUE PAIR, GET RID OF HIDDEN, AND CHANGE triggerId to a more generic ID.  The Tag as is is not
+    // going to work here, but for now leave as is/.
     @JsonInclude(Include.NON_EMPTY)
-    private Map<String, String> tags;
+    private Set<Tag> tags;
 
     // Null for API-generated Events. Otherwise the Trigger that created the event (@ctime)
     @JsonInclude(Include.NON_EMPTY)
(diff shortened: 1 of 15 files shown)


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
```diff
@@ -48,13 +51,16 @@
...
 
     @JsonInclude(Include.NON_EMPTY)
     private Map<String, String> context;
 
+    // TODO: WE NEED TO POTENTIALLY CONSOLIDATE OUR TAG APPROACH.  I THINK MAYBE WE SHOULD MOVE THE CURRENT Tag TO
+    // A NAME-VALUE PAIR, GET RID OF HIDDEN, AND CHANGE triggerId to a more generic ID.  The Tag as is is not
+    // going to work here, but for now leave as is/.
     @JsonInclude(Include.NON_EMPTY)
-    private Map<String, String> tags;
+    private Set<Tag> tags;
```
lucasponce: I am agree with the comment.
Perhaps the Tag model works fine but Trigger but I think we should study a lightweight approach for Events, or even for Triggers too.
I must say I liked the Map.tags, it looked intuitive coming from a REST approach.
I don't know perhaps we can talk about tags also to see which effort could be more interesting to consolidate.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/EventType.java
```diff
@@ -0,0 +1,27 @@
...
+package org.hawkular.alerts.api.model.event;
+
+/**
+ * What the trigger produces.
+ *
+ * @author jay shaughnessy
+ * @author lucas ponce
+ */
+public enum EventType {
+    ALERT, EVENT
```
lucasponce: Yesterday I was still thinking on the Heritage vs Composition approaches to represent the relationship between Alert and Event.
From different angles I see both approaches correct which is hard because we need to choose one and it's unclear which is better.
Here some of my arguments we can talk later:
- From natural language I see that a event/happening might need an Alert. But there are nuances that make me think that Event is the object and the Alert is the subject who operates with, so in that sense the composition pattern maps better.
- Also, to update an Alert shouldn't mean that an Event is being modified (as a happening has a time).
- As per requeriments, we might introduce new things on Alert (as comments, history of transitions, and others), I see that heritage is perhaps to overload the Event concept.
Anyway, I think this is better to comment in a talk as at the end we are talking about details/nuances.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -141,6 +142,24 @@ public void addAlerts(Collection<Alert> alerts) throws Exception {
             msgLog.errorDatabaseException(e.getMessage());
             throw e;
         }
+
```
lucasponce: I wasn't thinking on implementation yet, as it may vary from the API

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -293,16 +296,31 @@ rule AlertOnSatisfiedDampening
         $t  : Trigger( mode == Mode.FIRING, $tid : id )
         $d  : Dampening( triggerMode == Mode.FIRING, triggerId == $tid, satisfied == true )
     then
-        if (log != null && log.isDebugEnabled()) {
-            log.debug("AlertOnSatisfiedDampening! " + $d.log());
+        Event newEvent;
```
lucasponce: Same of my previous comment, as we are still having some doubts in the api, I'm trying to not start the implementation yet.
Just to isolate, and not to bring yet ergonomics from implementation than might influence in the model.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/EventType.java
jshaughn: I agree, it's not a clear win either way.  Even completely separate classes could be considered.  I think there are two  main advantages to inheritance over composition.  First is with respect to Actions.  Both Events and Alerts can perform actions and in this fashion the Actions can operate on an Event.  The second is maybe more subtle, because an Alert will have a related, but independent Event, I don't see that an Alert is actually composed of an Event, there is no Event that is actually making up the Alert, instead there is an Event representing the alert.  (I apologize that this is in English, it is subtle already.  Hopefully I am clear).

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
jshaughn: It helps me to think about the issues to follow the impl impact to some degree.  I think it is just style of work.

jshaughn:
I have discussed with Lucas and we have agreed to merge the PR.  As the merge is only into a branch things are still subject to change.  But we are trying for relatively small PRs...

PR #101, opened 2015-09-24
Upgrade to hawkular-parent 24

ppalaga:


ppalaga:
@jshaughn or @lucasponce  would you please review an eventually merge?

PR #102, opened 2015-09-28
HWKALERTS-91 Change Tag impl to Map of name-value pairs and store on …

jshaughn:
…Trigger

- This initial impl still uses Trigger's Tags for Alerts fetch.  That
  will change as the Events work is completed.
- Simplify Tag persistence into single table and improve distribution


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsCriteria.java
```diff
@@ -122,28 +122,21 @@ public void setTriggerIds(Collection<String> triggerIds) {
...
 
-    /**
-     * @param tags fetched Alerts must be for trigger with any of the specified Tags.
-     */
-    public void setTags(Collection<Tag> tags) {
+    public void setTags(Map<String, String> tags) {
         this.tags = tags;
     }
 
+    public void addTag(String tname, String tvalue) {
```
lucasponce: Minor comment: what about tagName, tagValue or simply name of value ? Just for consistency with the naming, in whole project we are using long variable names as tenantId os similar, I vote to maintain it. Or simply use name/value as with method is clear that we are talking about tags.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/TagType.java
```diff
@@ -0,0 +1,27 @@
...
+package org.hawkular.alerts.api.model;
+
+/**
+ * The type of Tag being stored or fethed.
+ *
+ * @author jay shaughnessy
+ * @author lucas ponce
+ */
+public enum TagType {
+    TRIGGER, ALERT, EVENT
```
lucasponce: Perhaps this enum is part of the implementation instead of the model ?
From a pure API angle, tags are part of a component, isn't it ?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
```diff
@@ -1909,11 +1928,15 @@ private Dampening mapDampening(Row row) {
                     throw new IllegalArgumentException("Unexpected ConditionType: " + cond);
                 }
 
+                // TODO: ARE THESE TAGS STILL NEEDED BY THE HAWKULAR UI? THIS IS CLUNKY.
+                // I DON'T THINK THEY ARE USED. REMOVING...
```
lucasponce: Today UI is fetching triggers/alerts mainly by triggerId, which is something composed with specific strings.
So, I'm not 100% sure but tags are not used in UI code.

lucasponce:
PR looks good. Nothing blocker.
I put a couple of comments in case you want to review them.
I see the TagType as something depending on the implementation, but not related to a pure API interaction between the client and the component (I think).
The other one is a minor thing about naming.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsCriteria.java
jshaughn: OK, I took that from what they did in Metrics but I agree, it's not consistent with our stuff.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/TagType.java
jshaughn: Yes, I don't think this needs to be exposed to the API where we can have dedicated services for each type.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
jshaughn: Thanks for confirming. I looked in the UI code and could not find any code using these built-in tags, which were originally added because the UI needed it.  But I think it's no longer the case and I want to remove them.

PR #104, opened 2015-09-30
Jay events

jshaughn:
This branch now builds and all tests actually pass.  Of course, there are lots of missing Event services and also not a lot of Event-specific tests.  But it does at least carry Events through the model such that
Alert subclasses Event, and an Event is being persisted for each Alert.    The tag work for using name-value pairs is pretty much complete.   And now Triggers and Alerts persist their tags independently
(as will Events).  Alert fetch via tag uses the explicit alert tags, not implicitly going through triggers.  Actions now take an Event as opposed to an Alert, and I added minor support for handling each.


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
```diff
@@ -35,6 +39,12 @@
  * @author Jay Shaughnessy
  * @author Lucas Ponce
  */
+@JsonTypeInfo(
```
lucasponce: Nice annotation.
Note: This comment is not related to this PR but we have a custom deserializer that could be substituted by this annotation. This is a change with low priority and it's not a problem, but at some point (a spring clean or similar kind of task, it's something we can do).

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
```diff
@@ -46,6 +56,10 @@
     @JsonInclude
     protected long ctime;
 
+    // category of Event, suitable for display, recommended to be, but not limited to, an EventCategory.name.
+    @JsonInclude
+    private String category;
```
lucasponce: +1
Having an string can give it flexibility to use own "types".
An enum might be difficult to maintain.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/EventCategory.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Every Event has a category to help identify what it represents.  It can be one of these, or it can be something
+ * specified by the Event creator.  This just tries to predefine common events.
+ *
+ * @author jay shaughnessy
+ * @author lucas ponce
+ */
+public enum EventCategory {
+    ALERT, DEPLOYMENT, LOG, TRIGGER
```
lucasponce: I can identify ALERT, DEPLOYMENT and LOG events. But not sure if I get the semantic of a TRIGGER Event.
Is intended for those Events created internally from a TRIGGER ? right ?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -488,34 +565,30 @@ private boolean filterByStatuses(String tenantId, Set<String> alertsId, AlertsCr
...
-                    alertsId.add(alertId);
+                    alertIds.add(alertId);
                 }
             });
-            /*
-                If there is not alertId but we have triggersId means that we have an empty result.
-                So we need to sure a alertId to mark that we have an empty result for future joins.
-             */
-            if (alertsId.isEmpty()) {
-                alertsId.add("no-result-fake-alert-id");
```
lucasponce: I didn't review with low detail the implementation changes, but this one came to my attention.
The "no-result-fake-alert-id" was needed here in the join implemented on client.
Is this being modified ?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/EventCategory.java
jshaughn: Good eyes :)  That is what it means but honestly it is pretty weak because TRIGGER does not really explain what the event is truly about, only the source.  But the alternative would be to add yet another field to Trigger, used only on Event triggers, to specify a category.  I suppose one possibility would be to take the current Trigger.eventText field and instead make it Trigger.category.  And then, set the Event text explicitly to the Trigger name.  That may not be bad at all. Let me know...

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
jshaughn: I couldn't figure out what this was for, despite the comment.  I need to know better what this is before adding anything back because it looks "hacky".

jshaughn:
This PR has had significant work since your last round of comments.  I think the feature is now more robust and the impl is better in several ways.  I suggest we merge this to the upstream Events branch, and then not delay too long before moving to master.

PR #107, opened 2015-10-04
HWKALERTS-78 Notification History API

lucasponce:
The goals of this PR are the following:
- Store actions in the backend.
- Fetch or delete actions based on criteria.
- Add support into the plugins architecture to send operation messages from the plugins to the engine.
- Support hawkular and standalone architectures.
- Store actions definitions with the action to track possible changes on actions definitions.
- Process actions in separate threads in standalone deployments.


hawkular-alerts-actions-bus/src/main/java/org/hawkular/alerts/actions/bus/BusActionPluginSender.java
```diff
@@ -0,0 +1,139 @@
...
+    private ProducerConnectionContext pcc;
+
+    public BusActionPluginSender(String actionPlugin) {
+        this.actionPlugin = actionPlugin;
+    }
+
+    private void init() throws Exception {
+        /*
+            Init and cache factories and contexts used.
+            ActiveMQ recommends to catch MessageProducer and MessageConsumer objects that are wrapped in context:
```
jshaughn: catch -> cache

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/Action.java
```diff
@@ -27,7 +29,7 @@
  * A Trigger definition can be linked with a list of actions.
  *
  * Alert engine only needs to know an action id and message/payload.
- * Action payload can optionally have an alert as payload.
+ * Action payload must have an alert as payload.
```
jshaughn: Comment: I expect this will be an Event in the near future.

hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/listener/ActionPluginOperationListener.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * A component that listens from the bus operation messages coming from plugins.
+ *
+ * @author Jay Shaughnessy
+ * @author Lucas Ponce
+ */
+@MessageDriven(messageListenerInterface = MessageListener.class, activationConfig = {
+        @ActivationConfigProperty(propertyName = "destinationType", propertyValue = "javax.jms.Queue"),
+        @ActivationConfigProperty(propertyName = "destination", propertyValue = "HawkularAlertsOperationsQueue")})
```
jshaughn: Part of me wonders why we have both "Action" terminology and "Operation" terminology.  Why is a plugin not listening for Actions to perform?  I feel like we're going to confuse people using the word operation and I'm wondering whether it can be eliminated (not a PR blocker).

hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/EmailPlugin.java
```diff
@@ -223,18 +235,46 @@ private void initMailSession() {
         return defaultProperties;
     }
 
-    @Override
-    public void process(PluginMessage msg) throws Exception {
-        Message message = createMimeMessage(msg);
-        Transport.send(message);
+    private void sendResult(Action action) {
```
jshaughn: Is the email plugin the only plugin currently persisting results?  Or did I miss something in the other plugins?

jshaughn:
This is a large amount of work so it's possible I missed some things but it looks good to me. No blockers.  There will need to be some immediate refactoring for Events but that is likely expected.

hawkular-alerts-actions-bus/src/main/java/org/hawkular/alerts/actions/bus/BusActionPluginSender.java
lucasponce: +1
Brain typo :-)

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/Action.java
lucasponce: Yes, as this work was done in the master branch with no notion of Event, I didn't want to mix concepts.
We will refactor once we can apply the Events branch.

hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/listener/ActionPluginOperationListener.java
lucasponce: The main design is that we are going to have a plugin developers.
They have to interact only with the org.hawkular.alerts.actions.api.
In that package the @Plugin defines a plugin implementation, this class must implement ActionPluginListener.process(ActionMessage action), which is the main method that "listens" Actions.

The "glue" code is completely hidden, a plugin developer can write its code without any worry if his code should be a JMS MDB or other implementation, as we have a hawkular deployment and a standalone deployment.

This is intended to hide all bus dependencies on the plugin side, and also to simplify the design.

This MDB is part of the implementation used in hawkular, but the standalone has other with no JMS involved.

Ok, main feature of the plugin is to listen Actions and process them. But now, we offer bidirectional communication, so a plugin can send messages to the engine.

I didn't choose to send an Action, because in essence the plugin should not modify an action, apart of the result, so sending the whole action back I think it was too heavy (indeed at some point we could review where we can lean payloads, but that's another fight for future).

So, before that plugins can send explicitly messages to the engine, we used a first one, called operation, to register a plugin into the engine (without it, the engine has not visibility of plugins deployed in an installation). My first approach was to reuse the concept of OperationMessage as a generic way that can be extended. I thought also in PluginMessage or ResultMessage as an alternatives.

In summary, we have ActionMessages that are coming from engine to the plugins, and OperationMessages that are sending from the plugins to the engine.

Writting this I think that perhaps a PluginMessage might be more meaningful as name.

I can change it later.

hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/EmailPlugin.java
lucasponce: I think yes, I missed to add a simple return.
I will fix it on master.

lucasponce:
Thx for the review Jay.
Yes, this work will need inmediate refactoring for Events, and also more things, as plugins should be adapted with conditional logic to cover Events and/or Alerts (for example, the messages/templates might be completely different if we are presenting raw Events or Alerts, as today the email templates are quite dependent on the lifecycle status).

hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/listener/ActionPluginOperationListener.java
jshaughn: perhaps ActionResponseMessage

PR #118, opened 2015-11-03
EventsConditions

lucasponce:
Main goals for this PR:
- EventConditions working at engine level.
- Events re-feed on the engine to define EventConditions on Events generated on other conditions.
- Filtering of events. Send only events with dataId active.
- Some minor fixes to make the rest profile working on linux with hawkular Alpha6 (as development tool).


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
```diff
@@ -31,48 +31,63 @@
 public class EventCondition extends Condition {
```
jshaughn: I would like to see more class-level and maybe field level jdoc on how this fairly complex condition is used.  The processExpression jdoc is good but that is a private method.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
```diff
@@ -100,6 +118,7 @@ public boolean match(Event value) {
     private static final String ID = "id";
     private static final String CTIME = "ctime";
     private static final String TEXT = "text";
+    private static final String CATEGORY = "category";
     private static final String CONTEXT = "context.";
```
jshaughn: I'm not sure I like that you can perform matching against context, this sort of breaks our general rule of not performing matching/filtering against data context and may lead to requests for other data to be able to do it and the other types are not set up in a fashion to provide that sort of feature.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
```diff
@@ -84,9 +99,12 @@ public void setExpression(String expression) {
...
 
     public boolean match(Event value) {
-        if (null == expression || expression.isEmpty() || null == value) {
+        if (null == value) {
             return false;
         }
+        if (null == expression || expression.isEmpty()) {
+            return true;
+        }
         String[] expressions = expression.split(",");
```
jshaughn: It's not clear from any jdoc that the "expression" field can actually contain a comma-separated list of individual expression items.  Also, a comma separator is a bit dangerous given all of the string manipulations/operations that are allowed here.  I could easily see a problem occurring when a comma shows up in a string used for a value or pattern.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
```diff
@@ -64,6 +64,10 @@
     @JsonInclude
     protected long ctime;
 
+    // Source of Event, used by EventCondition to evaluate events
+    @JsonInclude
+    private String dataId;
```
jshaughn: It's optional, should this have (Include.NON_EMPTY) ?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
```diff
@@ -98,23 +102,50 @@ public Event() {
...
+        this(tenantId, id, ctime, dataId, category, text, context, null);
     }
 
-    public Event(String tenantId, String id, String category, String text, Map<String, String> context,
-            Map<String, String> tags) {
+    public Event(String tenantId, String id, long ctime, String dataId, String category, String text,
+                 Map<String, String> context, Map<String, String> tags) {
         this.tenantId = tenantId;
         this.id = id;
+        this.ctime = ctime;
```
jshaughn: I think we should treat this like we do in Data, and assign currentTime if ctime <= 0.  Don't let bogus ctimes be passed in.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
```diff
@@ -171,4 +171,24 @@ void resolveAlertsForTrigger(String tenantId, String triggerId, String resolvedB
...
+
+    /**
+     * Send event into the alerting system for evaluation.
+     * Events are persisted after inference.
+     * addEvents() only persist events but not put it on the engine for evaluation.
+     *
+     * @param event Not Null. The events to be avaluated and persisted by the alerting engine.
+     * @throws Exception any problem
+     */
+    void sendEvent(Collection<Event> event) throws Exception;
```
jshaughn: probably should be sendEvents().  Unlike "Data" which is fine for singular or plural, Event is only singular.

hawkular-alerts-engine/pom.xml
```diff
@@ -227,12 +227,14 @@
...
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
+              <!--
               <systemProperties>
                 <property>
                   <name>testOutput</name>
                   <value>${project.build.testOutputDirectory}</value>
                 </property>
               </systemProperties>
+              -->
```
jshaughn: Either add a comment as to why this is here or delete it.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -68,6 +68,7 @@
     private int period;
 
     private final List<Data> pendingData;
+    private final List<Event> pendingEvent;
```
jshaughn: Again, probably better as a plural: pendingEvents

jshaughn: I think the clear() method should add:
pendingEvents.clear();

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -317,30 +335,61 @@ private synchronized void addPendingData(Data data) {
...
+        pendingEvent.add(event);
+    }
+
     private synchronized Collection<Data> getAndClearPendingData() {
         Collection<Data> result = new ArrayList<>(pendingData);
         pendingData.clear();
         return result;
     }
 
+    private synchronized Collection<Event> getAndClearPendingEvent() {
```
jshaughn: same again, plural...

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -317,30 +335,61 @@ private synchronized void addPendingData(Data data) {
...
+
     private class RulesInvoker extends TimerTask {
         @Override
         public void run() {
             int numTimeouts = checkPendingTimeouts();
 
-            if (!pendingData.isEmpty() || numTimeouts > 0) {
+            if (!pendingData.isEmpty() || !pendingEvent.isEmpty() || numTimeouts > 0) {
                 Collection<Data> newData = getAndClearPendingData();
+                Collection<Event> newEvent = getAndClearPendingEvent();
```
jshaughn: sorry, again plural...

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -317,30 +335,61 @@ private synchronized void addPendingData(Data data) {
...
+                if (!newEvent.isEmpty()) {
+                    /*
+                        Incoming Events are added to the result events list.
+                        Events are all persisted at the end of the inference.
+                        With this we will have on events list:
+                        - Input events.
+                        - New events generates as result of the inference process.
+                     */
+                    events.addAll(newEvent);
+                }
```
jshaughn: I'm not sure I understand this.  Aren't API-inserted Events persisted independently of being sent into the engine?  I don't think we should be worrying about persisting API-inserted event here in this code, only Events emitted by the engine.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -1293,6 +1296,43 @@ public void sendData(Collection<Data> data) throws Exception {
...
     }
 
+    @Override
+    public void sendEvent(Event event) throws Exception {
+        if (null == event) {
+            return;
+        }
+        if (event.getDataId() == null
+                || event.getDataId().isEmpty()
+                || !eventsCacheManager.getActiveDataIds().contains(event.getDataId())) {
```
jshaughn: may want to use  isEmpty(event.getDataId())

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -1293,6 +1296,43 @@ public void sendData(Collection<Data> data) throws Exception {
...
+    @Override
+    public void sendEvent(Collection<Event> event) throws Exception {
+        if (null == event || event.isEmpty()) {
+            return;
+        }
+        Set<String> activeDataIds = eventsCacheManager.getActiveDataIds();
+        Collection<Event> withoutConditions = new ArrayList<>();
+        Collection<Event> withConditions = new ArrayList<>();
+        for (Event e : event) {
+            if (e.getDataId() == null || e.getDataId().isEmpty() || !activeDataIds.contains(e.getDataId())) {
```
jshaughn: isEmpty(e.getDataId())

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -55,6 +56,7 @@
     private KieSession kSession;
 
     TreeSet<Data> pendingData = new TreeSet<>();
+    TreeSet<Event> pendingEvent = new TreeSet<>();
```
jshaughn: plural

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -147,15 +160,33 @@ public void fire() {
...
+                    log.debugf("Deferring more recent %1$s until older %2$s is processed", event, previousEvent);
+                }
+            }
+
+            if (log.isDebugEnabled()) {
+                log.debugf("Firing cycle [%s] - with these facts: ", fireCycle);
                 for (FactHandle fact : kSession.getFactHandles()) {
                     Object o = kSession.getObject(fact);
-                    log.trace("Fact: " + o.toString());
+                    log.debugf("Fact:  %s", o.toString());
```
jshaughn: did you mean tracef?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -147,15 +160,33 @@ public void fire() {
...
                 }
             }
 
             kSession.fireAllRules();
+            fireCycle++;
```
jshaughn: MAXINT issue :)

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/rules/RulesEngine.java
```diff
@@ -100,6 +101,20 @@
...
+     * @param event the event
+     */
+    void addEvent(Event event);
+
+    /**
+     * Add to the accumulated <code>Event</code> to be processed the next time {@link #fire()} is called. After the
+     * rules are fired on the accumulated <code>Event</code> it will be cleared.
+     * @param events the events
+     */
+    void addEvent(Collection<Event> events);
```
jshaughn: plural

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/service/AlertsEngine.java
```diff
@@ -57,6 +58,25 @@
...
+    void sendEvent(Event event) throws Exception;
+
+    /**
+     * Send event into the alerting system for evaluation.
+     * Events are persisted after inference.
+     *
+     * @param event Not Null. The events to be avaluated and persisted by the alerting engine.
+     * @throws Exception any problem
+     */
+    void sendEvent(Collection<Event> event) throws Exception;
```
jshaughn: plural

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/EventsHandler.java
```diff
@@ -97,7 +96,11 @@ public Response createEvent(
                 if (null != alertsService.getEvent(tenantId, event.getId(), true)) {
                     return ResponseUtil.badRequest("Event with ID [" + event.getId() + "] exists.");
                 }
-                alertsService.addEvents(Collections.singletonList(event));
+                /*
+                    New events are sent directly to the engine for inference process.
+                    Input events and new ones generated by the alerts engine are persisted at the end of the process.
+                 */
+                alertsService.sendEvent(event);
```
jshaughn: as mentioned we should maybe discuss this

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
lucasponce: +1

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
lucasponce: This is a mixed feeling to me.
As we have two types of inference here:
- Against data, where the data is explicit.
- Against Events, where we have only a text, but potentially possibility to "draw" a complex type using the context.
So, for events, I understand that context could be part of the "first class" data where an user is interested on inference.
I don't know which could be better, I vote to leave it as it seems events have its own scenario.
Perhaps we can talk more about this.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
lucasponce: +1 to better documentation
String manipulations operations are separated by simple quotes.
A complex expression for a type was something interesting as in the same event I would like to have some predicates for the expressions without moving that to several conditions for the same data.
Expressions brought to my mind as the simple solution here, I think.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
lucasponce: +1

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Event.java
lucasponce: +1 good catch

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
lucasponce: +1
Sorry, trying to replicate the sendData() signatures I introduced this english language typo (and I can see it in several places).
Thanks for commenting it.

hawkular-alerts-engine/pom.xml
lucasponce: +1
Thanks, this was leave by me while debugging an strange error on the rest profile, I should clean it.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
+1
Good catch

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
lucasponce: Original logic was:
- Events that are sent to the engine are persisted by the engine with the emitted one at the end of the inference.
- Events that are not sent to the engine are directly persisted.

Your comment could be to persist first and only let the engine to persist new events.
It could be valid, I don't think it change the algorithm, only the place where it is executed, so I'm +1 with the change if it is more clear.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
lucasponce: I've moved to debug as I missed it while debugging some scenario.
I guess it is useful to have it on debug as we can have all the info to debug external executions (thinking on some reported error or so), having this info in trace forces to increase a lot the logging output for debugging and I've spent a lot of time just filtering/cutting/chopping the log files to get it.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
lucasponce: Perhaps a long can help ?
It's mainly for debugging and check in a high fire() invocation how many cycles executes.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/rules/RulesEngine.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/service/AlertsEngine.java
lucasponce: +1

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/EventsHandler.java
lucasponce: Yes.
We have logic to persist events (without sending to the engine) and send events (that checks active dataId and sends to the engine if necessary).
For me better approach could be to have only 1 method with the logic that sendEvent does, and make the addEvents private.

lucasponce: Or renaming/accomodate wording but in essence what does sendEvent() is what it should be public.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
jshaughn: I agree that Events are not standard Data, but Context is available on Data, and on Triggers etc, and is never available for filtering.  I think we will send a strange, inconsistent message if we make it available here, just because we can.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
jshaughn: OK, no problem.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
jshaughn: I was joking, I don't expect it to be a problem, it's only when running with debug and I don't expect billions of cycles in one run.  But, you can make it a long and we can be very safe :)

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
jshaughn: I'm not sure I understand your response.  I don't have a problem with complex expressions.  I only meant that it was confusing that "expression" could contain multiple expressions. The terminology seems backwards and it wasn't clear from the jdoc,

Moreover I was concerned about the comma-separated parsing.  But maybe I am not reading it correctly.  In short, will this parse correctly:

event.id start 'ID,YZ', event.context.category == 'Ser,ver', event.tag.from end ',com'

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
lucasponce: I have a draft of a long response but at the end I think that this issue could have a better solution just limiting the inference against the tag map and as you suggest to not use context for it.
With proper documentation we can offer this and being uniform with Data and same use of the context map in other scenarios.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
lucasponce: Debugging is an ancient art :-) I thought you were serious. I'll leave it as int.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/EventCondition.java
lucasponce: event.id start 'ID,YZ', event.context.category == 'Ser,ver', event.tag.from end ',com'
will not be parsed correctly, as the split will break in incorrect tokens.
I'll update this with a better parsing algorithm covering escaped commas and quote inside the string argument.

lucasponce:
I have submitted a couple of commits with most of the comments applied.

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/EventsHandler.java
jshaughn: I agree that we should only have one service.  The client should not have to choose, all Events should be persisted and some may go to the engine for evaluation.  It probably doesn't make much difference but I'd probably keep addEvents() as opposed to sendEvents(), because (to me) the primary function is to persist events, and sending them to the engine is something of a side-effect.

The service and engine impls now look good to me with the changes you've made.

lucasponce: +1
Yeah, I'm going to unify all logic under addEvent() and make sendEvent() private to the implementation.
It has sense.

lucasponce:
Final fixes for this PR.
The optional endpoint to add multiple Events is not part of this PR, I think I will add it later once this is consolidated.

jshaughn:
Good Job, merging...

PR #119, opened 2015-11-04
Sync master with Events branch

lucasponce:
This is a copy of master with Events branch merged.
I have run locally default, standalone, perf and rest profiles.
Everything running as expected locally against Hawkular Alpha6 as integration platform for testing.


jshaughn:
yay for us!

PR #125, opened 2015-11-23
HWKALERTS-104 Add support for EAP64 on standalone profile

lucasponce:
The goal of this PR is to make compatible hawkular-alerts standalone with EAP64 and Wildfly9.
EAP64 is used for OpenShift deployments (a migration to WF9/WF10 is on the way, so this compatibility should be limited on time).
Inside of alerts there are minor changes to make them both compatible.
Including a third party libraries on the standalone and change some logging debug methods.
I ran some end-to-end tests against hawkular-examples and no issues showed up.
There is one known issue, on EAP64, urls with character '|' should be encoded.
So, when using '|' as a separator on a URL, this should be encoded by "%7C".


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -214,11 +214,11 @@ public void reloadTrigger(final String tenantId, final String triggerId) {
         try {
             trigger = definitions.getTrigger(tenantId, triggerId);
         } catch (Exception e) {
-            log.debugf(e.getMessage(), e);
+            log.debug(e.getMessage(), e);
             msgLog.errorDefinitionsService("Trigger", e.getMessage());
         }
         if (null == trigger) {
-            log.debugf("Trigger not found for triggerId [" + triggerId + "], removing from rulebase if it exists");
+            log.debug("Trigger not found for triggerId [" + triggerId + "], removing from rulebase if it exists");
```
jshaughn: we should use if ( log.debugEnabled() ) { log.debug(...); } whenthe debug logging is not of a literal string and not in  exception handling.  So, in several places in these changes.

As it turns out, this was already a poorly written log.debugf(), as it used string concat and not variable args...

jshaughn:
The loss of log.debugf() is annoying.  And the need for beans.xml is also annoying (is that to discover beans to be injected?).  We should wrap the debug logging where it makes sense.  As long as the dep changes don't adversely affect anything when running on newer WFly or EAP then this otherwise looks fine.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
Yeap, log.debugEnabled() is a good suggestion here, I'm going to review it to avoid unnecessary string manipulation.

lucasponce:
Thx for the review.
As I rebased this PR on the Clustering branch, I'm going to review these changes on the Clustering branch.
Also I need to check that Clustering works fine on Wildfly and EAP64.
So, I'm going to close this one and apply your suggestions in the main Clustering branch to avoid force rebasing.

PR #126, opened 2015-11-29
Clustering

lucasponce:
WIP, not to merge but start review.
It implements main scalability design discussed on F2F.
It is not finished but main design is mostly implemented.

Main goal is to partition triggers on AlertsEngine in a explicit way having fine granularity but without adding complexity. 
This design minimize the data that is shared and the calls between nodes trying to offer a lightweight state transfer.

I hope to finish the work soon, but comments are welcome @jshaughn .

Thanks


hawkular-alerts-actions-plugins/hawkular-alerts-actions-aerogear/src/main/webapp-standalone/WEB-INF/jboss-all.xml
```diff
@@ -0,0 +1,25 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<jboss umlns="urn:jboss:1.0">
+  <jboss-deployment-dependencies xmlns="urn:jboss:deployment-dependencies:1.0">
+    <!-- Ensure metrics is deployed, started and initialized before we try to establish a session -->
```
jshaughn: Is this really supposed to say, "Ensure metrics is deployed..." ?

hawkular-alerts-engine/pom.xml
```diff
@@ -169,15 +187,84 @@
...
+            <executions>
+              <execution>
+                <id>default-compile</id>
+                <phase>compile</phase>
+                <goals>
+                  <goal>compile</goal>
+                </goals>
+                <configuration>
+                  <excludes>
+                    <exclude>org/hawkular/alerts/engine/impl/PartitionManagerEapImpl.java</exclude>
```
jshaughn: Is this going to be true for all EAP versions or maybe this class should be PartitionManagerEap6Impl.java ?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -292,7 +322,7 @@ private void removeTrigger(Trigger trigger) {
                 return false;
             });
         } else {
```
jshaughn: Does this need:

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -365,8 +395,9 @@ public void run() {
                 Collection<Data> newData = getAndClearPendingData();
                 Collection<Event> newEvents = getAndClearPendingEvents();
 
-                log.debugf("Executing rules engine on [%1d] datums, [%2d] events and [%3d] dampening timeouts.",
-                        newData.size(), newEvents.size(), numTimeouts);
+                log.debug("Executing rules engine on " + newData.size() + " datums, "
+                                + newEvents.size() + " events and "
+                                + numTimeouts +" dampening timeouts.");
```
jshaughn: This and other places should use if ( log.debugEnabled() ){...}

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
```diff
@@ -0,0 +1,844 @@
...
+                            log.debug("New member: " + member.hashCode());
+                        });
+                    }
+                    processTopologyChange();
+                    invokePartitionChangeListener();
+                }
+            });
+            triggersCache.addListener(new IspnListener() {
+                @CacheEntryCreated
+                public void onNewTrigger(CacheEntryCreatedEvent event) {
```
jshaughn: maybe call the arg 'cacheEvent' to distinguish better from an HKA Event.

jshaughn:
In general this looks good to me so far!

hawkular-alerts-actions-plugins/hawkular-alerts-actions-aerogear/src/main/webapp-standalone/WEB-INF/jboss-all.xml
lucasponce: +1
I reused the jboss-all.xml from other project and I forgot to change that comment :)

hawkular-alerts-engine/pom.xml
lucasponce: No, if I can avoid it.
In my first attempt I saw that ISPN/JGroups versions are very different, and far in versions, so I introduce the conditional compilation for this case.
But after study and make some tests, I could use the ISPN 5 API that is present in both EAP and Wildfly versions.
So no conditional compilations needed so far.
I forgot to change this one.
Good catch.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
Good catch.
Yes, I forgot to notify PartitionManager on the removeTrigger.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
lucasponce: +1

lucasponce:
Thx for your review.
I'm going to update in the branch the feedback from the EAP64 branch and your last suggestions.
I will be working on this branch for it.

lucasponce:
Hi @jshaughn, I've updated the PR and now with tests for standalone scenario.
I will do more tests but locally I confirmed no regressions and expected behaviour.
Please, take a look into the changes to see if you spot something missing.
Thanks,

lucasponce:
I have left the compatible code for EAP64 on the branch and it is mixed with the Clustering one.
Changes are trivial and can be reverted later once we have confirmation that platform will not be needed.
I fixed a small issue for a use case where partition logic was wrong and now cluster works fine.
But this branch is still in WIP, I have noticed a cornercase that is not yet covered.
For "chained" triggers, the generated events should be propagated as well across nodes.
In the future we can work with "affinity" concepts (i.e. put all chained triggers together), but for a first version, propagating data should be enough.
I will fix it as last task for this PR.

lucasponce:
I have fixed a last issues with +3 nodes cluster where partition state was not properly synchronized.
I have updated Cluster tests to coverage more harder scenarios and take some preliminar figures about performance as well.
@jshaughn, from my side, the branch is ready for a final review.

PR #127, opened 2015-12-11
Wildfly10

lucasponce:
This should wait until the WF10 migration is clear on all components (and likely after the Alpha8).
But I'm going to prepare the PR just if it's needed.
Take note that I'm using srcdeps that need to be updated to good ones.
I have ran the itests with success but more tests are welcome.


lucasponce:
Outdated.
I'm rebasing with master and cherry-pick valid commits and updating with recent versions to have a clean Wildfly10 branch.
I will update the PR once is ready.

PR #129, opened 2015-12-15
HWKALERTS-112 Upgrade commons and accounts

lucasponce:
Upgrade to accounts 1.1.4.Final and commons 0.2.5.Final, both needed for release.
Upgrading the way persona is injected in a similar way as inventory does.


lucasponce:
Merged as it is needed for the release

PR #130, opened 2015-12-16
Wildfly10

lucasponce:
Wildfly10 work.
Some re-organization on profiles to use same itest name for integration.

### DIFF
--- a/hawkular-alerts-rest-tests/pom.xml
+++ b/hawkular-alerts-rest-tests/pom.xml
@@ -48,9 +48,9 @@
     </dependency>
 
     <dependency>
-      <groupId>org.hawkular.bus</groupId>
+      <groupId>org.hawkular.commons</groupId>
       <artifactId>hawkular-bus-rest-client</artifactId>
-      <version>${version.org.hawkular.bus}</version>
+      <version>${version.org.hawkular.commons}</version>
       <scope>test</scope>
     </dependency>
 
--- a/hawkular-alerts-rest-tests/pom.xml
+++ b/hawkular-alerts-rest-tests/pom.xml
@@ -439,7 +439,6 @@
                   <includes>
                     <include>**/ActionsITest*</include>
                     <include>**/AlertsITest*</include>
-                    <include>**/ClusterITest*</include>
                     <include>**/ConditionsITest*</include>
                     <include>**/DampeningITest*</include>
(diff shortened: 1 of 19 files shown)


hawkular-alerts-rest-tests/pom.xml
```diff
@@ -165,8 +165,9 @@
 
   <profiles>
 
+    <!-- This profile requires a running server of hawkular or hawkular-alerts/standalone -->
     <profile>
-      <id>rest-only</id>
+      <id>itest-tests</id>
```
jshaughn: Please change to -Pitest-external to make more verbose that this option requires the external server to be running.

hawkular-alerts-rest-tests/pom.xml
```diff
@@ -353,7 +437,14 @@
...
                   <includes>
-                    <include>**/*ITest*</include>
+                    <include>**/ActionsITest*</include>
+                    <include>**/AlertsITest*</include>
+                    <include>**/ConditionsITest*</include>
+                    <include>**/DampeningITest*</include>
+                    <include>**/EventsITest*</include>
+                    <include>**/EventsLifecycleITest*</include>
+                    <include>**/LifecycleITest*</include>
+                    <include>**/TriggersITest*</include>
```
jshaughn: Why did you do this?  It means that if we add a new Itest we'll need to remember to add it here.

jshaughn:
The PR looks good.  Please update that profile name unless you have an issue with the proposal.  I am not sure we should merge this until we have a hawkular that is also running WF10.

hawkular-alerts-rest-tests/pom.xml
lucasponce: +1

hawkular-alerts-rest-tests/pom.xml
lucasponce: For some reason, the exclusion rule (which is more simple) it doesn't work well on this plugin.
I'm going to re-check but that was the main reason.
Putting a excludes rules didn't work.

lucasponce:
Thanks for the review.
As commented, I'm going to prepare a commit for the feedback and I won't merge on master, instead, I will create a Wildfly10 branch on hawkular-alerts.
Better to maintain both until hawkular/master is fully moved to Wildfly10.

PR #131, opened 2015-12-17
HAWKULAR-877 - Enabled the Secret Store for the REST API

jpkrohling:
With this, the backend starts accepting tokens generated via Accounts, instead of only OAuth Bearer tokens and username/password.


lucasponce:
Does this PR depends on Wildlfy10 or it can go on Wildfly9/10 ?

jpkrohling:
This does not depend on WF10.

PR #133, opened 2015-12-21
HWKALERTS-118 Remove redundant dataId cache on bus

lucasponce:
@jshaughn I think we have commented this.
As the engine is implementing the same cache, we can improve the code and remove this one, saving a listener.
Please, take a look if you are ok with the proposal.


jshaughn:
I'm not actually sure I'm OK with this.  Perhaps we should discuss a bit further as the more I look at things the more I'm concerned we may have an issue with the current design, both with this change and possibly independent of this change.

It's true that the listeners perform a duplicate filtering and perhaps we can figure a better way to do things.  But using the existing filtering means that the vast majority of data delivered to the bus by metrics will be filtered and never reach the alerts engine.

As designed this could be critically important. This cuts down on work in that we don't need to translate every metric into an alerts datum. It cuts down on traffic in that we don't then forward that data for downstream filtering.  But moreover, and where I think we may have a more serious issue, is the AlertsService.sendData() just passes the data to alertsEngine.addPendingData.  It is likely going to be bad to allow all data to reach the engine before filtering.  The engine EJB is a singleton and I'm starting to think will become a bottleneck that will start timing out.

Let's discuss this.

lucasponce:
As commented, after Xmas break I am going to add additional performance tests that can show some figures of both alternatives.
I agree that there are reasonable scenarios to test to check which solution can be better.
I will close this one.

PR #135, opened 2015-12-21
Add the embedded cassandra prop when transforming standalone.xml

jshaughn:


lucasponce:
Normally I use the property specifically on the command prompt but this addition is perfectly fine.
Thx

PR #137, opened 2015-12-22
HWKALERTS-108 Tag criteria issue

jshaughn:
Make sure to fail on bad tag criteria syntax


hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/AlertsITest.groovy
```diff
@@ -45,7 +45,7 @@ class AlertsITest extends AbstractITestBase {
         resp = client.get(path: "", query: [endTime:now, startTime:"0",statuses:"OPEN,ACKNOWLEDGED,RESOLVED"] )
         assert resp.status == 200 : resp.status
 
-        resp = client.get(path: "", query: [tags:"data-01,data-02"] )
+        resp = client.get(path: "", query: [tags:"data-01|*,data-02|*"] )
```
lucasponce: It looks ok, but one question, the syntax of tags needs to use the asterisk always ?
I mean [tags:"data-01,data-02"] == [tags:"data-01|*,data-02|*"] by default ?

jshaughn: Yes, as documented in the REST API calls.  There is no implicit '*' for the values.

PR #138, opened 2015-12-22
HWKALERTS-117 Update ISPN version for WF10

lucasponce:


lucasponce:
I have tested a manual cluster of 3 nodes against Cluster specific integration tests and no regressions found.

PR #139, opened 2016-01-06
Hwkalerts 101

jshaughn:
Flesh out multi-tenancy support by ensuring that triggers are not referenced only by triggerId but rather by tenantId+triggerId.

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/AvailabilityConditionEval.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/AvailabilityConditionEval.java
@@ -31,6 +31,8 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include;
  */
 public class AvailabilityConditionEval extends ConditionEval {
 
+    private static final long serialVersionUID = 1L;
+
     @JsonInclude(Include.NON_NULL)
     private AvailabilityCondition condition;
 
@@ -64,6 +66,11 @@ public class AvailabilityConditionEval extends ConditionEval {
 
     public void setValue(AvailabilityType value) {
         this.value = value;
+    }
+
+    @Override
+    public String getTenantId() {
+        return condition.getTenantId();
     }
 
     @Override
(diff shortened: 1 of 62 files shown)


lucasponce:
Hi,
I've started with the review.
Travis is failing due this:
[WARNING] Missing header in: /home/travis/build/hawkular/hawkular-alerts/hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/listener/MetricDataListener.java
[WARNING] Missing header in: /home/travis/build/hawkular/hawkular-alerts/hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/messages/MetricDataMessage.java

I guess it is an easy fix.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/AvailabilityConditionEval.java
```diff
@@ -31,6 +31,8 @@
  */
 public class AvailabilityConditionEval extends ConditionEval {
 
+    private static final long serialVersionUID = 1L;
```
lucasponce: I thought that serialVersionUID variable from implementing Serializable are not mandatory with newer versions of JDK if I am not wrong.

jshaughn: I'm not an expert but this is my understanding:  It's not mandatory as a value will be generated if not specified.  But the generated value could change based on class changes that may not have actually made the Object serialization fail, so an unnecessary incompatibility.  Making it explicit just puts it under our control.  It doesn't affect xml/json-style serialization, just the built in Java Object serialization, so it's sort of a no-op for us at the moment, I think.  Anyway, it makes the Eclipse warning go away :)  But did you want me to remove them?

lucasponce: No need to remove them :-).
I remember that Eclipse tends to complain about it but I didn't see it on Idea, just curiosity.
It is fine to have them.

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/LifecycleITest.groovy
```diff
@@ -235,7 +235,7 @@ class LifecycleITest extends AbstractITestBase {
         // Instead of going through the bus, in this test we'll use the alerts rest API directly to send data
         Map<String,String> context = new HashMap<>(1);
         context.put("contextName","contextValue");
-        Data avail = Data.forAvailability("test-autoresolve-avail", System.currentTimeMillis(), DOWN, context);
+        Data avail = Data.forAvailability("", "test-autoresolve-avail", System.currentTimeMillis(), DOWN, context);
```
lucasponce: One doubt, an empty tenant is going to work on this test ?

lucasponce: I respond to myself: yes as it will be populated on the Rest handler.

PR #140, opened 2016-01-07
Accept metric data from datamining module

pavolloffay:
Accept predicted metrics data from datamining module

### DIFF
--- a/hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/listener/MetricDataListener.java
+++ b/hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/listener/MetricDataListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


pavolloffay:
@lucasponce Could you please review?

lucasponce:
Yes. Today I am on PTO but I am back tomorrow. Thx

jshaughn:
This means that datamining must now maintain a MetricDataMessage class that stays in sync with Metric's MetricDataMessage class, right?  This is probably OK as that class is probably unlikely to change.  Why not use the same class, to avoid a dependency?

pavolloffay:
I didn't use message class from metrics because it is not in public API. Component where it lives is packaged as war.

jshaughn:
OK, that makes sense.   Actually, perhaps you could just use org.hawkular.alerts.bus.messages.MetricDataMessage.  That's in the alerts-bus jar. It would maybe need to move to the alerts-bus-api jar.  Let's wait and see what Lucas has to say as well.

lucasponce:
I am agree with Jay.
Inside hawkular-bus there are a package called org.hawkular.alerts.bus.messages that at the beginning the use was internal for own listeners, but as these messages can be used externally as Pavol's work, the ideal would be to have them on a public api. 
So, I'm +1 to move them to org.hawkular.bus.api package in hawkular-alerts-bus-api, then design can be clean and we can avoid to check the class name.
If both are agree we can fill a JIRA for this and do it, it is not a big deal.
Would this impact in your work @pavolloffay ?

jshaughn:
I have created https://issues.jboss.org/projects/HWKALERTS/issues/HWKALERTS-119 for the class move.  Assigned to Lucas.

pavolloffay:
I'm ok with using message class  alerts. :+1:

pavolloffay:
Maybe the best solution would be move MetricDataMessage from metrics to some metrics bus API or bus commons because Alerts and Datamining are using the same message. https://github.com/pavolloffay/hawkular-alerts/blob/datamining-metric-data/hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/listener/MetricDataListener.java#L108

lucasponce:
I have prepared a PR for this topic:
https://github.com/hawkular/hawkular-alerts/pull/144

PR #146, opened 2016-01-11
Fix wrong return bus message type

lucasponce:


jshaughn:
Seems OK to me. Pavol?

hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/listener/AvailDataListener.java
```diff
@@ -115,7 +115,7 @@ protected void onBasicMessage(AvailDataMessage msg) {
     protected String convertReceivedMessageClassNameToDesiredMessageClassName(String className) {
 
         if (className.equals("org.hawkular.metrics.component.publish.AvailDataMessage")) {
-            return "org.hawkular.alerts.bus.messages.AvailDataMessage";
+            return "org.hawkular.alerts.bus.api.AvailDataMessage";
```
pavolloffay: Why not using CanonicalName?

lucasponce: You mean to change the return "org.hawkular.alerts.bus.api.AvailDataMessage" per return AvailDataMessage.getClass().getCanonicalName() ?
If so, I don't see major differences. I think maintaining the constant can be better readable.

pavolloffay: yes .class.getCanonicalName(), but constant is also ok. It was just and idea that could avoid previous error. 

https://github.com/hawkular/hawkular-datamining/blob/inventory-direct-query/hawkular-datamining-bus/src/main/java/org/hawkular/datamining/bus/listener/MetricDataListener.java#L93

lucasponce: Your suggestion is also correct.
I just followed the original code as I think is more readable and this class shouldn't change.
I will let to Jay to decide it.
Also, I've created a JIRA to improve the itest for this scenario to check in the next versions that things are working properly.
https://issues.jboss.org/browse/HWKALERTS-121

jshaughn: I like the idea of getCanonicalName() but will leave as-is in order to get this in asap.  It can be changed in passing as part of future work.

PR #148, opened 2016-01-13
HWKALERTS-121 Add integration tests for bus scenarios

lucasponce:
Minor PR to cover bus scenarios.
Repeating some tests but sending data directly through the bus instead using /data endpoint.
This will give better testing support for hawkular scenarios.
Noted that with new changes on bus, using javax.jms packages are needed to connect remotely to bus destinations instead to use the old abstract hawkular-bus api.

### DIFF
--- a/hawkular-alerts-rest-tests/pom.xml
+++ b/hawkular-alerts-rest-tests/pom.xml
@@ -42,16 +42,42 @@
 
     <dependency>
       <groupId>org.hawkular.alerts</groupId>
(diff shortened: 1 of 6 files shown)


jsanda:
I noticed that you are using a remote JMS connection for your tests. Metrics did the same originally. I refactored the metrics tests to use Arquillian which allowed me to test the components more easily. Take a look at https://github.com/hawkular/hawkular-metrics/tree/master/bus if you are interested.

PR #149, opened 2016-01-14
HWKALERTS-125 Close JMS cached connection on bus action plugin

lucasponce:
This fix a warning that comes from the new bus migration infrastructure.
It was triggered on hawkular showing an annoying warning.
This issue is not present on intergation tests and I ran some manual tests on hawkular and it seems gone.


jshaughn:
I actually thought this had already been fixed as I had stopped seeing it locally.  But I guess I was mistaken.  We can put this in 0.8.2 if we need to cut another release due to an accounts dependency.

PR #151, opened 2016-01-16
Unify init files

lucasponce:
- Use single file for alerts definitions import.
- Improve the flexibility of json format (only non default fields are necessary).
- Aligned how UI handles full triggers (more easy to move UI code into the server).
- Additional REST endpoint for create/get full trigger.


hawkular-alerts-engine/src/test/resources/hawkular-alerts/alerts-data.json
```diff
@@ -0,0 +1,367 @@
+{
```
jshaughn: perhaps triggers-data.json as the name of this file?

lucasponce: well, I don't have a strong position about naming, I choosed alerts-data.json, as we are storing triggers + actions.
perhaps with alerts-data.json we can get rid of the hawkular-alerts/ folder in the future.
For this version I'm maintaining the "hawkular-alerts" folder.

jshaughn:
What about the findTriggers (by criteria)?  Should we have the ability to fetch full triggers in this way?

What about createGroupTrigger?   Shouldn't it be changed to support full trigger in the same way as createTrigger?

lucasponce:
About findTriggers() I have doubts. The UI doesn't use it. As separating one query for the list of queries and a second for the detail is fine.
Probably the createGroupTrigger might have sense, but in the full trigger you can add a groups, if I'm not wrong in the alerts-data.json example the trigger-8 is a group and it is using the same API.

hawkular-alerts-engine/src/test/resources/hawkular-alerts/alerts-data.json
jshaughn: no problem, I forgot about actions

jshaughn:
It's true a group trigger can be defined via the json data file.  But it seems it should also be able to be defined as a full trigger in POST /triggers/members analogously to the change made to POST /triggers.

BUT, I am OK if we don't make that change at the moment.  In my POC for data-driven member triggers I have some more changes in the group triggers area. It may be worth waiting to see if that POC will fly, and then revisiting API at that point, and after a merge.

lucasponce:
+1
I think that once that the POC of groups triggers is progressed the API should be revisited.
I am going to merge this PR to consolidate it.

PR #152, opened 2016-01-18
Execute in separate threads heavy initialization tasks

lucasponce:
Moved reload() and initialData() tasks out of the @PostConstruct thread.
This enhancement will help to avoid locks in slow scenarios.
@jshaughn can you take a look and test ?


lucasponce:
I am reviewing this PR and I think that there is another aspect that it can be applied.
The singleton classes can be reduced just to one to store listeners & initialization state.
I am going to close this PR and prepare a new one with the changes.

PR #153, opened 2016-01-19
Init separate thread

lucasponce:
On this PR:
- Optimize Singleton beans.
- Initialize heavy tasks on separate thread.
- Fix warnings on bus usage.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -30,12 +30,14 @@ import java.util.TimerTask;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
+import javax.annotation.Resource;
 import javax.ejb.EJB;
 import javax.ejb.Local;
 import javax.ejb.Singleton;
 import javax.ejb.Startup;
 import javax.ejb.TransactionAttribute;
 import javax.ejb.TransactionAttributeType;
+import javax.enterprise.concurrent.ManagedExecutorService;
 
 import org.hawkular.alerts.api.model.condition.CompareCondition;
 import org.hawkular.alerts.api.model.condition.Condition;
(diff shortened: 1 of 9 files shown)


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -165,12 +170,14 @@ public void initServices() {
...
-                assigned to it and notify other nodes to load rest of the triggers.
-             */
-            reload();
+            executor.submit(() -> {
+                /*
+                    A reload() operation means that all triggers from the backend should be reloaded into
+                    the AlertsEngine memory. In a distributed environment, the node that execute the reload()
+                    operation loads the triggers assigned to it and notify other nodes to load rest of the triggers.
+                 */
+                reload();
```
jshaughn: are we in any danger of concurrent reloads?

jshaughn:
+1 for eliminating singletons, which have already generated intermittent timeout issues.  My only concern now is whether we are safe from unwanted concurrent processing wrt CRUD and engine handling.  Also, a bit unrelated, we may want to talk about how to improve notifications to better handle "rapid fire" changes.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
lucasponce: Not with that code, as it is executed on a Singleton marked a Startup.
But we have for historical reasons a /reload endpoint that it can be removed safely in the future.
I don't remember any test-case using the REST endpoint for reload it was there for the first primitive tests.

jshaughn: Please remove those REST endpoints as part of this PR.  I think we deprecated them already anyway.

lucasponce:
About the concurrent processing, a concurrent scenario has always a minimal probability as reaching a full deterministic design is hard, but I should say that the design has not visible concurrent flaws.
About the notifications, this is also a pending topic, as I didn't introduce them support for distribute scenarios as well. I guess that the design should be similar, someone (C* or us) should propagate the notification across the nodes.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java

jshaughn:
I wasn't really asking about distributed processing. I was asking just about the change from singletons and whether this would expose us to any obvious concurrent update issues, either with CRUD or with manipulation of the drools engine working memory.

lucasponce:
No, I think we are safe.
The components were singletons because it had some state, so, after state has been moved, the logic can be stateless. Having a pool of components (which is the main change) won't affect as state is still on singleton components, case is those have been reduced in the enhancement, optimizing the concurrent of most of the non-write queries.

PR #155, opened 2016-01-21
HWKALERTS-127 Data Driven Group Triggers

jshaughn:


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -1482,14 +1480,39 @@ private void handleResolveOptions(String tenantId, String triggerId, boolean che
...
+        checkDataDrivenGroupTriggers(data);
         alertsEngine.sendData(data);
     }
 
+    private void checkDataDrivenGroupTriggers(Collection<Data> data) throws Exception {
+        if (!dataDrivenGroupCacheManager.isCacheActive()) {
+            return;
+        }
+        for (Data d : data) {
+            if (isEmpty(d.getSource())) {
```
lucasponce: I wonder if we should check at this level is Data.source == __none__ as well.
If I understood the design, instead to use null/empty value (perhaps in other part of the code I can review it) we use that string constant. So, as checkDataDrivenGroupTriggers() will be executed very often, perhaps this optimitization would be fine.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DataDrivenGroupCacheManager.java
```diff
@@ -0,0 +1,201 @@
...
+
+        definitions.registerListener(new DefinitionsListener() {
+            @Override
+            public void onChange(DefinitionsEvent event) {
+                updateCache();
+            }
+        }, DefinitionsEvent.Type.CONDITION_CHANGE);
+    }
+
+    private synchronized void updateCache() {
```
lucasponce: Just a note.
As we commented, when we introduce the distributed listener this method can be optimized as well.
It can maintain a cache of the groups model, reducing queries to C*.
This is not to fix anything now, just a comment for future.

lucasponce:
No obvious issues found after a couple of reviews.
Two aspects to optimize the cache and the listeners once that we introduce distributed listeners.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
jshaughn: It's unlikely that a caller would provide the private SOURCE_NONE value but if they did it could cause an issue because we don't really want member triggers for the SOURCE_NONE source.  I added protection for this not here but in the needSourceMember() method, inside the cache code.

PR #156, opened 2016-01-26
Time on actions

lucasponce:
New features:
- Define actions by status.
- Add time constraints on action generation, absolute or based on calendar.

@jshaughn this PR is ready for your review. Thx


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/Action.java
```diff
@@ -26,28 +25,17 @@
...
  * An action is the abstract concept of a consequence of an event.
- * A Trigger definition can be linked with a list of actions.
  *
- * Alert engine only needs to know an action id and message/payload.
- * Action payload can optionally have an event as payload.
- *
- * Action plugins will be responsible to process the action according its own plugin configuration.
+ * An Action object represents a particular action linked with a specific event.
+ * Action objects are generated by the Alerts engine and processed by plugins.
+ * An Action object stores the eventId property and optionally may contain the full Event object with its linked.
```
jshaughn: "with its linked." ?   Seems like this comment was not finished.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/Action.java
```diff
@@ -59,9 +47,6 @@
     @JsonInclude(Include.NON_NULL)
     private Event event;
 
-    @JsonInclude(Include.NON_EMPTY)
-    private Map<String, String> properties;
-
     @JsonInclude(Include.NON_NULL)
     private String result;
```
jshaughn: It may be captured elsewhere but do we persist other "result" information like actionStartTime, actionEndTime, etc?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/ActionDefinition.java
```diff
@@ -0,0 +1,132 @@
...
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        ActionDefinition that = (ActionDefinition) o;
+
+        if (tenantId != null ? !tenantId.equals(that.tenantId) : that.tenantId != null) return false;
+        if (actionPlugin != null ? !actionPlugin.equals(that.actionPlugin) : that.actionPlugin != null) return false;
+        if (actionId != null ? !actionId.equals(that.actionId) : that.actionId != null) return false;
+        return properties != null ? properties.equals(that.properties) : that.properties == null;
```
jshaughn: Are you sure you want the properties involved in equals/hashcode or is the actionId sufficient?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
```diff
@@ -80,9 +80,9 @@
     @JsonInclude(Include.NON_EMPTY)
     protected Map<String, String> tags;
 
-    /** A map with key based on actionPlugin and value a set of action's ids */
+    /** A list of links to actions represented by TriggerAction*/
     @JsonInclude(Include.NON_EMPTY)
-    private Map<String, Set<String>> actions;
+    private Set<TriggerAction> actions;
```
jshaughn: I like this much better, using a Set of objects gives us a lot more flexibility.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/TriggerAction.java
```diff
@@ -0,0 +1,386 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+/**
+ * Link an ActionDefinition with a Trigger.
+ *
+ * It can add optional constraints that determine when an action will be executed:
+ *
+ * - A set of Alert.Status (represented by its string value).
+ *   The action will be executed if the Alert which is linked is on one of the states defined.
+ *   This is not applicable if the action is linked with an Event.
```
jshaughn: Perhaps replace "This is not applicable if the action is linked with an Event" with:

"Unlike Alerts, Events don't have lifecycle, TriggerActions on Events are all executed at Event creation time."

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/TriggerAction.java
```diff
@@ -0,0 +1,386 @@
...
+    private String actionPlugin;
+
+    @JsonInclude
+    private String actionId;
+
+    @JsonInclude(Include.NON_EMPTY)
+    Set<String> states;
+
+    @JsonInclude(Include.NON_NULL)
+    private String calendar;
```
jshaughn: I think it may be better to introduce something like a TimeConstraint class that encompasses the time handling, offers some simple methods like TimeConstraint.satisfied, easy constructors, etc, and may allow us to more easily build up the power of the time constraints.

Also, the current approach iiuc is to define one TriggerAction per TimeConstraint, so for example, send email to Joe if it's during evening hours.  And we'd need another TriggerAction to send email to Mary if it's during daytime hours.  Any thoughts about a single TriggerAction that can use different config based on a set of time constraints (or default config that is overrided by any active time constraint)?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/Action.java
lucasponce: How it sounds "may contain the full Event object which its linked" ?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/Action.java
lucasponce: Not at the moment.
I think is too much verbosity.
Today we have:
- Timestamp on the data used for generate event/alert.
- Timestamp on event creation and alert lifecycle (open, ack, resolve).
- Timestamp on action creation.
So, an actionEndTime would give more verbosity, but as the async nature of the Action, I'm not fully sure if this is needed.
We can add it, no big deal on that, but I thought in my first implementation that storing the result should be enough.
Perhaps, I can add it later as a separate tasks if needed.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/ActionDefinition.java
lucasponce: Point is that properties are mutable, so yes, I think those properties are needed on equals/hashCode

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/TriggerAction.java
lucasponce: Nice english structure. Thanks for the suggestion.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/TriggerAction.java
jshaughn: One more thought, perhaps using a well-known format for the time ranges would be useful?  Maybe cron?  Quartz scheduler has a well-used CronExpression class that could be useful. https://quartz-scheduler.org/api/2.2.1/org/quartz/CronExpression.html  (although, maybe it's overkill to include quartz just for this utility...it's not a huge jar though).

lucasponce: Two comments:
- Related to use a calendar expression versus a TimeConstraint object: I've found a expression more simple to read. As this object has a strong mapping with json. We can add some static helpers at that level to do more easy to create it, but for readability I would like to avoid complex nested objects on the   json representation.

- About where to place the constraint, yes, there are several candidate places. My goal for this design was to maintain previous behaviour as simple as we had, but also introduce this. I was thought which place could be more interesting, to place a time constraint on the link with the trigger or in the action itself. Well, both are valid, but I think that more common scenario will be to use in the trigger. For example, an action definition will be used more like a "group" (i.e. I may have an action to notify sysadmins, another action to notify developers, and another action for managers). So, as action definitions can be reused over multiple triggers, I thought that the constraints should live in the relation between trigger and action, hence that I put the logic there.

Of course, we can add both, in the link and the action definition, and add more complexity in the rules to check, but I would like to add that in an iterative way to check how these features looks like in real scenarios.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/util/ActionsValidator.java
```diff
@@ -0,0 +1,142 @@
...
+    private static final Logger log = Logger.getLogger(ActionsValidator.class);
+    /**
+     * Validate if an Event should generate an Action based on the constraints defined on a TriggerAction.
+     *
+     * @param triggerAction a TriggerAction where status and time constraints are defined.
+     * @param event a given Event to validate against a TriggerAction
+     * @return true if the Event is validated and it should generated an action
+     *         false on the contrary
+     */
+    public static boolean validate(TriggerAction triggerAction, Event event) {
```
jshaughn: For example, with use of quartz and cron much of this home-grown logic could be replaced by CronExpression.isSatisfiedBy().

lucasponce: Well, I thought as well, but I discarded as we don't need to use a programmer and to introduce that dependency just to use the Quartz format is not worth (from a tradeoff perspective).

Also the validate() method is not only related to time, as additional constraints can be added in the future, validate represents that a specific Event should generate an Action given a TriggerAction.
Additional constraints (time or not related) could be placed here and use this.

I admit that looking how Action objects are generated, it could have more sense to have a factory approach, given an Event/Alert and a TriggerAction it should generate the Action. Today this is done in one side calling just the constructor, but perhaps it makes sense to move then to this class as well.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/CassPersistenceTest.java
```diff
@@ -61,14 +61,7 @@ public static void initSessionAndResetTestSchema() throws Exception {
...
 
-        Session session = CassCluster.getSession();
-
-        // try and clean up if this was somehow left around by a prior run
-        try {
-            session.execute("DROP KEYSPACE " + keyspace);
-        } catch (Throwable t) {
-            // never mind, it may not be there
-        }
+        Session session = CassCluster.getSession(true);
```
jshaughn: It used to be that we would drop hawkular_alerts_test keyspace by default, is that still true?  It seems now that we must have hawkular-alerts.cassandra-keyspace set to "hawkular_alerts_test". Where is that done?

Perhaps maybe we want  CassCluster.getCleanSession(String keyspace) or something like that?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/Action.java
jshaughn: Oh, I see what you are saying.  I think just leave it as:

"An Action object stores the eventId property and optionally may contain the full Event object."

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/CassPersistenceTest.java
lucasponce: I've added a new one, to drop the scheme if exists. The getSession() works as previous one.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/Action.java
jshaughn: I'm fine with waiting for a user request to add more information.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/ActionDefinition.java
jshaughn: OK.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/util/ActionsValidator.java
jshaughn: >  that dependency just to use the Quartz format is not worth

It's not a quartz format, it's the unix cron format, which I think is pretty well known (even if somewhat arcane).  My point is just that this sort of time interval/schedule stuff has been done before and it may be useful to use a library for this part.

Also, i didn't say validate() would be replaced with CronExpression.isSatisfiedBy(), I said, " much of this home-grown logic could be replaced", meaning all of the proposed time-expression stuff.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/CassPersistenceTest.java
jshaughn: right, i see that, but what I'm asking is what keyspace gets dropped now, in CassPersistenceTest's new call to  CassCluster.getSession(true)?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/util/ActionsValidator.java
lucasponce: Well, I agree that unix cron format is well known, but not sure if it is an easy format for this.
In our case we want to define a startTime,endTime, and cron is more flexible for repetitive schedulings.
Anycase, I think that the calendar expression is quite straight forward:
startDate;endDate
when startDate can be absolute or relative.
And relative is based on [[Month]Day]Hour
Which is readable.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/CassPersistenceTest.java
lucasponce: Not, only it is dropped on initialization.
It means that once the keyspace is initialized and cached is not replaced.
Perhaps I can re-arrange this to be more clear, I admit it that only drop on initialization might be not clear enough.

jshaughn: I don't think I'm being clear. From what I can see this call to getSession(true) will drop keyspace hawkular_alerts, not keyspace hawkular_alerts_test.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/util/ActionsValidator.java
jshaughn: OK, yes, now that I think more I agree that cron is more about scheduling and less about defining an interval. I will look again in more detail at your proposed format.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/TriggerAction.java
jshaughn: I may need a demonstration of how to accomplish a few things. For example:
- An email action that sends mail to 'admins' but from 00:00-06:00 it sends to 'on-call'.
- An email action that sends mail to 'admins' on weekdays but on weekends to 'on-call'
- A sms action that is active only for alerts created from 22:00-06:00.

We can discuss this further off-PR...

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/CassPersistenceTest.java
lucasponce: I see what you mean. Yes, I will fix that.
Good catch.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/TriggerAction.java
lucasponce: It is fine.
I'm going to add an additional commit with examples you pointed.
It can be worth to comment out on the code as well.

lucasponce:
I've added additional examples and fix some issues from review.
Sorry, the commit of "built-in smtp testing" library was included.
It's harmless but it was part of another branch, I think is ok to add it.

lucasponce:
As commented, perhaps the terminology for Mx and Dx for relative dates could be less readable.
I can change them for short/long english constants as monday (mo) or january (jan), that would maintain the same philosophy or we can add a full object for that.
My suggestion would be to change the expression with short/long formats and check how this feature is accepted, and change into a TimeConstraint object later in future if we see that we have a lack of flexibility for that.

lucasponce:
I've introduced a TimeConstraint object to express in a better readable way the calendar expressions.
The PR is not finished yet.
I would like to add constraints (states and/or time) at Trigger level but also at ActionDefinition level.
So, if a constraint is defined at Trigger level it has priority over those defined at ActionDefinition level.
But constraints defined at Action level could be good to modify just the definition.
It's low hanging fruit with the present approach, so I think it could be good to add it for a first iteration.

lucasponce:
I have added support to have action constraints defined at trigger and at action level, so I think we can cover more uses cases.
please, @jshaughn can you review it ?
Thanks

jshaughn:
I like the addition of TimeConstraint (surprise :) ).  A little more definition around semantics and I think the PR can be merged.

lucasponce:
"Relative" term is used as "relative" to the current time/date.
So, a {startTime: "2016.02.01,09:00", endTime: "2016.02.01,18:00", relative: false} is doesn't need an additional info to calculate the interval.
But a {startTime:"09:00", endTime:"18:00"} is relative in the meaning that if now it is "15:00" the constraint is satisfied in the same way that yesterday/tomorrow at "15:00" will be satisfied as the interval is "relative".

You have brought an interesting question.
With the current implementation, {startTime:"Mon,09:00", endTime:"Fri,18:00"} has a "repetitive" behaviour, in the sense that if now it is "Monday,15:00" that will be satisfied the constraint, but if it is "Monday,18:01" the constraint won't be satisfied.

I tend to think that perhaps behaviour should be to consider the interval linear.
It is a flaw in the implementation, as it is validated by steps in the relative dates (first it validates month, if correct, it validates days, and finally it validates time).
But perhaps the behaviour is useful as it is and it is just a flaw in the documentation that is not clear enough.

I am for fix that and implement relative dates in linear behaviour, but not sure what both formats can be more useful (I think that we can include both with an additional flag, but perhaps it is "overarch" for this iteration).

jshaughn:
I am fine with whatever level of impl you want to have for this iteration.  My desire is only for clear documentation of the semantics, preferably with some examples. And a blog.  And a demo :)

jshaughn:
When you are satisfied and make your final commit I will merge.

lucasponce:
@jshaughn I've rebased with your last commit from yesterday.
I think PR is ready to be merged from my side.

PR #157, opened 2016-01-28
Email properties

lucasponce:
This PR moves the java.mail.Session to be managed via system properties.
This helps to configure from several points.
In alerts, it can be configured from hawkular-alerts.properties file, and in hawkular it can be configured in several places where system properties will be defined.
Also, it supports several definitions by tenant.

Other minor enhancements are to introduce an embedded smtp server for testing.

### DIFF
--- a/hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/pom.xml
+++ b/hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/pom.xml
@@ -40,6 +40,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.icegreen</groupId>
+      <artifactId>greenmail</artifactId>
+      <version>${version.com.icegreen}</version>
+    </dependency>
+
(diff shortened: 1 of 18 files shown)


hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/pom.xml
```diff
@@ -40,6 +40,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.icegreen</groupId>
+      <artifactId>greenmail</artifactId>
+      <version>${version.com.icegreen}</version>
```
jshaughn: should this be <scope>test</scope> ?

lucasponce: Oh, yes.
Good catch.
I'm going to prepare a minimal fix for it and pushed to master.

PR #158, opened 2016-01-28
HWKALERTS-129 improve context and tag handling when adding a member trigger

jshaughn:
- merge member-supplied context with group-supplied context
- allow members to provide tags at add-time.
  - merge member-supplied tags with group-supplied tags
- allow members to provide member-level description
- allow the trigger name to default to the group name

Also
- fix a bug in adding members with a RateCondition


lucasponce:
I have run the travis CI and I see this in the logs:

Tests in error: 
  TriggersITest.testGroupTrigger:128 Â» GroovyRuntime Could not find matching con...

Tests run: 41, Failures: 0, Errors: 1, Skipped: 0

Perhaps is related to the PR.

PR #159, opened 2016-02-03
Preliminar cleanup tasks

lucasponce:
PR with some cleanup tasks:
- Use the EmbeddedCassandraService from hawkular-commons, to prepare a better itests in next iteration.
- Clean the System.out.println calls left there in some preliminar tests.

I think is good to include this one as it touches several files, so we can save time in the future with conflicts.


lucasponce:
@jshaughn I've also added a travis command to take care of the perf test automatically.
It will take more time, but if it is ok with travis, it will help us to validate that rules engine and backend is under regular tests.

PR #160, opened 2016-02-04
HAWKULAR-985 - Removed direct dependency on secret store.

jpkrohling:


lucasponce:
Travis seems happy with the PR.
But please, don't upgrade main hawkular with alerts.
@jshaughn is working on main hawkular and hawkular-alerts side and he should control when to upgrade in order to sync and not break some features where he is working now.
Thanks

jpkrohling:
Alright. Is there another component I should *not* be updating in Hawkular main, from the Alert's perspective? Once everything is merged, I'll send a PR to Hawkular bumping everything, except Alerts.

jshaughn:
Actually, I think it is OK.  I will merge the PR and we can move Hawkular to a SRCDEP of Alerts.  In my local branch I am using alerts SNAPSHOT with Hawkular, but I'll resolve everything later when this stuff I'm working on is in better shape and ready for the real world.

PR #163, opened 2016-02-08
Import export triggers

lucasponce:
Main goals:
- Clean and unify JSON deserialization.
- Unify and review swagger doc for REST API (criteria imported from metrics to maintain same way to describe and visual rules.
- Add import/export by tenant features.


lucasponce:
I have fixed a small issues that travis detected but I could not reproduce locally.
This time travis was fine and spotted a real issue.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/GroupConditionsInfo.java
```diff
@@ -35,7 +39,11 @@
  * @author lucas ponce
  */
 public class GroupConditionsInfo {
+
+    @JsonInclude(Include.NON_EMPTY)
```
jshaughn: These fields should not really be empty but this is fine.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/JacksonDeserializer.java
```diff
@@ -377,6 +467,30 @@ public static Condition deserializeCondition(JsonNode node) throws JsonProcessin
         return map;
     }
 
+    public static class ConditionException extends JsonProcessingException {
```
jshaughn: Should this be @ApplicationException to avoid being wrapped in EJBException if thrown?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/export/AlertDefinitions.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * Representation of a list of full triggers (trigger, dampenings and conditions) and actions definitions.
+ *
+ * @author Jay Shaughnessy
+ * @author Lucas Ponce
+ */
+public class AlertDefinitions {
```
jshaughn: Why not FullTriggers?  This seems to introduce new terminology.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/GroupConditionsInfo.java
lucasponce: I put it to avoid conditions = null values during json deserialization.
It is not a big deal we can leave it as it or just @JsonInclude.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/JacksonDeserializer.java
lucasponce: Not sure if the RestEasy will handle the @ApplicationException, but anyway, adding the annotation is harmless and it can save long stack of exceptions.
I will add it.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/export/AlertDefinitions.java
lucasponce: FullTriggers don't include ActionDefinitions IMO.
So a concept of FullTriggers + ActionDefinitions is needed.
AlertDefinitions came to me as the most direct definition.
But I can be open to negotiate :).

jshaughn: I think we went through this discussion before in some other PR, when discussing the name of the consolidated alerts-data.json file.  The problem I have is that there is nothing called an "Alert" definition.  I find it confusing.  This should either be "AlertsDefinitions" signifying that these are the definitions for Hawkular Alerts, or possibly just Definitions.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/GroupConditionsInfo.java
jshaughn: just leave it, i was only noting that it would likely not be empty given the purpose of the utility class.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/JacksonDeserializer.java
jshaughn: If you add it make sure you add the appropriate attributes.

jshaughn:
I am done with the review. Once the comments are settled I can merge.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/export/AlertDefinitions.java
lucasponce: "Definitions" works ok.
I think the only scenario is a wrapper of FullTriggers + ActionDefinitions so I'm fine with that term.

PR #166, opened 2016-02-15
HAWKULAR-1026 unshallow the git clone so that license plugin stopps flagging fil

ppalaga:


ppalaga:
@lucasponce or @jshaughn could you please review and eventually merge?

PR #169, opened 2016-03-03
Upgrade to Metrics 0.13.0.Final

ppalaga:


ppalaga:
@lucasponce @jshaughn this is a blind shot without testing locally. I hope Travis will like the new Metrics so that you can release Alerts for Hawkular M11 after merging this PR.

PR #171, opened 2016-03-13
Adding a simple webhook plugin

lucasponce:
@pilhuhn I've moved the previous webhook plugin as "webhook-global" as it was designed as pure gateway.
I've added a new one that I think it can suit on your needs.
Please let me know if this works for you, I've tested on standalone without issues.

@jshaughn for some reason, the itest profile doesn't work fine with last changes, I didn't receive any travis error, perhaps we need to take a look. Does it works for you (as I changed my machine, I don't discard that I have something wrong locally).

Thanks


lucasponce:
@jshaughn strange. Yesterday I can't execute a mvn clean install -Pitest and I remember you commented me that it was an issue as well in your machine.
Today I ran another round and it passed.
Also travis doesn't show anything wrong.
So perhaps it's something related to the environment.

PR #176, opened 2016-04-14
HWKALERTS-140 Validate actions definitions against plugins deployed

lucasponce:
Highlights of this PR:
* Actions are validated against plugins:
 - An action should have a plugins deployed in the system.
 - An action should follow the properties defined in a plugin.
* Initialization files logic has been removed:
 - As we have import/export logic, initialize the definitions via files was only used for some tests and had some caveats in clustering logic too, so I think that with the import/export we are covered for the same functionality and more consistency.
- Also it was problematic as plugins deployment needs the Engine fully deployed, so it should be moved in another place.
* The use case of what to do when a plugin is undeployed is not covered, now it maintains the action definitions but it will fail at the moment of execution of the action. Perhaps a nice to have is to mark some state, but then we need to add a way to detect "undeployments" of .war, so for the moment I think that could be a cornercase and it is good as it is working today.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
@@ -16,8 +16,6 @@
  */
 package org.hawkular.alerts.engine.impl;
 
-import java.io.File;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -31,15 +29,12 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
-import javax.annotation.PostConstruct;
-import javax.annotation.Resource;
 import javax.ejb.AccessTimeout;
 import javax.ejb.EJB;
 import javax.ejb.Local;
 import javax.ejb.Stateless;
 import javax.ejb.TransactionAttribute;
 import javax.ejb.TransactionAttributeType;
-import javax.enterprise.concurrent.ManagedExecutorService;
 
 import org.hawkular.alerts.api.json.JsonUtil;
 import org.hawkular.alerts.api.model.Severity;
@@ -95,13 +90,6 @@ import com.google.common.util.concurrent.Futures;
 @Stateless
 @TransactionAttribute(value = TransactionAttributeType.NOT_SUPPORTED)
 public class CassDefinitionsServiceImpl implements DefinitionsService {
-    /**
-     * Used on distributed environments.
-     * If present, the initial data are not loaded on this node.
-     */
-    public static final String SKIP_INIT_DATA = "hawkular-alerts.skip-init-data";
-    private static final String JBOSS_DATA_DIR = "jboss.server.data.dir";
-    private static final String INIT_FOLDER = "hawkular-alerts";
     private final MsgLogger msgLog = MsgLogger.LOGGER;
     private final Logger log = Logger.getLogger(CassDefinitionsServiceImpl.class);
 
@@ -110,9 +98,6 @@ public class CassDefinitionsServiceImpl implements DefinitionsService {
 
     @EJB
     AlertsContext alertsContext;
-
-    @Resource
-    private ManagedExecutorService executor;
 
     public CassDefinitionsServiceImpl() {
     }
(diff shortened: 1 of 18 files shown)


lucasponce:
@jshaughn this is ready for review from my side.
I've added a last commit as the examples espaced from my initial review.

jshaughn:
> The use case of what to do when a plugin is undeployed is not covered, now it maintains the action definitions but it will fail at the moment of execution of the action. Perhaps a nice to have is to mark some state, but then we need to add a way to detect "undeployments" of .war, so for the moment I think that could be a cornercase and it is good as it is working today.

+1, I don't think we need to worry too much about plugins disappearing.  A validation at startup would be a reasonably sufficient validation, if we don't do that now we would do that later.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
```diff
@@ -225,6 +129,16 @@ public void addActionDefinition(String tenantId, ActionDefinition actionDefiniti
...
         }
+        String plugin = actionDefinition.getActionPlugin();
+        if (!getActionPlugins().contains(plugin)) {
+            throw new IllegalArgumentException("Plugin: " + plugin + " is not deployed");
+        }
+        Set<String> pluginProperties = getActionPlugin(plugin);
+        actionDefinition.getProperties().keySet().stream().forEach(property -> {
+            if (!pluginProperties.contains(property)) {
+                throw new IllegalArgumentException("Property: " + property + " is not valid on plugin: " + plugin);
+            }
```
jshaughn: I'm not sure if throwing an exception is too strong, perhaps a warning log message would be sufficient, but I am OK with it.  I can't recall, can action plugins declare required properties?  If so, I don't see a check to ensure required properties are provided, only whether provided properties are valid.

lucasponce: Plugins declare which are supported properties.
This check is to ensure that a property defined in an action are present in the plugin.
No more validation defined (types or values).
I think an exception is ok to force to use correct properties, if not validation has not sense at all (I think).
The REST API is modified, so, when no properties defined, a "Bad request" response is sent instead of "Internal error".
Perhaps in the future we can add validation at value level, but that is more complex as it might required more changes in the API and more communication between plugins are engine.

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java
```diff
@@ -451,6 +457,9 @@ public Response updateTrigger(
 
         } catch (Exception e) {
             log.debug(e.getMessage(), e);
+            if (e.getCause() != null && e.getCause() instanceof IllegalArgumentException) {
+                return ResponseUtil.badRequest("Bad arguments: " + e.getMessage());
+            }
```
jshaughn: I know it may be tedious, but are there other services where we should now be returning "Bad arguments" for consistency?

jshaughn:
Lucas, only a few comments, this may be ready for merge as-is.  If you decide to make changes that is fine too. Let me know if you want me to merge...

hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java
lucasponce: It's not tedious. I think it's harmless to extend same discrimination in the catch to other areas.
I will send a fix.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
jshaughn: Right, I understand this check and am OK with an exception.  I was pointing out that the check does not ensure that all [necessary or required] properties are supplied, only that the supplied properties are supported.  It's not clear to me whether an action can pass validation yet fail execution because it did not supply sufficient properties.

lucasponce: Yes. Required parameters are not validated as that concept is not yet introduced in the plugins architecture.

jshaughn: OK, no problem.  What we have here is still a good improvement.

PR #178, opened 2016-05-05
HWKALERTS-145 Remove Hawkular Accounts dependency

lucasponce:
Preparing Hawkular Alerts for Hawkular Core Services.
No more dependency on Hawkular Accounts, tenant will be explicitly passed as a parameter (like in the standalone profile).
Fixing the itest to use hawkular-nest-feature-pack to build the server to validate integration tests.
Minor changes on the way that internal ISPN caches are accessible from JNDI.
Please @jshaughn, could you take a look if there is some missing change for this one ?
Thanks


pom.xml
```diff
@@ -93,7 +93,7 @@
     <version.org.drools>6.3.0.Final</version.org.drools>
     <version.org.freemarker>2.3.17</version.org.freemarker>
     <version.org.hawkular.accounts>2.0.26.Final</version.org.hawkular.accounts>
```
jshaughn: Do we still need this?  I see it referenced in a few poms as well.

lucasponce: +1
Good catch, it would be good to clean the pom.

PR #180, opened 2016-05-10
Increase waiting time for C* in itest

lucasponce:
Main issue with ITests is that the server needs to initialize Cassandra scheme.
Sometimes (randomly) the first test is executed before the scheme is created.
This also might depends of several factors of the machine where tests are running.
So, I've tested several ways, but all of them are focus on increase the waiting time of the tests until everything is ready on the server.
This PR also increase the SMTP server to avoid some annoying but harmless errors on log.

### DIFF
--- a/hawkular-alerts-rest-tests/pom.xml
+++ b/hawkular-alerts-rest-tests/pom.xml
@@ -449,7 +449,7 @@
             <artifactId>maven-antrun-plugin</artifactId>
             <configuration>
               <tasks>
-                <sleep seconds="5" />
+                <sleep seconds="45" />
               </tasks>
             </configuration>
             <executions>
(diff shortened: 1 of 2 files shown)


tsegismont:
Lucas, while I'm sure the change will solve your problem, perhaps a more
reliable fix would be to change Alerts startup sequence. In Metrics, schema
setup is also async, but there is a flag which makes the startup bean wait
until setup is complete. So Wildfly then waits before marking the
deployment as complete as well. And the maven surefire/failsafe plugin
never starts the suite before everything is OK.

lucasponce:
@tsegismont Yes, a review on the scheme is on our TODO list and also additional checks.
This is a temporal workaround to not block other priorities tasks.

lucasponce:
@tsegismont I've saw the logic in metrics.
Yes, in a following PR I will incorporate a similar logic into the status endpoint.
It's a good suggestion.
Thx

jshaughn:
I'll merge this, will we remember to change the approach later or should we maybe add a jira...

PR #181, opened 2016-05-10
Add global tenant validation on REST endpoints

lucasponce:
Improve common SMTP used on itests

SMTP server for itest can be improved if started shared for all tests in a suite.
This helps to reduce the waiting time.
Also, using a suite will help in future itest improvements, as at the beginning of the suite the waiting of cassandra scheme can be done.
Using a suite is compatible to exucute separate tests when needed.


jshaughn:
LGTM, merging

PR #184, opened 2016-05-18
Fix pom versions.

jshaughn:
The pom update somehow did not happen after the 1.0.2.Final release.  Update to a snapshot of next release version.

@lucasponce Please review and merge.

### DIFF
--- a/hawkular-alerts-actions-api/pom.xml
+++ b/hawkular-alerts-actions-api/pom.xml
@@ -23,7 +23,7 @@
   <parent>
     <groupId>org.hawkular.alerts</groupId>
     <artifactId>hawkular-alerts</artifactId>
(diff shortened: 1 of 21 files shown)


abonas:
not related directly to the change in this PR (and should not block its merge), but why are the alerts/alerts plugin artifact definitions repeating themselves as is many times and not inherited? (via dependencyManagement/dependency sections)

lucasponce:
I am not sure if I fully understood your question @abonas.
In this PR the change is in the parent section, in hawkular we define the artifact version explictly in the parent.
dependencyManagement/dependency are used for regular dependencies.
In this case, there are several repeating changes as each plugin is an independent artifact.

lucasponce:
LGTM

PR #185, opened 2016-05-18
Update commons to 0.7.1.Final

jshaughn:
Update parent to 43

@lucasponce, please review.  NOTE: this includes the pom version update, so if you merge, merge after that one is merged.

PR #186, opened 2016-05-19
Update commons to 0.7.1.Final

lucasponce:
Update parent to 43

Similar as #185 but from a rebased master.


lucasponce:
I have added a minor commit to re-arrange the way PersistenceTest waits on a couple of async results.
This helps travis to not fail and not slow down tests on faster machines.

PR #192, opened 2016-06-21
HWKALERTS-152 Update examples to work with hawkular-services

lucasponce:
Minor PR to make compatible Hawkular Alerting examples with hawkular-services distribution.
Standalone scenario remains as it was but now examples can be easy executed against hawkular-services backend (helping to test the component).

Please @jshaughn, could you take a look ?
Thanks.


jshaughn:
Ignoring irrelevant Travis failure.  Lucas, looks good, thanks for keeping things up to date and for pointing out that unix-specific command.

PR #195, opened 2016-08-01
Add additional Timeout parameters for C* driver at configuration level

lucasponce:
Adding fine control on C* driver parameters.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
@@ -29,6 +29,7 @@ import com.datastax.driver.core.Cluster;
 import com.datastax.driver.core.KeyspaceMetadata;
 import com.datastax.driver.core.ProtocolVersion;
 import com.datastax.driver.core.Session;
+import com.datastax.driver.core.SocketOptions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.CharStreams;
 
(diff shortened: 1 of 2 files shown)


lucasponce:
Hi @jshaughn, 
Please, could you take a look on it ?
The root cause is a slow Cassandra, there are space for future long term optimizations (schema review, more async work in the engine) but for now I think that adding possibility to fine control the timeouts for these environments is good.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
```diff
@@ -144,6 +149,11 @@ public static synchronized Session getSession(boolean overwrite) throws Exceptio
             String nodes = AlertProperties.getProperty(ALERTS_CASSANDRA_NODES, ALERTS_CASSANDRA_NODES_ENV, "127.0.0.1");
             int attempts = Integer.parseInt(AlertProperties.getProperty(ALERTS_CASSANDRA_RETRY_ATTEMPTS, "5"));
             int timeout = Integer.parseInt(AlertProperties.getProperty(ALERTS_CASSANDRA_RETRY_TIMEOUT, "2000"));
+            int connTimeout = Integer.parseInt(AlertProperties.getProperty(ALERTS_CASSANDRA_CONNECT_TIMEOUT,
+                    ALERTS_CASSANDRA_CONNECT_TIMEOUT_ENV, "5000"));
+            int readTimeout = Integer.parseInt(AlertProperties.getProperty(ALERTS_CASSANDRA_READ_TIMEOUT,
+                    ALERTS_CASSANDRA_READ_TIMEOUT_ENV, "12000"));
```
jshaughn: Are these the datastax defaults as well, or are you setting them differently from out-of-box settings?  Perhaps by default we should not override the driver?

lucasponce: These are the default values of the driver.
I think it is safe.
I wanted to avoid something like checking the parameters for cleanness code.
Perhaps a comment can help here.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
```diff
@@ -144,6 +149,13 @@ public static synchronized Session getSession(boolean overwrite) throws Exceptio
             String nodes = AlertProperties.getProperty(ALERTS_CASSANDRA_NODES, ALERTS_CASSANDRA_NODES_ENV, "127.0.0.1");
             int attempts = Integer.parseInt(AlertProperties.getProperty(ALERTS_CASSANDRA_RETRY_ATTEMPTS, "5"));
             int timeout = Integer.parseInt(AlertProperties.getProperty(ALERTS_CASSANDRA_RETRY_TIMEOUT, "2000"));
+            /** @see SocketOptions.DEFAULT_CONNECT_TIMEOUT_MILLIS */
+            int connTimeout = Integer.parseInt(AlertProperties.getProperty(ALERTS_CASSANDRA_CONNECT_TIMEOUT,
+                    ALERTS_CASSANDRA_CONNECT_TIMEOUT_ENV, "5000"));
+            /** @see SocketOptions.DEFAULT_READ_TIMEOUT_MILLIS */
+            int readTimeout = Integer.parseInt(AlertProperties.getProperty(ALERTS_CASSANDRA_READ_TIMEOUT,
+                    ALERTS_CASSANDRA_READ_TIMEOUT_ENV, "12000"));
```
jshaughn: OK, as we discussed in IRC, let's not override (keep the datastax OOB settings) unless our sysprops have been set.  After that this looks good to me.

jshaughn:
Lucas, Looks good although I'd recommend adding a _MILLIS and -millis suffix on the property and env var.

lucasponce:
I have added additional comments to clarify that the units are milliseconds.
But I wouldn't change the variables names for consistency.
We had these variable documented some time ago in the (now broken) doc:
https://github.com/hawkular/hawkular.github.io/blob/pages/src/main/jbake/content/docs/components/alerts/index.adoc

PR #197, opened 2016-08-16
HWKALERTS-156 Process type from bus messages

lucasponce:
I have created a preliminar PR to address HWKALERTS-156.
This one depends on https://github.com/hawkular/hawkular-metrics/pull/568 on hawkular-metrics project.
I can use this PR to add the commands orders, this will be sent later but I wanted to create the PR to start reviewing it.


lucasponce:
Please @jshaughn, could you review this one in the context of metrics work ?
This is the counterpart needed to send the publish command to the JMS messaging.
Once metrics received, it will start sending requested data.

There are end-to-end integration tests missing that should be placed at higher level (hawkular-services) (when this design is approved and merged).

hawkular-alerts-bus-api/src/main/java/org/hawkular/alerts/bus/api/MetricDataMessage.java
```diff
@@ -86,6 +86,8 @@ public String toString() {
      */
```
jshaughn: It might be nice if Alerting and Metrics could share classes for this messaging.  I don't think metrics makes the message classes public at the moment...

hawkular-alerts-bus-api/src/main/java/org/hawkular/alerts/bus/api/PublishCommandMessage.java
```diff
@@ -0,0 +1,164 @@
...
+package org.hawkular.alerts.bus.api;
+
+import java.util.List;
+
+import org.hawkular.bus.common.AbstractMessage;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * A bus message that will indicate which metrics will be published/unpublished to the bus.
```
jshaughn: As mentioned in the other PR, could we communicate via an ISPN cache as opposed to publishing messages for the ids to publish?

hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/init/CacheManager.java
```diff
@@ -140,4 +167,49 @@ private synchronized void updateActiveIds() {
...
+        } else {
+            if (log.isDebugEnabled()) {
+                log.debug("DataId " + dataId + " doesn't have a valid metrics type. Skipping publishing command.");
+            }
+            return;
+        }
+        PublishCommandMessage msg = new PublishCommandMessage(PUBLISH_COMMAND, tenantId,
+                asList(new PublishCommandMessage.MetricKey(type, metricId)));
+        publishCommandSender.send(msg);
+    }
```
jshaughn: OK, so it looks like what you've proposed is prefixing the dataId with the metricType so that the TenantId+DataId is unique, right?   That seems reasonable.  It's certainly important for Trigger generation, but since that's basically being done by you in MIQ, you're the only one that has to know what the dataIds need to look like.  A couple of things, I don't think you need to prefix Avail dataIds because the type is implicit.

You could possibly make the GAUGE prefix optional and consider it a default, because it is the most common.

hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/listener/AvailDataListener.java
```diff
@@ -85,11 +86,12 @@ protected void onBasicMessage(AvailDataMessage msg) {
         List<Data> alertData = null;
         Set<String> activeAvailabilityIds = cacheManager.getActiveAvailabilityIds();
         for (SingleAvail a : data) {
-            if (isNeeded(activeAvailabilityIds, a.getId())) {
+            String fullMetricId = AVAILABILITY + "-" + a.getId();
```
jshaughn: As mentioned, I don't think you need this prefix.

stefannegrea:
I've been looking in this PR for the code that requests metrics to be published. Can you explain how that process works, I see the new code in CacheManager but I am not sure how/when is activated. Do you mind explaining?

lucasponce:
CacheManager stores a cache of the dataId currently active on alerting definitions.
Via alerting listeners is synced internally to have this info for filtering which potential data is interesting for alerting.
In the previous schema, having all metrics data on the JMS topic, it was used mainly for filtering which data is interesting for alerting.
Now, this logic is enhanced in the sense that when a new dataId is present, a publish command is sent.
This publish command is processed by metrics to start fordwarding the relevant data into the JMS topic.

hawkular-alerts-bus-api/src/main/java/org/hawkular/alerts/bus/api/MetricDataMessage.java
lucasponce: +1
the package https://github.com/hawkular/hawkular-alerts/tree/master/hawkular-alerts-bus-api was created for that.
I agree that at some point it could go to a commons repo to have more clean maven dependencies.

hawkular-alerts-bus-api/src/main/java/org/hawkular/alerts/bus/api/PublishCommandMessage.java
lucasponce: Yes.
The ISPN cache proposed as storage of the ids
https://github.com/hawkular/hawkular-metrics/pull/568/files#diff-c8641cf1ba58303ba15fba81de0f0c1aR40
can be updated externally.
That would save logic.

hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/init/CacheManager.java
lucasponce: Yes, I know what you mean.
I decided to make them explicit because I think in easy to understand the logic, even if some prefix like "availability" can be deducted from the wrapped message type.
I think make them explicit helps to maintain them in the future without paying a lot.

stefannegrea:
I do not see any unpublish commands being sent from Alerts to Metrics. What is the expectation there? Once you need something to be published it always needs to be published?

What if somebody else sends an unpublish command to Metrics for a metricid relevant to Alerts? What mechanisms do you have in place to correct that?

With your proposed changes, what happens after an application container restart or crash? Can you please go into detail because it is not obvious from the two PRs (Metrics and Alerts)?

lucasponce:
>I do not see any unpublish commands being sent from Alerts to Metrics. What is the expectation > > there? Once you need something to be published it always needs to be published?

The expectation here is to add the unpublish command when that dataId is not active in alerting.
That is not part of the current PR yet.

> What if somebody else sends an unpublish command to Metrics for a metricid relevant to Alerts? What mechanisms do you have in place to correct that?

Could you elaborate "somebody else" and the time frame ? 
In this PR that use case is not contemplate  yet.
In that case the table that store the info would be extended to handle the info.

> With your proposed changes, what happens after an application container restart or crash? Can you please go into detail because it is not obvious from the two PRs (Metrics and Alerts)?

The main table is backed with an ISPN cache, which via configuration it can be extended to have a data-store. You can take as an example the standard mechanism used by Wildfly to store web sessions / ejb that prevents that info is restored after a crash.
For the use case proposed I think is enough, of course, if we want to add additional usecases, this one can be revisit.

stefannegrea:
Based on your reply and current implementation the cache on the Metrics side has unbounded growth with no possibility to adjust it.If we add persistence to that mix that will be even more problematic.

lucasponce:
Reopened to follow @jshaughn suggestions to share a ISPN between both components.
A following commit will update that on this PR.

hawkular-alerts-bus-api/src/main/java/org/hawkular/alerts/bus/api/MetricDataMessage.java
jshaughn: I think what would be best would be to bypass commons and just have metrics make it public.

hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/init/CacheManager.java
```diff
@@ -46,98 +51,236 @@
...
 
-    Set<String> activeDataIds;
-    //long activeDataIdsTime = 0L;
-    Set<String> activeAvailabityIds;
+    private final String AVAILABILITY = "availability";
+    private final String GAUGE = "gauge";
+    private final String COUNTER = "counter";
+    private final String COUNTER_RATE = "counter_rate";
+    private final String STRING = "string";
+    private final String GAUGE_RATE = "gauge_rate";
```
jshaughn: As discussed in the other PR, I think these constants should be more unique to avoid conflict with non-hmetrics dataids.

hawkular-alerts-bus/src/main/java/org/hawkular/alerts/bus/init/CacheManager.java
```diff
@@ -46,98 +51,236 @@
...
+            }
+            return null;
+        }
+        return new MetricId(tenantId, type, metricId);
+    }
+
+    public static class DataIdKey {
+        private String tenantId;
+        private String dataId;
+
```
jshaughn: perhaps just store the metricId in here as another field and strip out the prefix in the constructor.

jshaughn:
Aside from the few comments, this is looking good to me.

lucasponce:
@stefannegrea @jshaughn on this PR additional logic has been added to:
- enable/disable whole publishing mechanism.
- clear the cache at restart of the component. 
(Cache was always re-created at restart but not cleared, so, now any potential corrupted data should be clean).
Please, could you review and check if there is any missing topic to cover ?
Thanks

lucasponce:
As commented on metrics PR, once this one is merged, at top level Hawkular Services we need to define the configuration for the shared ISPN.
Also we should add some integration tests there as we can't test the full cycle within components.

PR #198, opened 2016-08-16
Force English locale in mail decimal formatting

jotak:
Locale should be forced to English (rather that system default) since the whole mail is written in English.
Without it, we could end up with mixed languages/formatting such as "GC Duration greater than 1000,0 ms" with wrong decimal separator.

Note that the issue was seen from JUnit EmailTemplateTest:

Expected :Tiny template: Alert [open] message: GC Duration greater than 1000,0 ms for App Server thevault~Local
Actual   :Tiny template: Alert [open] message: GC Duration greater than 1000.0 ms for App Server thevault~Local


lucasponce:
Hi @jotak this PR has failed with the license checker.
If you could review it and sync I can merge it.
More details on travis CI log.

PR #199, opened 2016-08-30
HWKALERTS-157 Process ConditionEval grouped per trigger

lucasponce:
This PR groups the processing of ConditionEval inside the rules engine.
For multiple conditions triggers there are some corner cases that may lead to false satisfied status.

For example, given the following trigger:

	Trigger1. Matching ANY
		Condition1: HeapUsed < 10 
		Condition2: HeapUsed > 20
		
	Default dampening

Incoming data:

t1.	HeapUsed = 9

	(CE = ConditionEval)
	
	CE: c2. HU (9) > 20 [False]
		** Update Dampening
			CurrentEvals: 1) None
						  2) 9 > 20 [F] 
	CE: c1. HU (9) < 10 [True]
		** Update Dampening
			CurrentEvals: 1) 9 < 10 [T]
						  2) 9 > 20 [F]
--> EVENT

t2. HeapUsed = 11

	CE: c2. HU (11) > 20 [False]
		** Update Dampening
			Current Evals: 1) 9 < 10 [T]
						   2) 11 > 20 [F]
						   
--> EVENT (INCORRECT)

t3. HeapUsed = 21

t4. HeapUsed = 19

Or this other example:

	Trigger1. Matching ALL
		Conditions 
			"myapp.war" text == 'DOWN'
			"datacenter1" text starts 'ERROR'
			"datacenter2" text starts 'WARN'
			
	Default dampening
	
			
Incoming data:

t1. "myapp.war" "DOWN"
t2. "datacenter1" "ERROR [Time] This is a sample as app logging"
t3. "datacenter2" "WARN [Time] This is a sample as app logging"

--> EVENT

t4. "myapp.war" "UP"
t5. "datacenter1" "ERROR [Time] This is a sample as app logging"
t6. "datacenter2" "WARN [Time] This is a sample as app logging"

--> EVENT (INCORRECT)

t7. "myapp.war" "UP"
t8. "datacenter1" "ERROR [Time] This is a sample as app logging"
t9. "datacenter2" "WARN [Time] This is a sample as app logging"

--> EVENT (INCORRECT)

So, this PR introduces a grouping rule in the engine, in a way that ConditionEval are evaluated in a single step, instead to a sequencial way interacting with the working memory.

As a result, the previous behaviour	is maintained (all tests previous tests are ok with this change) but we cover these cornercases with multiple conditions.

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/TriggerConditionEval.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/TriggerConditionEval.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.alerts.api.model.condition;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.hawkular.alerts.api.model.trigger.Mode;
+
+/**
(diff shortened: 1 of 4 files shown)


jshaughn:
@lucasponce  I'd like to talk about this on a call.   I can see the issue in the first, ANY trigger, example.  That's a good description of a faulty corner case.  I may need some more explanation for the second example, the AND trigger (did you mean ALL?).  I'm not sure I understand this one yet, and the updated test code did not add a test for this, just a big comment, so I'm a little bit confused.

Certainly you have found an issue, I want to make sure I understand it completely before further evaluating the proposed solution.

lucasponce:
@jshaughn perfect. Yes, in the second example I mean ALL, I've updated the informal description.
The second example is a draft of another faulty scenario.
Let's have a call to analyze them and move forward.

PR #200, opened 2016-09-02
HWKALERTS-157 Proposed Enhancements

jshaughn:
- Optimize the single-condition Trigger or single ConditionEval use cases
  - This common use case deserves streamlined processing.
  - Added some more comments in .drl to better describe the HA-157 problem use case
    and provide an explanation of the rule additions
- Changes to TriggerConditionEval
  - rename to TriggerConditionEvalSet for clarity
  - repackage to org.hawkular.alerts.api.model.dampening as this is used
    specfically for dampening reasons (actually, it would be nice if we
    could keep this out of api but I don't think we can.)
  - removed unnecessary 'mode' attribute
- .drl logging (tangential, sorry)
  - changed to log.debugf for clarity
  - ensured debug logging all rule firings

Performance Note
--------------------
I did some simple perf testing by running  PerfRulesEngineTest+RulesEngineTest 3 times each for the current version in master and this revision.  The perf gain was 5.13% using the code optimized for the single condition eval.

@lucasponce For your consideration, please review at your convenience.


hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -314,44 +317,81 @@ rule ProvideDefaultDampening
...
+        retract( $ce2 )
+
+        TriggerConditionEvalSet tces = new TriggerConditionEvalSet( $tenantId, $tid, $ce1, $ce2 );
+        insert( tces );
+
+        log.debugf("ProvideTriggerConditionEvalSet %s", tces);
 end
 
-rule DampenTrigger
+rule AddToTriggerConditionEvalSet
```
lucasponce: I am not sure about this one.
A potential problem could happen:
- In a situation where we have a large # of conditions, imagine a (far cornercase but possible) trigger with 5 conditions.
- TriggerConditionEvalSet is inserted in "ProvideTriggerConditionEvalSet" but populated in "addToTriggerConditionEvalSet", and the object in the working memory is not updated.
- At the point is inserted "DampenTriggerMultiEval" is chosen to fire, but there is a potential race where you can fire "DampenTriggerMultiEval" before that all conditions eval have been populated.

The previous "CollectConditionEvals" populate the TriggerConditionEval atomically, so that race situation is addressed.

Certainly, having an improvement for the single ConditionEval situation is good, as we know that multiconditions are less used.

Perhaps these rules can be arranged mixing both approaches, having a case for single ConditionEval, but in the case of multi-conditions, using a rule to accumulate in a single step, not using a strategy of separate rules, perhaps that could work with salience attributes, but I'd prefer if we could avoid them and just using an accumulate sentence for this case.

jshaughn: I don't think there is an issue because addToTriggerConditionEvalSet should match repeatedly due to each firing being against a different ConditionEval.  But I'll take a look now to see if a combined approach may be more clear while maintaining a performance gain.

lucasponce: At the moment a TriggerConditionEvalSet is inserted, the rule "DampenTriggerMultiEval" is activated.
So, there can be a race between the update of the TriggerConditionEvalSet within "addToTriggerConditionEvalSet" and "DampenTriggerMultiEval".
So, to be safe, control flow mechanism should be needed, but I guess if the collect can be done in a single step, that would be more clear from a high level perspective.
Thanks !

lucasponce:
These last changes are looking good.
I am going to wait Travis and merge if everything is ok.

lucasponce:
Thanks @jshaughn, it was an interesting review of all of these topics.

PR #201, opened 2016-09-07
HWKALERTS-158 Prefix properly data back from the bus on HM integration

lucasponce:
Please @jshaughn, can you review this patch ?
I will prepare a srcdeps with hawkular-services to fix & validate everything is working as expected.
Then we can release alerting and hawkular-services.
Thanks.

### DIFF
--- a/hawkular-alerts-bus-api/pom.xml
+++ b/hawkular-alerts-bus-api/pom.xml
@@ -46,6 +46,12 @@
       <version>${project.version}</version>
     </dependency>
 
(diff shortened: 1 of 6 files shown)


lucasponce:
I have found some strange behaviour that needs an additional commit.

lucasponce:
Fixing some logging levels to debug and missing @Startup annotation on CacheManager.

jshaughn:
LGTM.

PR #203, opened 2016-09-08
Upgrade MetricsId serialized class for ISPN uses

lucasponce:
It seems that ISPN can handle non-serialized objects.
This has created some problems when using MetricId as a key on ISPN between Metrics and Alerting.
A preliminar workaround was to use a string type as a key, but a better approach is to use a MetricId which implements Serialized.
This PR is on hold until we can test that a serialized MetricId class is working fine in metrics.
Then we will use that for key on published cache.


lucasponce:
As finally the MetricId class is not modified, we don't need to wait until hawkular-metrics releases a 0.19.x.
We can send this PR and release later (for alerting).
For Hawkular Services we need new release of Hawkular Metrics.
/cc @jshaughn

PR #204, opened 2016-09-14
HWKALERTS-160 Change CompareCondition Handling

jshaughn:
- Now allows for the relevant Data to be reported at different times.
- Add new test and update existing test to use distinct times
- Also, null-protect some log stmts (although I'm not sure when log may
  be null, perhaps in some test envs)


lucasponce:
Changes proposed look good.
Thanks @jshaughn .

PR #205, opened 2016-09-16
Metrics w/ Alerting

jshaughn:


jshaughn:
This is for Review Only at this point and Goes hand in hand with PRs for Hmetrics and Hservices.

lucasponce:
A general comment before to enter in code's details.
This branch is large and I guess that is touching a lot of things.
After review in the code I think is better to not merge as it is and group functionality in small PRs for sanity.
I know is extra work but I think is better for clarity and sustainability of the code.

hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/PluginMessageDescription.java
```diff
@@ -477,7 +476,7 @@ public String compare(CompareCondition condition) {
      * @return a description to be used on email templates
      */
     public String external(ExternalCondition condition) {
-        String description = "SystemId: " + condition.getSystemId();
+        String description = "AlerterId: " + condition.getAlerterId();
```
lucasponce: Is this just a renaming of systemId -> alerterId or it is introducing a new concept.
(I guess that this will be showed below in the code, but this was on the first modified files showed in the PR).

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/json/JacksonDeserializer.java
```diff
@@ -374,8 +374,8 @@ public static Condition deserializeCondition(JsonNode node) throws JsonProcessin
                 try {
                     condition = new ExternalCondition();
                     ExternalCondition eCondition = (ExternalCondition) condition;
-                    if (node.get("systemId") != null) {
-                        eCondition.setSystemId(node.get("systemId").textValue());
+                    if (node.get("alerterId") != null) {
+                        eCondition.setAlerterId(node.get("alerterId").textValue());
```
lucasponce: Same here

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalCondition.java
```diff
@@ -37,11 +37,11 @@
...
 
     /**
-     * An identifier assigned by the external system to identify this condition as being handled by that
-     * external system. It should be unique enough such that externalSystemIds are unique.
+     * An identifier assigned by the external alerter to identify this condition as being handled by that
+     * alerter. It should be unique enough such that external AlerterIds are unique.
      */
     @JsonInclude(Include.NON_NULL)
-    private String systemId;
+    private String alerterId;
```
lucasponce: What is the motivation of terminology change ?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/data/CacheKey.java
```diff
@@ -0,0 +1,101 @@
...
+package org.hawkular.alerts.api.model.data;
+
+import java.io.Serializable;
+
+public class CacheKey implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private String tenantId;
+    private String dataIdPrefix;
+    private String dataIdSuffix;
```
lucasponce: What is the logic for storing suffix on the dataId ?
Is this CacheKey related to the publishing workflow ?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -0,0 +1,312 @@
...
+        try {
+            CacheSet<CacheKey> currentlyPublished = publishCache.keySet();
+
+            log.debugf("Published before update=%s", currentlyPublished.size());
+            if (log.isTraceEnabled()) {
+                publishCache.entrySet().stream().forEach(e -> log.tracef("Published: %s", e.getValue()));
+            }
+
+            // This will include group trigger conditions, which don't need to be published but it's
+            // not easy to efficiently filter them out.
```
lucasponce: Good point.
Assuming as our main users (MiQ from MW and CM OPS) are going to use Group Triggers perhaps it's good to include it.
The updateActiveIds() is not as critical in time as other task, so I guess it is feasible to fetch all triggers and to build a list of group triggers ids where you can use to filter the conditions below in the code.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -0,0 +1,312 @@
...
+            String suffix = null;
+            for (String p : DATA_ID_PREFIXES) {
+                if (dataId.startsWith(p)) {
+                    prefix = p;
+                    suffix = dataId.substring(p.length());
+                    break;
+                }
+            }
+            if (null != prefix) {
+                this.cacheKey = new CacheKey(tenantId, prefix, suffix);
```
lucasponce: Ok, this explains my comment below about suffix.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -0,0 +1,312 @@
...
+            return;
+        }
+    }
+
+    private void publish(DataIdKey dataIdKey) {
+        CacheKey cacheKey = dataIdKey.getCacheKey();
+        if (cacheKey != null && publishCache != null) {
+            // This logic assumes that alerting is the only writer for the shared publishCache
+            log.debugf("Publishing:%s ", cacheKey);
+            publishCache.put(cacheKey, dataIdKey.getDataId());
```
lucasponce: If we put the CacheKey class in ISPN, we need to share classloader between consumer (metrics) and producer.
I guess this was already commented, but just leaving a note here.

lucasponce: And also the CacheKey class should be a dependency on consumer (metrics) which I guess dependencies are from metrics now, not on the contrary as we had in the previous code to use the MetricId class.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/util/CompareData.java
```diff
@@ -0,0 +1,74 @@
...
+import org.hawkular.alerts.api.model.data.Data;
+
+/**
+ * CompareConditions often work against data arriving at different times. We convert Data into CompareData and
+ * apply special handling such that CompareData can live longer than one engine firing.
+ *
+ * @author Jay Shaughnessy
+ * @author Lucas Ponce
+ */
+public class CompareData {
```
lucasponce: I think CompareData was already on master.
Perhaps this branch is not sync-ed ?

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -106,25 +107,70 @@ rule ThresholdRange
...
+// CompareCondition is a bit different because it involves two dataIds in one condition evaluation.  For example,
+// Heapused > 80% Heapmax.  The original implementation required that the two necessary Data facts (Heapused and
+// Heapmax) both be supplied in the same engine firing.  This proved to be an unrealistic constraint as, we needed
+// to allow Heapused and Heapmax Data facts to be supplied on different schedules.  The updated semantic is to
+// perform CompareCondition evaluations using the two most recent values for the relevant dataIds, holding data
+// across engine firings as needed.  So, if Heapused is reported every 30 seconds and Heapmax every 5 minutes, we'll
+// hold and update Heapused until we have a Heapmax value for comparison.  After that we'll update each as they
+// arrive, performing comparisons on each update.  We use CompareData as the special Fact to hold relevant data
+// across engine firings.
+rule ProvideCompareData1
```
lucasponce: Same comment.

hawkular-alerts-metrics/src/main/java/org/hawkular/alerts/external/metrics/Expression.java
```diff
@@ -58,44 +60,22 @@
...
-        /** number of distinct [trait] values for the group. No period. */
-        card(false),
-        /** percent [0.0,1.0] with last-reported availability down. No period */
-        down(false),
-        /** percent [0.0,1.0] with last-reported availability up. No period */
-        up(false);
-
-        private final boolean single; // true if supported by single metric/ requires a period
+        /** Count of UP avail for the period. */
+        heartbeat;
```
lucasponce: I don't follow these changes.
Perhaps this is related to the changes on external alerter logic commented in some meeting.

hawkular-alerts-metrics/src/test/groovy/org/hawkular/alerts/external/ExternalMetricsITest.groovy
```diff
@@ -620,6 +621,100 @@ class ExternalMetricsITest extends AbstractExternalITestBase {
...
+        // Tag the trigger as a HawkularMetrics:MetricsCondition so it gets picked up for processing
+        triggerTestHeartbeat.addTag("HawkularMetrics", "MetricsCondition" );
+
+        resp = client.post(path: "triggers", body: triggerTestHeartbeat)
+        assertEquals(200, resp.status)
+
+        // ADD external metrics heartbeat condition. Note: alerterId must be "HawkularMetrics"
+        // Average over the last 1 minute > 50, check every minute
+        ExternalCondition firingCond = new ExternalCondition("trigger-test-heartbeat", Mode.FIRING, "external-dataId-heartbeat",
+            "HawkularMetrics", "metric:1:heartbeat(" + availId + " <= 2),1");
```
lucasponce: This perhaps is a side comment, but I wonder if the uniqueness concerns on metrics with the new MetricId bringing the type could affect into the ExternalCondition logic, specially the expressions where I think is where ids are handled on this context.

hawkular-alerts-metrics/src/test/java/org/hawkular/alerts/external/metrics/ExpressionTest.java
```diff
@@ -37,8 +37,8 @@ public void expressionParseTest() {
         assertTrue(x.getThreshold() == 100.5);
         assertEquals(x.getMetric(), "foo");
 
-        x = new Expression("TAG:5m:avgd(bar <= 100),10");
-        assertEquals(x.getTarget(), Target.Tag);
+        x = new Expression("metric:5m:avgd(bar <= 100),10");
```
lucasponce: Same comment, I guess if how to distinguish if bar is counter/gauge.
I guess by default is read as gauge.

hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/PluginMessageDescription.java
jshaughn: This is just a renaming for clarity.  I had a TODO for this and decided to do it as part of reviving the metrics external alerter.  So, no need to worry too much about this refactor.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalCondition.java
jshaughn: Purely for clarity, it unifies the terminology of "External Alerter" and how that Alerter identifies itself  when defining its conditions.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/data/CacheKey.java
jshaughn: The CacheKey is the key used in CacheManager's ispn cache.  The prefix/suffix mechanism is an attempt to make the prefixing more generic, and a bit more configurable by the external client.  The set of acceptable prefixes (hm_g, hm_c, etc for Hm) can be configured externally.  I found this approach a little more flexible, you can see how it is used as you progress through the PRs.  We can talk about it more...

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
jshaughn: I was thinking about this more yesterday when looking at the addpendingData filtering and that logic, and actually I think we *do* want to include the group trigger data ids, and I think I want to update or remove this comment in the PR code.  The reason is data-driven group triggers.  I actually had forgotten about this sort-of-cool feature.  But if you recall (or go back and read the docs like I had to do :)  We can generate member triggers based on the "source" values provided  on incoming Data.  In this use case the dataIds *are the generic group trigger dataids*!  So we actually do want to process Data with the group trigger dataIds!

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
jshaughn: You are right about the classloader.  Because we are deployed as an EAR this works fine for the new metrics distribution.  And that is why I chose to avoid using Strings and stay with this CacheKey class that I already had in place when you discovered the classloader issue.  It could be a problem for other clients that want to filter but aren't packaged as an EAR.  But then again, we may be moving the filtering into alerting, in the sendData processing.

And you are right again, one of the ideas here was to remove Metrics dependencies in Alerting and have only the reverse, let Metrics depend on Alerting.  So don't use Hm's MetricId class as a key managed by Ha.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/util/CompareData.java
jshaughn: Right, I actually cherry-picked the PR for master from this branch, where I found the problem...  So this should disappear from the PR after another rebase, I think.  Even without a rebase it will not be a conflict.

hawkular-alerts-metrics/src/main/java/org/hawkular/alerts/external/metrics/Expression.java
jshaughn: Yes, this is a new feature for the metrics external alerter.  It is something I discussed with Stefan who was wondering whether we could do something like this.  See here, this is the updated external alerter doc for metrics:

https://docs.google.com/document/d/1vKV7ddPo4-LnHfP35jPoV0kSTLbhicCl1goldRzfSQA/edit#heading=h.kk5oqu4purxl

hawkular-alerts-metrics/src/test/groovy/org/hawkular/alerts/external/ExternalMetricsITest.groovy
jshaughn: This is a good point.  The reason it works today is because the external alerter for metrics assumes GAUGE metrics (unless its an avail-related expression).  Because the external alerter is in charge of evaluating the expression, we can make changes as needed.  One possibile enhancement may be to specify the metric type as the expression "target" (see the doc for what that means).  But considering this stuff is not yet used, we have time to make changes as needed.

lucasponce:
There are some errors on travis.
With the integration of hawkular-metrics it is needed to refactor to provide several artifacts at same time.
Standalone artifacts can be used for alerting itest meanwhile we can prepare a proper maven layout.
I am going to merge this one, to move things forward and working in the commented:
https://issues.jboss.org/browse/HWKALERTS-165
No logic changes expected, just accommodate the layout to simplify the use of conditional profiles as we have today to respond to the need of multiple artifacts we have.

lucasponce:
Note that until HWKALERTS-165 is done, travis will fail.
I count with that temporarily.

PR #206, opened 2016-09-19
HWKALERTS-161 Filter alerts by resolved and ack timestamps

lucasponce:
This PR covers:
- Store lifecycle history on alert.
- Extend alerts criteria to support filtering on resolved and ack timestamps.
- Include new changes on alerts criteria on REST handlers.

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
@@ -60,10 +60,10 @@ public class Alert extends Event {
     private String resolvedBy;
 
     @JsonInclude(Include.NON_EMPTY)
-    private List<Note> notes;
+    private List<Note> notes = new ArrayList<>();;
 
     @JsonInclude(Include.NON_EMPTY)
-    private List<LifeCycle> lifecycle;
+    private List<LifeCycle> lifecycle = new ArrayList<>();
 
     @JsonInclude(Include.NON_EMPTY)
     @Thin
@@ -88,6 +88,7 @@ public class Alert extends Event {
         this.status = Status.OPEN;
         this.severity = trigger.getSeverity();
         this.eventType = EventType.ALERT.name();
+        addLifecycle(this.status, "system", this.ctime);
     }
 
     @JsonIgnore
@@ -161,9 +162,6 @@ public class Alert extends Event {
     }
 
     public List<Note> getNotes() {
-        if (null == notes) {
-            this.notes = new ArrayList<>();
-        }
         return notes;
     }
 
@@ -185,9 +183,6 @@ public class Alert extends Event {
     }
 
     public List<LifeCycle> getLifecycle() {
-        if (null == lifecycle) {
-            this.lifecycle = new ArrayList<>();
-        }
         return lifecycle;
     }
 
@@ -202,7 +197,8 @@ public class Alert extends Event {
         getLifecycle().add(new LifeCycle(status, user, stime));
     }
 
-    public LifeCycle lastLifecycle() {
+    @JsonIgnore
+    public LifeCycle getCurrentLifecycle() {
         if (getLifecycle().isEmpty()) {
             return null;
         }
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
@@ -63,6 +63,9 @@ public class Alert extends Event {
     private List<Note> notes;
 
     @JsonInclude(Include.NON_EMPTY)
+    private List<LifeCycle> lifecycle;
+
+    @JsonInclude(Include.NON_EMPTY)
     @Thin
     private List<Set<ConditionEval>> resolvedEvalSets;
 
@@ -181,6 +184,31 @@ public class Alert extends Event {
         getNotes().add(new Note(user, text));
     }
 
+    public List<LifeCycle> getLifecycle() {
+        if (null == lifecycle) {
+            this.lifecycle = new ArrayList<>();
+        }
+        return lifecycle;
+    }
+
+    public void setLifecycle(List<LifeCycle> lifecycle) {
+        this.lifecycle = lifecycle;
+    }
+
+    public void addLifecycle(Status status, String user, long stime) {
+        if (status == null || user == null) {
+            throw new IllegalArgumentException("Lifecycle must have non-null state and user");
+        }
+        getLifecycle().add(new LifeCycle(status, user, stime));
+    }
+
+    public LifeCycle lastLifecycle() {
+        if (getLifecycle().isEmpty()) {
+            return null;
+        }
+        return getLifecycle().get(getLifecycle().size() - 1);
+    }
+
     @Override
     public String toString() {
         return "Alert [alertId=" + id + ", status=" + status + ", ackTime=" + ackTime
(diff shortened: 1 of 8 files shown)


lucasponce:
Please @jshaughn could you review it ? Thanks.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
```diff
@@ -181,6 +184,31 @@ public void addNote(String user, String text) {
         getNotes().add(new Note(user, text));
     }
 
+    public List<LifeCycle> getLifecycle() {
+        if (null == lifecycle) {
+            this.lifecycle = new ArrayList<>();
```
moolitayer: Is it possible to do this in a constructor and not on get?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
```diff
@@ -63,6 +63,9 @@
     private List<Note> notes;
 
     @JsonInclude(Include.NON_EMPTY)
+    private List<LifeCycle> lifecycle;
```
moolitayer: For an alert is it true that alert.status == alert.lastLifecycle.state? (and the same for resolvedTime & resolvedBy)
Would it be easier to return those from the last alert instead of having to maintain them separately?

cben:
Nice!

I'm surprised this doesn't tie in with the mechanism storing free-form `Note`s.
I suppose I was expecting a single log of (time, user, status, note) entries.

- That would also solve the current strangeness in returned `notes` array that ack/resolve notes are indistinguishable from regular notes.

OK, I can see how the `{start,end}{Resolved,Ack}Time` REST API doesn't care about adding notes that didn't change the status, so is naturally efficiently implemented in separate table.
=> Perhaps a more generic API filtering by any change in status OR note within a time window might make more sense for other users?
Example: imagine a "chatroom" style UI that wants to reflect all ongoing handling of multiple alerts — it would also want efficient polling to include added notes.

[Background: the `startResolvedTime` API was suggested in a discussion in ManageIQ how to efficiently poll specifically for auto-resolved alerts.  There we have the peculiar use case where other human actions —  ack, addNote — would all be handled inside ManageIQ, perhaps without telling hawkular.  So we only care about the auto-resolving which is done by hawkular.  But we also won't mind a wider API.]

lucasponce:
@cben originally, adding notes was not tied to a particular state.
Indeed, is some orthogonal to the state.
Ack or resolve actions add notes to an alert, but doesnt mean that any note represents a state change.
I.e. in the context of a discussion of an issue or similar, these notes were defined as general purpose.
Also, for "non-polling" mechanism inside hawkular alerting there is an Action mechanism that can be configured to implement the use case described (with more specific filtering or additional logic about what to do or what to filter).
These PR aims to increase the coverage on the filtering mechanism we have in place.
React to notes addition was not contemplated in the original design.

Also, the idea was not enter in some BPM scenario, the lifecycle mechanism is simple, but it's far of a proper processes/tasks assignation model.

We can comment offline your scenario and check how to map it within the existing implementation or what we can add into the existing model to cover it.

Thanks !

cben:
Sorry I wasn't clear.
For my purposes this PR is perfectly fine.
I just suspected that adding a status column to notes table — making it into note-and/or-status-change — *might* be simpler and cleaner model, and that I pushed you into a *possibly* more complicated direction by suggesting specifically `startResolvedTime`.
But if you find this direction natural and good for Hawkular, no objections from me.

lucasponce:
@cben I think that separate notes and lifecycle suits better in the current model.
A first point is that they are referring to different concepts (close and related, but different).
Other interesting point is that notes can be editable by the user, so we can add new notes or delete existing ones meanwhile the lifecycle will register the history of the changes.
Also a last one is just related to the non-sql modelling, that could be more complex to mix notes and states for filtering purposes.
Anyway, this feedback suggestion is good and we are collecting them for our future version 2.x, in that one we plan to review the model and we can revisit if a merge of notes+status can be more interesting.
These feedback has a lot of value for us.
Thanks Beni.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
jshaughn: Before continuing the review I guess I don't understand the motivation for adding LifeCycle to Alert.  Alert already had an ackTime and a resolvedTime, why not use these for the desired filtering?   We had made a conscious decision to keep lifecycle handling very light and I'm not aware of a need to support further lifecycle states.

lucasponce: Storing the lifecycle is needed for two reasons:
- In the model we might have open -> ack -> resolved -> ack -> resolved (transitions from resolved to ack states are valid).
- At the moment that we index resolved/ack dates for filtering with an auxiliar table we need to track the state to maintain this.
So the proper way was to store the transitions on the alert and it solves two purposes here, to show the history of the alert status and internally to help to index by these fields.

jshaughn:
@cben @lucasponce, yes good feedback, thanks.  One thing we've tried to do with Halerts (at least so far) is to try and limit what we do in the Alert Management space, with a preference for alert generation and a strong actions framework.  Because many users may have an alert-management tool in place, the idea is to be able to hand-off to those tools.  Having said that, notes are very likely to be made at lifecycle-change time, in fact the original Hawkular UI did exactly that.  So more formally tying a note to a lifecycle change is something for us to keep in mind, although we'll have to see whether the more simple running set of notes suffices.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
jshaughn: > In the model we might have open -> ack -> resolved -> ack -> resolved

I'm not aware of this being valid or at all common.  This sort of lifecycle support has not been a goal for us, and actually, I believe there is code to actively prevent going back to Ack from Resolved.

> At the moment that we index resolved/ack dates for filtering with an auxiliar table we need to track the state to maintain this.

I'm not sure I understand this comment.  Yes, we would have to index the fields but we already have that pattern for a variety of fields, no?   If we are to add Lifecycle then I wonder the same as @moolitayer, why keep the explicit ack/resolve fields?

lucasponce: > I'm not aware of this being valid or at all common. This sort of lifecycle support has not been a goal for us, and actually, I believe there is code to actively prevent going back to Ack from Resolved.
This was valid from the original Hawkular UI usecases. It was discussed in UX + BU meetings and we took it as valid. 
No, we don't prevent going back manually.
But with the MiQ intergation this logic was deferred in a less priority status.

> Yes, we would have to index the fields but we already have that pattern for a variety of fields, no?
Yes, but the pattern in place is not enough when the ack or resolved time fields may change. There is a need to "delete the last one" when maintaining the auxiliar indexes columns group.

> If we are to add Lifecycle then I wonder the same as @moolitayer, why keep the explicit ack/resolve fields?
This is a valid concern. As we introduced these fields, I don't want to delete them from the json object yet. Also there are some changes on the actions architecture that are related to this change.
So I think is a better/safer approximation to introduce these changes that basically affects to the filtering and remove those changes in a following iteration.

jshaughn: As long as you are confident that we allow only valid status transitions I guess I am OK with its allowance.  I do think we need to remove the redundant explicit fields for ack and resolved as it confuses/bloats the impl.  It is not required for the PR but there should be another Jira for this work, and the priority should be high.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
jshaughn: Should this initially be OPEN?  Or should LifeCycle really be more like StatusChange?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
lucasponce: > As long as you are confident that we allow only valid status transitions I guess I am OK with its allowance. I do think we need to remove the redundant explicit fields for ack and resolved as it confuses/bloats the impl. It is not required for the PR but there should be another Jira for this work, and the priority should be high.

Yes, +1 on that. It is not done on this PR due the implications on the Actions architecture, where legacy resolved and ack fields are propagated on several places (email templating included).

So, as we have planned to review the Actions architecture on this week for the metrics+alerting  effort, this removal will be done in that task.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
```diff
@@ -181,6 +184,31 @@ public void addNote(String user, String text) {
...
+    }
+
+    public void addLifecycle(Status status, String user, long stime) {
+        if (status == null || user == null) {
+            throw new IllegalArgumentException("Lifecycle must have non-null state and user");
+        }
+        getLifecycle().add(new LifeCycle(status, user, stime));
+    }
+
+    public LifeCycle lastLifecycle() {
```
jshaughn: Should this have a "get" prefix?  Also, would "Current" maybe be better than "Last"?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsCriteria.java
```diff
@@ -68,6 +72,50 @@ public void setEndTime(Long endTime) {
         this.endTime = endTime;
     }
 
+    public Long getStartResolvedTime() {
+        return startResolvedTime;
+    }
+
+    /**
+     * @param startResolvedTime fetched Alerts must have resolvedTime greater than or equal to startResolvedTime
```
jshaughn: I think we need more sematics here. What about alerts that have been resolved multiple times?  The same goes for the other start/end setters.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsCriteria.java
```diff
@@ -68,6 +72,50 @@ public void setEndTime(Long endTime) {
...
+    }
+
+    public Long getStartAckTime() {
+        return startAckTime;
+    }
+
+    /**
+     * @param startAckTime fetched Alerts must have ackTime greater than or equal to startAckTime
+     */
+    public void setStartAckTime(Long startAckTime) {
```
jshaughn: Ack is an optional status, we may need to add a note on what this returns if an alert went from open->resolved.  I guess this would not include that alert?

lucasponce:
Comments addressed and created new JIRAs to simplify the redundant fields in the context of the review of the Actions architecture.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
```diff
@@ -202,7 +197,8 @@ public void addLifecycle(Status status, String user, long stime) {
         getLifecycle().add(new LifeCycle(status, user, stime));
     }
 
-    public LifeCycle lastLifecycle() {
+    @JsonIgnore
+    public LifeCycle getCurrentLifecycle() {
         if (getLifecycle().isEmpty()) {
```
jshaughn: Now that LC is initialized to OPEN, I don't think this can be empty.

PR #207, opened 2016-09-19
HWKALERTS-162 Allow to not store external events

lucasponce:
This PR adds a new "/data" end point to the events section.
With this we can send external events to the engine without persisting them in a similar way as we do today for "/data" endpoint coming from metrics.
Events generated are persisted.
This work comes in the context of APM integration where the event is already persisted on their backend.


lucasponce:
Please @jshaughn, could you review this one ? Thanks.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
```diff
@@ -64,11 +64,14 @@
...
-     * Persist the provided events and send to the engine for alerts evaluation.
+     * Send events to the engind for alerts evaluation.
+     * Persist the provided events if persist flag is set to true.
+     *
      * @param events Set of unpersisted Events.
+     * @param persist if true events are persisted
      * @throws Exception any problem
      */
-    void addEvents(Collection<Event> events) throws Exception;
+    void addEvents(Collection<Event> events, boolean persist) throws Exception;
```
jshaughn: The events/data endpoint is a good addition to the API.   It may be nice thought if addEvents() remained unchanged and perhaps we add a new signature, sendEvents(), that is analogous to the already-existing sendData().

lucasponce: +1
I thought on it when I submitted the PR.
Thanks for pointing out.

lucasponce:
Feedback for this PR addressed and commit re-sync.

jshaughn:
Perhaps at some point in the future we will want to let Triggers generate these kind of "Transient" events, I believe today if the Trigger's EventType is Event it always stores the event.  Thanks for making the change, merging...

PR #209, opened 2016-09-21
Unify actions engine communication

lucasponce:
This PR unify actions plugins communication under the standalone implementation.
It will simplify future metrics+alerting scenarios in hawkular-services or standalone deployments.


hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -20,7 +20,7 @@
 <jboss-deployment-structure>
   <deployment>
     <dependencies>
-      <module name="org.hawkular.bus" />
+      <module name="deployment.hawkular-alerts-rest.war" />
```
jshaughn: Will there be any problem if alerting WAR is deployed inside of a metrics EAR, and not independently? Perhaps we need to to add  optional="true", and potentially add the other deployment scenario?

hawkular-alerts-actions-plugins/hawkular-alerts-actions-pagerduty/src/main/webapp/WEB-INF/jboss-all.xml
```diff
@@ -0,0 +1,24 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<jboss umlns="urn:jboss:1.0">
+  <jboss-deployment-dependencies xmlns="urn:jboss:deployment-dependencies:1.0">
+    <dependency name="hawkular-alerts-rest.war"/>
```
jshaughn: A similar concern, but with jboss-all I don't think there is an "optional".

hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
lucasponce: Yes. We need to touch a couple of things for .ear.
One are dependencies, and second is to extract/parametrize the JNDI location to include the .ear.
As we need to coordinate, I put that on my todo list to do it once the .ear is consolidate.
This PR is just to work with the present upstream status.

jshaughn: Maybe what we can do is merge this PR into master, rebase the m+a PR on master, and then update the dependencies to work with the m+a PR.  In that way, if we merge that PR it will work and not break with the m+a distribution.  But then it will not work for standalone alerting (which is the secondary use case) and we will have to figure that out.

lucasponce: Yes, I think we will have a couple of days of transition state until we can consolidate all changes.
+1 to merge and fix in a following one.

jshaughn: OK, will merge now, and will proceed with rebase of m+a PR, and will make that work again. We'll need to eventually solve the issue somehow such that Hm+a and standalone Ha both work.

PR #210, opened 2016-09-21
HWKALERTS-164 Unify resolved and ack fields with lifecycle

lucasponce:
This PR drops the old resolved* and ack* fields with new lifecycle list on Alert.
This change is propagated into actions architecture and plugins.

### DIFF
--- a/hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/EmailPlugin.java
+++ b/hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/EmailPlugin.java
@@ -302,9 +302,9 @@ public class EmailPlugin implements ActionPluginListener {
             if (alert.getStatus().equals(Status.OPEN)) {
                 email.setSentDate(new Date(alert.getCtime()));
             } else if (alert.getStatus().equals(Status.ACKNOWLEDGED)) {
-                email.setSentDate(new Date(alert.getCurrentLifecycle().getStime()));
+                email.setSentDate(new Date(alert.getLastAckTime()));
             } else {
-                email.setSentDate(new Date(alert.getCurrentLifecycle().getStime()));
+                email.setSentDate(new Date(alert.getLastResolvedTime()));
             }
         } else {
             email.setSentDate(new Date());
--- a/hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/EmailPlugin.java
+++ b/hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/EmailPlugin.java
@@ -302,9 +302,9 @@ public class EmailPlugin implements ActionPluginListener {
             if (alert.getStatus().equals(Status.OPEN)) {
                 email.setSentDate(new Date(alert.getCtime()));
             } else if (alert.getStatus().equals(Status.ACKNOWLEDGED)) {
-                email.setSentDate(new Date(alert.getAckTime()));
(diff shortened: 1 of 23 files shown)


lucasponce:
This is one of the commented on the discussion.
Please @jshaughn, can you review ?
Thanks.

hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/main/java/org/hawkular/alerts/actions/email/EmailPlugin.java
```diff
@@ -302,9 +302,9 @@ protected Message createMimeMessage(ActionMessage msg) throws Exception {
             if (alert.getStatus().equals(Status.OPEN)) {
                 email.setSentDate(new Date(alert.getCtime()));
             } else if (alert.getStatus().equals(Status.ACKNOWLEDGED)) {
-                email.setSentDate(new Date(alert.getAckTime()));
+                email.setSentDate(new Date(alert.getCurrentLifecycle().getStime()));
             } else {
-                email.setSentDate(new Date(alert.getResolvedTime()));
+                email.setSentDate(new Date(alert.getCurrentLifecycle().getStime()));
             }
```
jshaughn: I like the getting rid of the redundant fields but it may make sense to keep some convenience getters like:
```
  alert.getOpenTime
  alert.getAckTime
  alert.getResolvedTime
  alert.getStatusTime( status )
```
These convenience methods would reverse-traverse the lifecycle list and return the most recent time for the given status. What do you think?    We could potentially have similar methods that return the whole LifeCycle.  I think it's important to offer an easy way to get to the ackTime even if the current status is Resolved.

lucasponce: For REST users they will deal with the JSON representation then they will deal just with the lifecycle list.
But for Plugins developers who have access to Java Alerts API these methods can be useful so I will include them on the PR.

lucasponce: By the way,
-  alert.getAckTime
-  alert.getResolvedTime

can be ambiguous so I guess we can use 
-  alert.getLastAckTime
-  alert.getLastResolvedTime

as we introduced getCurrent* terminology getLast* makes sense to me.

jshaughn: For me the "Last" was implied (and jdoc) but I'm fine if you prefer to be explicit.

lucasponce:
Added getLast* methods.

hawkular-alerts-actions-plugins/hawkular-alerts-actions-email/src/test/resources/template.html.default_en_US.ftl
```diff
@@ -147,15 +147,15 @@
                                                 <tr>
                                                     <td align="center" style="color:#333333; font-family:Open sans,sans-serif; font-size:13px; line-height:21px;">
                                                         Acknowledge time:
-                                                        <span style="font-size:15px;">${alert.ackTime?number_to_datetime}</span>
+                                                        <span style="font-size:15px;">${alert.currentLifecycle.stime?number_to_datetime}</span>
```
jshaughn: Did you forget to use lastAckTim() here or is this the way you want it to be?  Fine either way, just checking.

lucasponce: No, it's on purpose. as we use the same alert currentLifecycle.user to get the user.
So I guess that maintaining same terminology in the template is better for clarity.

PR #213, opened 2016-09-26
ServiceName fix

jshaughn:
Because 'standalone' is what we explicitly set (in the web.xml), it can't also be the default.  Otherwise
'-metrics' deployments get the standalone jndi names.


jshaughn:
@lucasponce Please review.

hawkular-alerts-actions/hawkular-alerts-actions-impl/src/main/java/org/hawkular/alerts/actions/standalone/ServiceNames.java
```diff
@@ -33,20 +35,24 @@
...
     }
 
     private static Map<Service, String> services;
 
     static {
         services = new HashMap<>();
-        if (System.getProperty(HAWKULAR_ALERTS_ACTIONS_ENV, STANDALONE).equals(STANDALONE)) {
+        String env = null;
+        try {
+            env = (String) new InitialContext().lookup("java:comp/env/" + HAWKULAR_ALERTS_ACTIONS_ENV);
```
lucasponce: Why the change to move this from system property to env ?
Sorry I don't see it.
For consistency we should use system properties as it is the preferred way.
This has not affect about if it works or not, but I don't think is good to have these differences at this level.

jshaughn: The reason is simple, because it didn't work for me.  If you can show me that it works as it was I'm happy to change it back.  If we do this in some othr places we should ensure that is working as expected.  We can discuss off-pr if you like.

lucasponce: Oh, it seems that context-param is mapped into system properties in other containers.
I guess you are right.
The fact itests works on the standalone masked this one.

jshaughn: Right, because 'standalone' was being used as the default for the getProperty, it masked the fact that it was not actually resolving the property.  When I tried to run in Hservices it didn't work and I discovered the default-value issue.  When I fixed that I then found it didn't work in standalone...

lucasponce: Sorry :-).
Glad you catch this one.

PR #214, opened 2016-09-27
HWKALERTS-166 Refactor CassCluster component:

lucasponce:
- Move from JVM singleton to container singleton
- Session injected through components
- Fix Shutdown method invoked on end of lifecycle
- Check when a schema is fully created to avoid race conditions on clustering environments

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassActionsServiceImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassActionsServiceImpl.java
@@ -34,6 +34,7 @@ import javax.ejb.Stateless;
 import javax.ejb.TransactionAttribute;
 import javax.ejb.TransactionAttributeType;
 import javax.enterprise.concurrent.ManagedExecutorService;
+import javax.inject.Inject;
 
 import org.hawkular.alerts.api.json.JsonUtil;
 import org.hawkular.alerts.api.model.action.Action;
@@ -86,6 +87,10 @@ public class CassActionsServiceImpl implements ActionsService {
     @Resource
     private ManagedExecutorService executor;
 
+    @Inject
+    @CassClusterSession
+    Session session;
+
     public CassActionsServiceImpl() {
         log.debug("Creating instance.");
     }
@@ -100,6 +105,10 @@ public class CassActionsServiceImpl implements ActionsService {
 
     public void setExecutor(ManagedExecutorService executor) {
         this.executor = executor;
+    }
+
+    public void setSession(Session session) {
+        this.session = session;
     }
 
     @Override
@@ -171,7 +180,6 @@ public class CassActionsServiceImpl implements ActionsService {
             action.setResult(WAITING_RESULT);
         }
         try {
-            Session session = CassCluster.getSession();
             PreparedStatement insertActionHistory = CassStatement.get(session,
                     CassStatement.INSERT_ACTION_HISTORY);
             PreparedStatement insertActionHistoryAction = CassStatement.get(session,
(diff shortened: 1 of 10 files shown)


lucasponce:
There is a minor thing on cluster environment.
In the current tests, we set up a domain with 4 servers (alerts[1-3]+cassandra) when stop signal is thrown there is not a preserved stop order, so if cassandra node stops first, it shows some errors on the other nodes.
Also it happens with the ISPN nodes, there are some log errors at shutdown.
I will fix them in a following PR.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassActionsServiceImpl.java
```diff
@@ -102,6 +107,10 @@ public void setExecutor(ManagedExecutorService executor) {
         this.executor = executor;
     }
 
+    public void setSession(Session session) {
```
jshaughn: It seems a little odd to Inject session and also have a setter for session, is that intended?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
```diff
@@ -64,21 +76,111 @@
     private static final String ALERTS_CASSANDRA_READ_TIMEOUT = "hawkular-alerts.cassandra-read-timeout";
     private static final String ALERTS_CASSANDRA_READ_TIMEOUT_ENV = "CASSANDRA_READ_TIMEOUT";
 
+    /*
+        GLOBAL OVERWRITE
```
jshaughn: Can you add a little more doco as to what we're doing here and why?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassActionsServiceImpl.java
lucasponce: Yes, Cass* classes have some setters methods to run performance tests outside of the container. Just from a standalone class.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
lucasponce: ops, this comment was left here.
I forgot to remove it.
This is to replace the getSession(overwrite) method with the possibility to drop/recreate the existing scheme.
I will update the comment.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassActionsServiceImpl.java
jshaughn: OK. for testing reasons.

lucasponce:
> I'll be honest, I don't totally get what is going on here :) I think what we're now doing is, from every node, ensuring that every table is created before we proceed. is that correct? If so, somewhere I'd like to see some high level description of the startup/schema validation approach, just so when I come back months from now I'll remember what we are doing.

Previous logic checks that a keyspace is generated.
This normally works fine in one single node scenarios or multi nodes scenarios where schema is previosly generated (i.e. production environments).
A cornercase, but important, is when we have several nodes that start at same time, and some of them take the responsability to generate the schema.
This PR add an extra checker, in the form of validates that all tables are correctly generated. 
On this way, meanwhile one node is generating tables, others two have an active wait until the fully scheme is ready to go.

> Alternately, maybe in the future we should thing about adding a schema_status sort of table and use that possibly as a marker as to whether the schema is ready.

Well, it can be a possibility, but if some table is dropped via an external cqlsh operation that perhaps won't save too much.
I guess checker operation is save an it happens only at initialization.
I don't discard that when we introduce cassalog library we might revisit some configuration properties and move them into Cassandra.
Probably we might need a combination of both, but I guess that can go in a following task.

PR #215, opened 2016-09-27
Add C* dependencies on metrics deployment

lucasponce:
@jshaughn for your review

### DIFF
--- a/hawkular-alerts-rest/deployments/hawkular-alerts-rest-metrics/pom.xml
+++ b/hawkular-alerts-rest/deployments/hawkular-alerts-rest-metrics/pom.xml
@@ -40,6 +40,26 @@
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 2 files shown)


jshaughn:
@lucasponce You can close this PR.  The -standalone war dist is appropriate for use with metrics.  Thanks for noticing that, it saved us from a release!

lucasponce:
Glad it worked.
As a summary, today we have hawkular-alerts.war in three flavors:
- For Hawkular Services (metrics + alerting ear) + deployment on hw nest.
- For Hawkular Standalone (single alerting war) + deployment on standard wildfly.
- For Hawkular Metrics (metrics + alerting ear) + deployment on standard wildfly.

If standalone package can be used for metrics+alerting ear + wildfly, that's good, perhaps we don't need an specific package for Hawkular Metrics, but I think we will needed as soon as we introduce some differences on the authentication process (or other specific).

So, we can maintain the metrics deployment, today is practically similar to standalone one but it might diverge soon.

jshaughn:
>  but I think we will needed as soon as we introduce some differences on the authentication process (or other specific).

+1.  We should probably add the C* driver before we forget.  I'm still not sure about those other deps but if you think they are needed we can leave as is and I can merge the PR.

lucasponce:
I remember those dependencies are needed.
I am +1 to merge it, as this package will change shortly when we introduce differences.

PR #218, opened 2016-09-28
[WIP] HWKALERTS-169 Global front-line filtering

jshaughn:
- maintain the publish cache but makes alerting its client (as
  opposed to metrics, for example).
- simplify CacheManager and CacheKey to be based only on tenantId
  and dataId.
- perform global front-line filtering in AlertsService, as Data and
  Events come into the system.
- optimize the engine to use the node-specific AlertsEngineCache
  only in distributed deployment.  With global front-line filtering the
  node-specific filtering is redundant in the single node use case.
* sorry for some formatting changes, eclipse seems to be struggling with
  some of the long stream and lambda stuff.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheClient.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheClient.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.alerts.engine.cache;
+
+import java.util.Collection;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.annotation.Resource;
+import javax.enterprise.context.ApplicationScoped;
+
+import org.hawkular.alerts.api.model.data.CacheKey;
+import org.hawkular.alerts.api.model.data.Data;
+import org.hawkular.alerts.api.model.event.Event;
+import org.infinispan.Cache;
+
+/**
+ * Provide access to the cache of dataIds in use by the global trigger population (not node specific). It is
+ * used to perform front-line filtering of incoming data and events.  Data with dataIds not found in this cache
+ * can be immediately discarded as it is not needed for trigger evaluation (on this or or other alerting nodes).
+ *
+ * The cache is a shared ISPN cache.
+ *
+ * @author Lucas Ponce
+ * @author Jay Shaughnessy
+ */
+@ApplicationScoped
+public class CacheClient {
+
+    /** key=dataId, value="" */
+    @Resource(lookup = "java:jboss/infinispan/cache/hawkular-alerts/publish")
+    private Cache<CacheKey, String> cache;
+
+    public Set<CacheKey> keySet() {
+        return cache.keySet();
+    }
+
+    public boolean containsKey(CacheKey key) {
+        return cache.containsKey(key);
+    }
+
+    public String get(CacheKey key) {
+        return cache.get(key);
+    }
+
+    public Collection<Data> filterData(Collection<Data> data) {
+        final CacheKey tester = new CacheKey("", "");
+        return data.stream()
+                .filter(d -> cache.containsKey(fillKey(tester, d)))
+                .collect(Collectors.toList());
+    }
+
+    public Collection<Event> filterEvents(Collection<Event> events) {
+        final CacheKey tester = new CacheKey("", "");
+        return events.stream()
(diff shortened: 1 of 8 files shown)


jshaughn:
@lucasponce  Take a look when you get the chance and let me know what you think of this approach.  Because this also affects metrics I have a branch at jshaughn/hawkular-metrics that goes with this, but I wanted to get your feedback before doing anything else.

lucasponce:
Before to enter in details I've fetched the PR and I can confirm that travis errors are there.
A "mvn clean install -Pitest" fails, so that can give you hints that something is causing a regression into the code.
Also it could be nice if you can test the clustering env "mvn clean install -Denv=cluster" to confirm that there are also not regressions in this environment.
Travis cannot tested automatically as it is heavier for him.

I will comment more at code detail.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheClient.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private CacheKey fillKey(CacheKey key, Data data) {
+        key.setTenantId(data.getTenantId());
+        key.setDataId(data.getId());
+        return key;
+    }
+
+    private CacheKey fillKey(CacheKey key, Event event) {
+        key.setTenantId(event.getTenantId());
+        key.setDataId(event.getId());
```
lucasponce: key.setDataId(event.getId()); -> key.setDataId(event.getDataId());

lucasponce:
Clustering environment fails with these proposed changes.
I am investigating it.

jshaughn:
> A "mvn clean install -Pitest" fails

Yes, I'm aware, that is why I set it DoNotMerge and mentioned that I was looking for initial feedback...  I'll be looking at the failures this morning...

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -1143,12 +1149,12 @@ public Event getEvent(String tenantId, String eventId, boolean thin) throws Exce
             List<Event> ordered = events;
             if (pager.getOrder() != null) {
                 pager.getOrder()
-                        .stream()
-                        .filter(o -> o.getField() != null && o.getDirection() != null)
-                        .forEach(o -> {
-                            EventComparator comparator = new EventComparator(o.getField(), o.getDirection());
-                            Collections.sort(ordered, comparator);
-                        });
+                .stream()
```
jshaughn: Yes, it's annoying to get the settings just right in Eclipse.  I'll recheck my settings, I don't want it to reformat the entire file.  Sometimes I think I accidentally format the file myself and don't realize until it's time to commit.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheClient.java
jshaughn: good catch!

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -470,32 +481,34 @@ public void sendEvents(Collection<Event> events) {
...
+        Collection<Data> filteredData = data;
+        if (distributed) {
+            filteredData = new ArrayList<>(data);
+            for (Iterator<Data> i = filteredData.iterator(); i.hasNext(); ) {
+                Data d = i.next();
+                if (!alertsEngineCache.isDataIdActive(d.getId())) {
+                    i.remove();
+                }
+            }
+        } else
```
lucasponce: This "else" has created on me a nice debugging session :-)

jshaughn: You can fix my code anytime :)  Thanks!

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -56,43 +56,23 @@
...
     private static final String RESET_PUBLISH_CACHE_PROP = "hawkular-alerts.reset-publish-cache";
     private static final String RESET_PUBLISH_CACHE_ENV = "RESET_PUBLISH_CACHE";
 
-    private static final Set<String> DATA_ID_PREFIXES = new HashSet<>();
-
-    static {
-        // TODO: For now just use the built-in H Metrics prefixes as a OOB convenience, but in the
-        //       future this should maybe be configurable where a list of prefixes could be supplied
-        //       via system prop.
-        DATA_ID_PREFIXES.add("hm_a_");  // Metrics AVAILABILITY
```
lucasponce: Do you remember if these prexises are documented in the metrics.ear code ?
Just wondering for documentation purposes.
It could be nice to have an interface with swagger to maintain them in some way.
I remember I introduced one in the preliminar code but with the merging of m+a not sure if this is missing somewhere.
Usecase: to reference an external user the correct syntax on triggers creation.

jshaughn: The prefixes are now known only on the metrics side of the code, because it needs to generate the proper prefixes for the dataIds.  This makes sense to me because it really is the metrics client's responsibility.  The prefixes are a convention established due to the way Hm generates its primary keys.  It would be useful documentation for Hm where we discuss how to integrate with Ha.

lucasponce:
I have reviewed and it is good to merge.
I can do that once you remove the do-not-merge label and the metrics counterpart is ready.

jshaughn:
OK, great.  Let me get the metrics side in shape and open a PR for that as well, so we can review it formally.

PR #220, opened 2016-09-30
external alerter

jshaughn:
- change some deps to provided.
- remove unnecessary class


jshaughn:
@lucasponce please review. I am going to merge this now but if there is any issue we can address it in a follow-on PR.

PR #221, opened 2016-10-03
HWKALERTS-171 Support SSL on Cassandra connections

lucasponce:
Please @jshaughn take a look on this.


lucasponce:
/cc @jsanda @mwringe
I guess that at app level this is the flag needed.
If there is any pending task, just let me know.
Thanks

mwringe:
from a quick glance, I think that looks good

PR #225, opened 2016-10-05
Reduce the set of modules to build for container deliverable.

stefannegrea:


jshaughn:
Thanks @stefannegrea for this PR to add a slim build of just the necessary components for Openshift.  I have made a couple of small changes.  @lucasponce please review the support for an openshift profile.  Moving forward we may tweak the plugin line-up, this was a first guess.

lucasponce:
A preliminar comment.
In alerting we normally fork and we create PR from your own repo to upstream.
Let's not create branches on hawkular-alerts repo for substitute a fork/clone step.
That creates risk because you commit on upstream for a working branch instead of origin (your clone).

In gitthub things can be done in thousand of variations, and everything may be correct but let's follow this on alerting.

Thanks.

pom.xml
```diff
@@ -135,6 +141,19 @@
         <module>hawkular-alerts-actions</module>
       </modules>
     </profile>
+
+    <profile>
+      <id>openshift</id>
+      <properties>
+        <skipTests>true</skipTests>
```
lucasponce: In general, I am against adding a skipTests inside a profile.
If this profile will be used to package things specific for openshift, testing is necessary.
Skipping tests can be added at command line.

What is the context of this profile ? It is some developer helper or it is part of other automated process ?

lucasponce:
I'm not going to block this PR.
I'm going to merge but a couple of things:
- Please, let's delete the working branch and next time use a fork/clone steps.
- skipTests should not go on profile, if this profile is needed in travis or similar simply adding at command line has the same effect, skipping tests on profiles increase the risk of lack testing coverage.
Thanks @stefannegrea.

jshaughn:
> In alerting we normally fork and we create PR from your own repo to upstream.

+1, I meant to mention that to @stefannegrea, please just use your own forked repo for PR branches in the future.

> In general, I am against adding a skipTests inside a profile.

Certainly a good thing to not have that in general.  I'm guessing in this packaging profile that the reason is to avoid pulling in and therefore productizing test deps.   I'm thinking this may be standard practice for this sort of thing, but @stefannegrea should verify.

stefannegrea:
@lucasponce, I will follow your fork convention for future PRs.

Regarding `skipTests`, skipping tests is a requirement because the build is done in a bare build environment with no network access. Plus, we want to avoid anything test related (compiling tests, pulling in test dependencies, or attempting to run tests). I would not have done this for any other profiles, but this is the default mode for productization.

PR #229, opened 2016-10-13
hwkmetrics-507

jshaughn:
- make injecting CacheClient easier by moving it and CacheKey to its own jar
- add ignoreFiltering option to sendData/Events to allow clients to
  perform their own filtering and (java clients only for now)

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
@@ -218,6 +218,15 @@ public interface AlertsService {
     void sendData(Collection<Data> data) throws Exception;
 
     /**
+     * Send data into the alerting system for evaluation.
+     *
+     * @param data Not Null.  The data to be evaluated by the alerting engine.
+     * @param ignoreFiltering  An optimization. Set true *only* if you are sure the data is useful for evaluation.
+     * @throws Exception any problem.
+     */
+    void sendData(Collection<Data> data, boolean ignoreFiltering) throws Exception;
+
+    /**
      * Send event to the engine for alerts evaluation.
      * The event sent is not persisted into the alerts engine.
      *
@@ -234,4 +243,14 @@ public interface AlertsService {
      * @throws Exception
      */
     void sendEvents(Collection<Event> events) throws Exception;
+
+    /**
+     * Send events to the engine for alerts evaluation.
+     * The event sent are not persisted into the alerts engine.
+     *
+     * @param events Not null. The events to be evaluated by the alerting engine.
+     * @param ignoreFiltering  An optimization. Set true *only* if you are sure the data is useful for evaluation.
+     * @throws Exception
+     */
+    void sendEvents(Collection<Event> events, boolean ignoreFiltering) throws Exception;
 }
--- a/hawkular-alerts-engine/pom.xml
+++ b/hawkular-alerts-engine/pom.xml
@@ -39,6 +39,11 @@
       <artifactId>hawkular-alerts-api</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.hawkular.alerts</groupId>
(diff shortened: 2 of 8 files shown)


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
```diff
@@ -234,4 +243,14 @@ void resolveAlertsForTrigger(String tenantId, String triggerId, String resolvedB
...
+
+    /**
+     * Send events to the engine for alerts evaluation.
+     * The event sent are not persisted into the alerts engine.
+     *
+     * @param events Not null. The events to be evaluated by the alerting engine.
+     * @param ignoreFiltering  An optimization. Set true *only* if you are sure the data is useful for evaluation.
+     * @throws Exception
+     */
+    void sendEvents(Collection<Event> events, boolean ignoreFiltering) throws Exception;
```
lucasponce: Same here.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
```diff
@@ -218,6 +218,15 @@ void resolveAlertsForTrigger(String tenantId, String triggerId, String resolvedB
     void sendData(Collection<Data> data) throws Exception;
 
     /**
+     * Send data into the alerting system for evaluation.
+     *
+     * @param data Not Null.  The data to be evaluated by the alerting engine.
+     * @param ignoreFiltering  An optimization. Set true *only* if you are sure the data is useful for evaluation.
+     * @throws Exception any problem.
+     */
+    void sendData(Collection<Data> data, boolean ignoreFiltering) throws Exception;
```
jshaughn: In general this flag would not be used.  It is only when a client is sure they don't need the incoming data to be filter-checked.  In this case metrics would be the only java client in the foreseeable future. A config prop is probably not appropriate because it is client-specific, perhaps even call-specific for a particular client.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
jshaughn: same response :)

lucasponce:
Is there a counterpart PR for metrics that I can review too ?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
lucasponce: A client shouldn't have visibility of this filtering process IMO.
Another thing is that for packages where we know that metrics or apm are in front and performing some kind of filtering process, then a flag at config should activate this.
But for API purposes I don't think this is appropiate.

jshaughn:
> Is there a counterpart PR for metrics that I can review too ?

Sure, let me update that side as well, and you can take a look, thanks...

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
jshaughn: A configuration option would mean that every call from that client would be subjected to the same setting.  If you are blocking the PR on this I'll remove the option.  Although if the client can guarantee valid data I don't really see an issue. Metrics obviously is aware of filtering.  Let me know if this a blocking issue.

lucasponce:
Metrics client PR is this one (for future references):
https://github.com/hawkular/hawkular-metrics/pull/638

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/AlertsService.java
jshaughn: OK, I think it's OK for now. I mean, it's the java API and we only have one external client, and that is our tight integration with metrics.  Going forward we can certainly iterate.  I didn't include any option via the rest-api.

PR #230, opened 2016-10-14
HWKALERTS-176 Add support for MissingConditions

lucasponce:
As described in HWKALERTS-176 this PR implements a new type of condition: MissingCondition.
The purpose of this type is the ability to alert when a data/event has not been received into the system.
Some use cases (when metrics are involved) can be handled via ExternalCondition and Alerter add-ons, others simply not.
This enhancement fills that gap.
Main use case is the ability to respond to "Do I have received an event for this ? If not, alert".
The implementation details basically maintains a timer state when this condition is active resetting when new Events or Data arrives.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
@@ -43,6 +43,7 @@ import javax.enterprise.concurrent.ManagedExecutorService;
 import org.hawkular.alerts.api.model.condition.CompareCondition;
 import org.hawkular.alerts.api.model.condition.Condition;
 import org.hawkular.alerts.api.model.condition.ConditionEval;
+import org.hawkular.alerts.api.model.condition.MissingCondition;
 import org.hawkular.alerts.api.model.dampening.Dampening;
 import org.hawkular.alerts.api.model.data.Data;
 import org.hawkular.alerts.api.model.event.Alert;
@@ -59,6 +60,7 @@ import org.hawkular.alerts.engine.service.PartitionManager;
 import org.hawkular.alerts.engine.service.PartitionManager.Operation;
 import org.hawkular.alerts.engine.service.PartitionTriggerListener;
 import org.hawkular.alerts.engine.service.RulesEngine;
+import org.hawkular.alerts.engine.util.MissingState;
 import org.jboss.logging.Logger;
 
 /**
@@ -96,6 +98,7 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
     private final Set<Dampening> pendingTimeouts;
     private final Map<Trigger, List<Set<ConditionEval>>> autoResolvedTriggers;
     private final Set<Trigger> disabledTriggers;
+    private final Set<MissingState> missingStates;
 
     private final Timer wakeUpTimer;
     private TimerTask rulesTask;
@@ -133,6 +136,7 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
         pendingTimeouts = new HashSet<>();
         autoResolvedTriggers = new HashMap<>();
         disabledTriggers = new HashSet<>();
+        missingStates = new HashSet<>();
 
         wakeUpTimer = new Timer("CassAlertsServiceImpl-Timer");
 
@@ -217,6 +221,7 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
         pendingTimeouts.clear();
         autoResolvedTriggers.clear();
         disabledTriggers.clear();
+        missingStates.clear();
 
         rulesTask = new RulesInvoker();
         wakeUpTimer.schedule(rulesTask, delay, period);
@@ -263,6 +268,7 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
         rules.addGlobal("pendingTimeouts", pendingTimeouts);
         rules.addGlobal("autoResolvedTriggers", autoResolvedTriggers);
         rules.addGlobal("disabledTriggers", disabledTriggers);
+        rules.addGlobal("missingStates", missingStates);
 
         rulesTask = new RulesInvoker();
         wakeUpTimer.schedule(rulesTask, delay, period);
@@ -371,6 +377,9 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
                             alertsEngineCache.add(entry2);
                         }
                     }
+                    if (c instanceof MissingCondition) {
+                        missingStates.add(new MissingState(trigger, (MissingCondition) c));
+                    }
                 }
 
                 rules.addFact(trigger);
@@ -441,6 +450,13 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
                 }
                 return false;
             });
+            Iterator<MissingState> it = missingStates.iterator();
+            while (it.hasNext()) {
+                MissingState missingState = it.next();
+                if (missingState.getTriggerId().equals(triggerId)) {
+                    it.remove();
+                }
+            }
         } else {
             if (log.isDebugEnabled()) {
                 log.debug("Trigger not found. Not removed from rulebase " + trigger.toString());
@@ -537,7 +553,9 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
         public void run() {
             int numTimeouts = checkPendingTimeouts();
 
-            if (!pendingData.isEmpty() || !pendingEvents.isEmpty() || numTimeouts > 0) {
+            handleMissingStates();
+
+            if (!pendingData.isEmpty() || !pendingEvents.isEmpty() || numTimeouts > 0 || !missingStates.isEmpty()) {
                 Collection<Data> newData = getAndClearPendingData();
                 Collection<Event> newEvents = getAndClearPendingEvents();
 
@@ -669,6 +687,14 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
         } finally {
             autoResolvedTriggers.clear();
         }
+    }
+
+    private void handleMissingStates() {
+        missingStates.stream().forEach(missingState -> {
+            rules.removeFact(missingState);
+            missingState.setTime(System.currentTimeMillis());
+            rules.addFact(missingState);
+        });
     }
 
     /*
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
@@ -704,8 +704,8 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
                 missingState.setPreviousTime(System.currentTimeMillis());
                 missingState.setTime(System.currentTimeMillis());
                 rules.addFact(missingState);
-                rules.addFact(eval);
             }
+            rules.addFact(eval);
         });
     }
 
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
@@ -269,7 +269,6 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
         rules.addGlobal("pendingTimeouts", pendingTimeouts);
         rules.addGlobal("autoResolvedTriggers", autoResolvedTriggers);
         rules.addGlobal("disabledTriggers", disabledTriggers);
-        rules.addGlobal("missingStates", missingStates);
 
         rulesTask = new RulesInvoker();
         wakeUpTimer.schedule(rulesTask, delay, period);
@@ -380,6 +379,7 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
                     }
                     if (c instanceof MissingCondition) {
                         MissingState missingState = new MissingState(trigger, (MissingCondition) c);
+                        // MissingStates are modified inside the rules engine
                         missingStates.add(missingState);
                         rules.addFact(missingState);
                     }
@@ -697,10 +697,11 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
     private int checkMissingStates() {
         int numMatchingEvals = 0;
         for (MissingState missingState : missingStates) {
-            MissingConditionEval eval = new MissingConditionEval(missingState.getCondition(),
-                    missingState.getPreviousTime(),
-                    System.currentTimeMillis());
-            if (eval.isMatch()) {
+            long now = System.currentTimeMillis();
+            if (missingState.getCondition().match(missingState.getPreviousTime(), now)) {
+                MissingConditionEval eval = new MissingConditionEval(missingState.getCondition(),
+                        missingState.getPreviousTime(),
+                        now);
                 rules.removeFact(missingState);
                 missingState.setPreviousTime(System.currentTimeMillis());
                 missingState.setTime(System.currentTimeMillis());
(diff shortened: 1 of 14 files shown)


lucasponce:
I added the do-not-merge label to give time to review this.
@jshaughn feel free to give your comments.

lucasponce:
/cc @objectiser
This enhancement could be useful for some APM scenarios as well.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -441,6 +450,13 @@ private void removeTrigger(Trigger trigger) {
                 }
                 return false;
             });
+            Iterator<MissingState> it = missingStates.iterator();
+            while (it.hasNext()) {
+                MissingState missingState = it.next();
+                if (missingState.getTriggerId().equals(triggerId)) {
```
lucasponce: Missing tenantId validation here.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/RulesEngineTest.java
```diff
@@ -1888,4 +1893,84 @@ public void checkEqualityInRulesEngine() throws Exception {
...
+        rulesEngine.addFact(t1);
+        rulesEngine.addFact(fmt1);
+
+        // Rules Invoker thread updates missing states periodically
+        MissingState missingState = missingStates.iterator().next();
+
+        rulesEngine.removeFact(missingState);
+
+        // It simulates that we have not data in 4 seconds > of 3 seconds interval defined
+        missingState.setPreviousTime(missingState.getTime());
```
lucasponce: missingState.previousTime should not be updated here.
This comes from a preliminar version.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/RulesEngineTest.java
```diff
@@ -1888,4 +1893,84 @@ public void checkEqualityInRulesEngine() throws Exception {
...
+        assertEquals(1, alerts.size());
+
+
+        // Rules Invoker thread updates missing states periodically
+        missingState = missingStates.iterator().next();
+
+        rulesEngine.removeFact(missingState);
+
+        // It simulates that we have not data in 4 seconds > of 3 seconds interval defined
+        missingState.setPreviousTime(missingState.getTime());
```
lucasponce: Same here.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/RulesEngineTest.java
```diff
@@ -1888,4 +1893,84 @@ public void checkEqualityInRulesEngine() throws Exception {
...
+
+        rulesEngine.addFact(t1);
+        rulesEngine.addFact(fmt1);
+
+        MissingState missingState = missingStates.iterator().next();
+
+        rulesEngine.removeFact(missingState);
+
+        // It simulates that we have not data in 4 seconds > of 3 seconds interval defined
+        missingState.setPreviousTime(missingState.getTime());
```
lucasponce: Same here.

jshaughn:
Hola @lucasponce , I have started reviewing the PR but will need a bit more time.  I ran out of time on Friday and want to make sure I give this enough time.  My initial impression is that I think this works and the PR seems thorough (although I didn't see 'interval' added in hawular-alerts-schema.cql).  I am thinking more on whether we can make it a bit more efficient.  I may have one potential modification for you to consider, I'll think more on it and if it still makes sense I'll pass it by you on Monday.

lucasponce:
@jshaughn any update on the review of this one ?
It is not urgent but it would be good if we can get in before the feature freeze.
Thanks.

jshaughn:
@lucasponce I have a detailed comment coming soon.  Did you see this in my initial comment:

> ... (although I didn't see 'interval' added in hawular-alerts-schema.cql).

lucasponce:
> ... (although I didn't see 'interval' added in hawular-alerts-schema.cql).

There is no more hawkular-alerts-schema.cql file :-).
With Cassalog we have an initial scheme created from bootstrap.groovy and iterative ones from updates/*.groovy.
On this way existing schemas can be manipulated in an iterative way.

For 'interval' column it is added in a schema modification.

jshaughn:
I like the fundamental approach you have here.  MissingCondition, MissingConditionEval and MissingState are all good additions.  I think it would work as is, I am now only wondering whether the approach could be streamlined.

Using the same elements you have introduced, I'd like to propose this modification to the algorithm:

1. Inject 1 MissingState fact into the engine for each MissingCondition
    - this only needs to be done once, the MissingState just stays in working memory.
      - add/remove the MissingState facts only in response to added/removed MissingConditions.
      - no need for 'HandleMissingStates' fact updates on each run

2. In the rulebase, update the MissingState time when a Data/Event in WM matches
    - So, basically the same two rules you have but with the following differences:
       - I don't think you need the salience
       - No retraction
       - Add the updated MissingState to a global set updatedMissingStates (I think this would replace the current missingStates global)

3. Like what we do with checkPendingTimeouts, perform checkMissingState on each engine run.
    - If the interval has expired we directly add a satisfied MissingConditionEval into WM
    - No need for 'Missing' rule
    - No 'no match' MissingConditionEvals
     
Let me know what you think, either here or we can talk about it off-PR.  It's not far off from what you have already, it uses all of the same elements and the same fundamental approach.  Assuming the modifications work (I haven't implemented it) I think the benefits are: a reduction in MissingState/WorkingMemory churn, removal of negatively evaluated MissingConditionEvals on each run, and that it is fairly close to the existing logic for pendingTimeouts.

jshaughn:
> There is no more hawkular-alerts-schema.cql file :-).

Good to know :)

lucasponce:
> this only needs to be done once, the MissingState just stays in working memory.

I guess I understand your proposal. But there is

lucasponce:
Sorry, closed by mistake, I was updating some comment.
Reopened.

lucasponce:
I have applied most of the suggestions proposed.
I would prefer to have all Evals generated from single place but saving rules is good, so it is a valid trade-off.

jshaughn:
> would prefer to have all Evals generated from single place

I understand, that is a departure from normal.  One thing I am wondering now is whether never generating negative MissingConditionEvals is an issue with dampening.  Perhaps we do need generate a false MissingConditionEval whenever we update MissingState due to a Data/Event.   I will review your changes now...

lucasponce:
> One thing I am wondering now is whether never generating negative MissingConditionEvals is an issue with dampening. 

That's true, without negative evals dampening is dead.
Perhaps we should maintain the ConditionEvals in the rules engine.
I try to test your suggestions and I forgot the Dampening variable on that.
Well, we have two options, to maintain the original design or to send a negative eval from outside.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -671,6 +694,21 @@ private void handleAutoResolvedTriggers() {
...
+        missingStates.stream().forEach(missingState -> {
+            MissingConditionEval eval = new MissingConditionEval(missingState.getCondition(),
+                    missingState.getPreviousTime(),
+                    System.currentTimeMillis());
+            if (eval.isMatch()) {
+                rules.removeFact(missingState);
+                missingState.setPreviousTime(System.currentTimeMillis());
+                missingState.setTime(System.currentTimeMillis());
+                rules.addFact(missingState);
+                rules.addFact(eval);
```
lucasponce: I'm testing locally to add the eval in any case.
So, MissingConditionEval negatives work in a regular way with Dampening model.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -537,7 +558,9 @@ private void addPendingEvent(Event event) {
         public void run() {
             int numTimeouts = checkPendingTimeouts();
 
-            if (!pendingData.isEmpty() || !pendingEvents.isEmpty() || numTimeouts > 0) {
+            checkMissingStates();
```
jshaughn: This isn't quite what I was suggesting.  Let me try to clarify, the idea is closer to what we do with `checkPendingTimeouts()`.   `MissingStates` is a `Set` of `MissingState` with one entry per `MissingCondition` and an associated Fact in working memory.   This should not be a global shared with the rules engine.  Instead, we'd want a new set, like `Set<MissingState> updatedMissingState<>()`, that would be the shared global.

So, here, instead of:
`checkMissingStates();`
We'd have something like:
`int numMissingTooLong = checkMissingStates();`

The `checkMissingStates()` method would do a couple of things:

1. It would run through (the new global) `updatedMissingStates` and reflect those update times in the relevant `MissingStates` entry.
2. It would clear `updatedMissingStates` in preparation of the next engine run.
3. It would run through `MissingStates` and for any that had surpassed the defined interval of inactivity it would generate the positive `MissingConditionEval` fact and inject it into working memory.  And, in this case probably also reset the interval timer.

So, the check:
`if (!pendingData.isEmpty() || !pendingEvents.isEmpty() || numTimeouts > 0 || !missingStates.isEmpty())`
would change to:
`if (!pendingData.isEmpty() || !pendingEvents.isEmpty() || numTimeouts > 0 || numMissingTooLong > 0`

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -671,6 +694,21 @@ private void handleAutoResolvedTriggers() {
         }
     }
 
+    private void checkMissingStates() {
```
jshaughn: See the comments above regarding this method.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -263,6 +269,7 @@ public void reload() {
         rules.addGlobal("pendingTimeouts", pendingTimeouts);
         rules.addGlobal("autoResolvedTriggers", autoResolvedTriggers);
         rules.addGlobal("disabledTriggers", disabledTriggers);
+        rules.addGlobal("missingStates", missingStates);
```
jshaughn: See the comments below regarding this global.

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -260,6 +264,33 @@ rule Rate
...
+        $md : MissingState( $tenantId : tenantId, $tsource : source, $did : dataId, previousTime < time )
+        $d  : Data( tenantId == $tenantId, source == $tsource, id == $did)
+   then
+        if (log != null && log.isDebugEnabled()) {
+            log.debugf("UpdateMissingStateFromData: %s ", $md);
+        }
+        modify( $md ) {
+            setPreviousTime($d.getTimestamp()),
+            setTime($d.getTimestamp())
+        }
```
jshaughn: I think this would do a couple more things:

1. `updatedMissingStates.add( $md )`
2. create a new, negative,` MissingConditionEval` and add it to WM

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -260,6 +264,33 @@ rule Rate
...
+        $e  : Event( tenantId == $tenantId, dataSource == $tsource, dataId == $did)
+    then
+        if (log != null && log.isDebugEnabled()) {
+            log.debugf("UpdateMissingStateFromEvent: %s ", $md);
+        }
+        modify( $md ) {
+            setPreviousTime($e.getCtime()),
+            setTime($e.getCtime())
+        };
+end
```
jshaughn: Analogous comment to the above rule.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
lucasponce: I think you are changing the design here.
I don't think 
int numMissingTooLong = checkMissingStates();

make sense, as missingState is a timer, so it's need updated and evaluated. a !missingStates.isEmpty() sounds enough to me to run the rules engine with no data, missing states needs to be in sync with MissingConditions here.

I don't see what use case are we missing in the present design (if any).

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
lucasponce: Responded.

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
lucasponce: Responded.

lucasponce:
@jshaughn finally added a similar
> int numMissingTooLong = checkMissingStates();
and moved the non matching evaluations to be fired from inside the rules.

I see the optimization, thanks for the discussion.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -671,6 +694,24 @@ private void handleAutoResolvedTriggers() {
         }
     }
 
+    private int checkMissingStates() {
+        int numMatchingEvals = 0;
+        for (MissingState missingState : missingStates) {
+            MissingConditionEval eval = new MissingConditionEval(missingState.getCondition(),
+                    missingState.getPreviousTime(),
+                    System.currentTimeMillis());
+            if (eval.isMatch()) {
```
jshaughn: Not a blocker but you could save some Object creation if you do:
```
long now = System.currentTimeMillis();
...
if ( missingState.condition.match(previousTime, now)) {
  MissingConditionEval eval = new...
```

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -63,7 +67,7 @@ global List events;
 global Set pendingTimeouts;
 global Map autoResolvedTriggers;
 global Set disabledTriggers;
-
+global Set missingStates
```
jshaughn: This can be removed, I think.

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -260,6 +264,33 @@ rule Rate
...
 
+rule UpdateMissingStateFromData
+    when
+        $md : MissingState( $tenantId : tenantId, $tsource : source, $did : dataId, previousTime < time )
+        $d  : Data( tenantId == $tenantId, source == $tsource, id == $did)
+   then
+        if (log != null && log.isDebugEnabled()) {
+            log.debugf("UpdateMissingStateFromData: %s ", $md);
+        }
+        modify( $md ) {
```
jshaughn: Please comment, either here or somewhere in AlertsEngineImpl that this change is reflected in AlertsEngineImpl, it's important to realize this toi understand why checkMissingState works.

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -260,6 +264,33 @@ rule Rate
         insert( ce );
 end
 
+rule UpdateMissingStateFromData
+    when
+        $md : MissingState( $tenantId : tenantId, $tsource : source, $did : dataId, previousTime < time )
```
jshaughn: perhaps rename this variable from $md to $ms, that is sort of the convention we have.

PR #231, opened 2016-10-14
Add additional performance tests

lucasponce:
This improvement will help to not block Java clients as metrics that send directly data into the engine.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -50,6 +50,8 @@ public class DroolsRulesEngineImpl implements RulesEngine {
     // private final MsgLogger msgLog = MsgLogger.LOGGER;
     private final Logger log = Logger.getLogger(DroolsRulesEngineImpl.class);
     private static final String SESSION_NAME = "hawkular-alerts-engine-session";
+    private static final long PERF_FILTERING_THRESHOLD= 3000L; // 3 seconds
+    private static final long PERF_FIRING_THRESHOLD = 5000L; // 5 seconds
 
     private KieServices ks;
     private KieContainer kc;
@@ -155,20 +157,18 @@ public class DroolsRulesEngineImpl implements RulesEngine {
         // execution of the rules.  So, if we find multiple Data instances for the same Id, defer all but
         // the oldest to a subsequent run. Note that pendingData is already sorted by (id ASC, timestamp ASC) so
         // the iterator will present Data with the same id together, and time-ordered.
-        log.debug("firing rules...");
+        int initialPendingData = pendingData.size();
+        int initialPendingEvents = pendingEvents.size();
+        log.debugf("Firing rules... PendingData [%s] PendingEvents [%s]", initialPendingData, initialPendingEvents);
         int fireCycle = 0;
+        long startFiring = System.currentTimeMillis();
         while (!pendingData.isEmpty() || !pendingEvents.isEmpty()) {
-
-            if (log.isDebugEnabled()) {
-                log.debug("Data found. Firing rules on [" + pendingData.size() + "] datums and " +
-                        "[" + pendingEvents.size() + "] events.");
-            }
-
-            TreeSet<Data> batchData = new TreeSet<Data>(pendingData);
+            TreeSet<Data> batchData = new TreeSet<>(pendingData);
             Data previousData = null;
 
             pendingData.clear();
 
+            long startFiltering = System.currentTimeMillis();
             for (Data data : batchData) {
                 if (null == previousData || !data.getId().equals(previousData.getId())) {
                     kSession.insert(data);
@@ -181,6 +181,12 @@ public class DroolsRulesEngineImpl implements RulesEngine {
                                 "processed");
                     }
                 }
+            }
+            long filteringTime = System.currentTimeMillis() - startFiltering;
+            log.debugf("Filtering BatchData [%s] took [%s]", batchData.size(), filteringTime);
+            if (filteringTime > PERF_FILTERING_THRESHOLD) {
+                log.warnf("Filtering Filtering BatchData [%s] took [%s] ms exceeding [%s] ms",
+                        batchData.size(), filteringTime, PERF_FILTERING_THRESHOLD);
             }
 
             if (!pendingData.isEmpty() && log.isDebugEnabled()) {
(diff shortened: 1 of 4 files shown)


jshaughn:
I agree that sendData should be more of a simple delivery of data with a fast return.  But lets hold on this PR until we can evaluate the incoming data flow a bit more, and see the effects of the change.  I don't think we have an issue here, so we can address this after a bit more study.

lucasponce:
I have reverted the asynchronous suggestion.
I also used this PR to add an additional performance test, it also complements other specific tests used from the engine and CRUD scenarios.
Added also some useful warning flags that will help to early determine if filtering/firing processing is under pressure.
I have been studying several scenarios to create some realistic perf tests that can help into perf investigations, I guess these ones are a step forward in that sense.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -155,20 +157,18 @@ public void fire() {
         // execution of the rules.  So, if we find multiple Data instances for the same Id, defer all but
         // the oldest to a subsequent run. Note that pendingData is already sorted by (id ASC, timestamp ASC) so
         // the iterator will present Data with the same id together, and time-ordered.
-        log.debug("firing rules...");
+        int initialPendingData = pendingData.size();
+        int initialPendingEvents = pendingEvents.size();
+        log.debugf("Firing rules... PendingData [%s] PendingEvents [%s]", initialPendingData, initialPendingEvents);
```
jshaughn: Shouldn't these 3 lines be under the 'while' below?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -50,6 +50,8 @@
     // private final MsgLogger msgLog = MsgLogger.LOGGER;
     private final Logger log = Logger.getLogger(DroolsRulesEngineImpl.class);
     private static final String SESSION_NAME = "hawkular-alerts-engine-session";
+    private static final long PERF_FILTERING_THRESHOLD= 3000L; // 3 seconds
```
jshaughn: We are using the term 'Filtering' to mean a few different things.  I think we should leave 'filtering' to mean rejecting data that is not used in conditions.  The pendingData is already 'filtered' and now we are ordering/deferring/batching it for submission to the engine.  Perhaps we can change this to PERF__BATCHING_THRESHOLD or something like that, and make analogous naming changes below?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -182,6 +182,12 @@ public void fire() {
                     }
                 }
             }
+            long filteringTime = System.currentTimeMillis() - startFiltering;
+            log.debugf("Filtering BatchData [%s] took [%s]", batchData.size(), filteringTime);
+            if (filteringTime > PERF_FILTERING_THRESHOLD) {
+                log.warnf("Filtering Filtering BatchData [%s] took [%s] ms exceeding [%s] ms",
```
jshaughn: Duplicate 'Filtering' in message (also see comment about changing 'Filtering' terminology).

Also, not show here in the diff, but above in the code we have this log message:
```
if (log.isDebugEnabled()) {
    log.debug("Deferring more recent " + data + " until older " + previousData + " is " + "processed");
}
```

Perhaps this should be removed completely.  For large amounts of data this may generate too much logging.  If we keep it, please convert to debugf.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -220,6 +226,14 @@ public void fire() {
             kSession.fireAllRules();
             fireCycle++;
         }
+        long firingTime = System.currentTimeMillis() - startFiring;
+        if (log.isDebugEnabled()) {
+            log.debugf("Firing took [%s] ms", firingTime);
+        }
+        if (firingTime > PERF_FIRING_THRESHOLD) {
+            log.warnf("Firing rules... PendingData [%s] PendingEvents [%s] took [%s] msexceeding [%s] ms",
```
jshaughn: 'msexceeding' -> 'ms exceeding'

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
lucasponce: +1 to PERF_BATCHING_THRESHOLD

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
lucasponce: initialPending* variables are used at the end of the computation to show final message.

jshaughn: OK, then perhaps just the log message so we don't log on cycles with no pending data.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
lucasponce: I guess the deferring info is really a large amount of data, on perf scenarios, it's not really interesting I think.
We take the assumption that the ordering is working fine.
I'm going to move them to tracef just in case.
It's harmless and it will be there just if needed.

lucasponce: > Duplicate 'Filtering' in message (also see comment about changing 'Filtering' terminology).

Well, one is on debug and other in warning, content message is similar but not equals.

lucasponce: Oh, I missed the "Filtering Filtering" string :)
Thanks for catching up, even after your review I have to read it twice :)

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -155,36 +157,40 @@ public void fire() {
...
-                        "[" + pendingEvents.size() + "] events.");
-            }
-
-            TreeSet<Data> batchData = new TreeSet<Data>(pendingData);
+            TreeSet<Data> batchData = new TreeSet<>(pendingData);
             Data previousData = null;
 
             pendingData.clear();
 
+            long startFiltering = System.currentTimeMillis();
```
jshaughn: If you don't mind, please change timing variables to 'startBatching' and 'batchingTime'.

PR #234, opened 2016-10-18
HWKALERTS-181 Add clustering information on status endpoint

lucasponce:
Refactor Status logic to have its own service

Basically two things on this PR:
- To have a proper StatusService on the impl, before that the status endpoint was calculated by a simple DefinitionsService call, StatusService might hide and encapsulate other future complex checks.
- Add clustering info on status endpoint.

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/StatusService.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/StatusService.java
@@ -41,10 +41,10 @@ public interface StatusService {
     /**
      * Show additional information about distributed status.
      * In distributed scenarios
-     *  - getDistributedStatus().get("currentNode") will store a string with the identifier of the current node
-     *  - getDistributedStatus().get("members") will store a string with a list comma identifiers of the nodes of the topology
+     *  - getDistributedStatus().get("currentNode") returns a string with the identifier of the current node
+     *  - getDistributedStatus().get("members") returns a string with a list comma identifiers of the topology nodes
      *    at the moment of the call
-     * In standalone scenarios getDistributedStatus() will return an empty map.
+     * In standalone scenarios getDistributedStatus() returns an empty map.
      *
      * @return Map with currentNode and members information for distributed scenarios
      */
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/StatusService.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/StatusService.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/StatusService.java
```diff
@@ -0,0 +1,52 @@
...
+     * @return true if system is running on a distributed scenario.
+     *         false if system is running on a standalone scenario.
+     */
+    boolean isDistributed();
+
+    /**
+     * Show additional information about distributed status.
+     * In distributed scenarios
+     *  - getDistributedStatus().get("currentNode") will store a string with the identifier of the current node
+     *  - getDistributedStatus().get("members") will store a string with a list comma identifiers of the nodes of the topology
```
jshaughn: In the above two comment lines did you mean "will store"  or "will return"?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
```diff
@@ -183,6 +186,19 @@ public boolean isDistributed() {
         return distributed;
     }
 
+    @Override
+    public Map<String, String> getStatus() {
+        if (distributed) {
+            /*
+                Members are dynamic, so on each status request this info is refreshed
+             */
+            status.put("members", cacheManager.getMembers().stream()
```
jshaughn: Block comments are so C :)   Aside from a preference for //, this is pretty cool, it looks like you are able to actually get member info, and not just a member count, that will be very useful, I think.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/service/PartitionManager.java
```diff
@@ -72,6 +73,18 @@
     boolean isDistributed();
 
     /**
+     * Show additional information about partition status.
+     * In distributed scenarios
+     *  - getStatus().get("currentNode") will store a string with the identifier of the current node
+     *  - getStatus().get("members") will store a string with a list comma identifiers of the nodes of the topology
```
jshaughn: same comment about "will store".  Actually, you probably only need the comments here in the interface, not in the impl.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/StatusService.java
lucasponce: Well, I guess I translated here, will return is better.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
lucasponce: I prefer the C syntax for comments, it makes comment more clear for me :-).
Having this extra space makes me to pay better attention to it.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/service/PartitionManager.java
jshaughn: certainly it's harmless.  IDE typically just grabs from the super, so it saves you from having to have them in both places.  Not a big deal, just a note.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/StatusService.java
stefannegrea: Nitpicking here .... but I would drop `will`.

PR #236, opened 2016-10-20
[HWKALERTS-183] Remove xmlpull and xpp3

pgier:
These two transitive deps do not seem to be needed at build time or runtime, so they should
be excluded from the build and the distribution war files.


lucasponce:
Yes, I guess those are used internally at runtime by drools-compiler in the decision-tables import.
Let's see if travis is ok with that.

jshaughn:
Travis is happy so I think we're good to go. Thanks @pgier !

PR #237, opened 2016-10-22
Add missing reactor module that was leading to build errors.

stefannegrea:


jshaughn:
It looks like this makes the OS profile the same as the default, can we just get rid of the openshift profile?

jshaughn:
@stefannegrea, Just to expedite things I'll merge this as-is in order to generate the desired 1.3.1 release.  If you agree that the openshift is redundant, please generate another PR to remove it completely from this pom. Thanks.

stefannegrea:
@jshaughn, I will submit shortly a patch for master branch only to remove the extra profile. In the long run, we might end up removing the openshift profile anyway.

PR #243, opened 2016-10-27
[POC][WIP] [TurnBackNow!] [BewareOfDog!]

jshaughn:
@lucasponce This is a total POC just to look at and talk around.  I figured I'd look at RxJava and see what it may be able to do.  As this stands it does work, the i-tests pass.  If I understand things it will currently do filtering (if necessary, not necessary when sent from metrics) and then return the the caller and do the rest async.


jsanda:
Take this for what it is worth, but in my experience trying to learn something like RxJava which is both functional and reactive is a lot more difficult when you mix it with imperative, non-reactive code. In fact it won't surprise me if you find it to be little more than an obstruction. Start with something small like a method that fetches some records from the database, applies some transformations, fetches related data, etc, and try to implement it using only Rx operators. It will probably take you a while :) You will likely end up code that is much more concise than the imperative version.

jshaughn:
> Take this for what it is worth, but in my experience trying to learn something like RxJava which is both functional and reactive is a lot more difficult when you mix it with imperative, non-reactive code. 

I don't really disagree.  But we're certainly not going to replace existing, working code just for something shiny.  And we're likely not going to use Rx here, either.  But, given the provider/consumer scenario here, I thought it was worth giving it a try as a potential entry point.  And honestly I wanted to just play with it a bit.   It did drop in fairly seamlessly and the async handoff is natural.

But I imply from your comment that you think it's likely a bad idea for use here.

hawkular-alerts-engine/pom.xml
```diff
@@ -88,6 +88,12 @@
       <version>${version.com.google.guava}</version>
     </dependency>
 
+    <dependency>
+      <groupId>io.reactivex</groupId>
+      <artifactId>rxjava</artifactId>
```
jshaughn: @lucasponce I don't have anything against RxJava, and I thought perhaps it was something to consider for the provider/consumer scenario.  This in no way indicates any plan for re-write of existing, working code.   I was just wondering if it's a good solution to *this* problem.  I'm not so sure I understand the "mixed technology" thing, it's just a java API.

jshaughn: > I am going to put this further revision on hold...

Regardless of impl, one thing in the PR that I want to promote is the idea of IncomingDataManager, or basically a component where we can perform various filtering/throttling/massaging of the incoming data in an async way.

lucasponce: > I don't have anything against RxJava
> I'm not so sure I understand the "mixed technology" thing, it's just a java API.

I don't have anything against RxJava, Spring, Netty, Vertx to cite some examples of Java APIs.
My comment was more related to the point that in Java EE some of those networks can overlaps resources and pattern than are available in Wildfly. 
Mixing Java EE and non EE technologies work but at some cost, I tend to maximize the use of the Java EE resources unless there is a blocker/need for additional things.


> Regardless of impl, one thing in the PR that I want to promote is the idea of IncomingDataManager, or basically a component where we can perform various filtering/throttling/massaging of the incoming data in an async way.

I got that. #244 is an approach of this idea with an alternative implementation.

jsanda:
I am not necessarily suggesting that it is a bad idea here. While I think RxJava and reactive/functional is great, it is difficult to grasp, and mixing with imperative, non-reactive code may not help with making a good first impression.

I definitely do not think it is a bad idea to use here :)

jsanda:
> I don't have anything against RxJava, Spring, Netty, Vertx to cite some examples of Java APIs.
My comment was more related to the point that in Java EE some of those frameworks can overlaps resources and pattern that are available in Wildfly and Java EE. 

I was going to avoid starting a debate, but what the hell, it's Friday :)  I haven't used Spring in a long time so I won't comment on it. RxJava, Netty, and Vertx have something in common that separates them from JEE. They are libraries, not frameworks which require you to write and run your code in some container or managed environment. Why do I care about this particular fact? One main reason - testing. Testing RxJava code is easy. In Hawkular Metrics we can easily write tests directly against our service layer which I think this is important to have and do. I do not see the same in Alerts, and I imagine that is at least partly due to the fact that it is not easy to test JEE/EJB code. Arquillian can/maybe/sorta/kinda make things easier if/when you actually get it set up and working. Suffice it to say Arquillian brings with it a steep investment.

I disagree about how much WildFly/JEE offers for a Cassandra based application such as Hawkular Metrics/Alerts. JEE offers a lot in terms of resource management for RDBMS, but it offers none of that for Cassandra. In fact, the DataStax driver handles all aspects of resource management. 

Cassandra does not have transactions, which are one of the biggest selling points of EJB. On that basis alone, I do not a good reason for using EJB with Cassandra. I suppose one could argue that with EJB you have container managed concurrency. Once again, there is no need for it whatsoever since the Cassandra driver is async. If you use the asyc APIs and make your code async, there is a lot less need for thread pools.

Async code requires less resources and offers better performance. There isn't anything in WildFly/JEE to help with writing and building async code. There is stuff in Java 8 like CompletableFuture, but the driver predates Java 8 and was implemented using Guava's Futures. Guava's ListenableFuture is a major improvement over plain Futures, but there are still plenty of shortcomings. In Metrics we have found RxJava to be a much better fit. If we were to rewrite Metrics without RxJava using more imperative code, we would likely have at least twice as much code as we do now. No thanks.

Reactive streams are coming in Java 9 which means we will probably see them in JEE at some point, like 4 or 5 years down the road :)

jshaughn:
I'm going to close this PR, although the discussion is good.  Thanks @jsanda for your feedback!  We will keep an eye on Rx, particularly where we may be able to introduce it in a botton-up use-case.  For now I expect we will not drop this paradigm into the middle of the mix.

PR #244, opened 2016-10-28
Using Java EE concurrent resources

lucasponce:
This is suggestion about how we can implement the same goal using the internal ManagedExecutorService.
It performs same tasks of the first approach with the benefits to not mix technologies.

Also, having the managed-executor-service at configuration level is useful.
Today we are using the default one, but a new one can be created at configuration level if necessary.

Detaulf one could be a good starting point, perf results can give more info to adjust the parameters.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
@@ -54,9 +54,9 @@ import org.hawkular.alerts.api.services.AlertsCriteria;
 import org.hawkular.alerts.api.services.AlertsService;
 import org.hawkular.alerts.api.services.DefinitionsService;
 import org.hawkular.alerts.api.services.EventsCriteria;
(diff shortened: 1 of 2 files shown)


jshaughn:
This looks fine to me.  We could use it as a starting point.  Perhaps what we should explore is to actually make IncomingDataManager the actual Data depot, such that perhaps the engine pulls from here rather than us pushing data into the engine's singleton.  In that way perhaps being able to manage a maximum size of in-memory data, already sorted, protecting against redundant datums, etc...  What do you think?

Also, we'd need similar handling for Events, right?

lucasponce:
I think that in the IncomingDataManager we can put two things:
- Extract the ordering batching from the AlertsEngine to here, so in that case AlertsEngine will received ordered data/events.
- Buffer the data with protection for big buffers (not a single query with a big buffer, but a large amount of requests), on this case we can use some strategy to control overflows and paginate what we send to the AlertsEngine (using ISPN in local model is an ideal candidate as it offers functionality we need in a configurable way).

Perhaps there is more space for this, but with these two topics we can move forward of our goals.

lucasponce:
I have squashed the commits with proper description messages into one.

PR #246, opened 2016-11-01
HWKALERTS-190 IncomingDataManager enhancements

jshaughn:
- Separate incoming data concerns from engine/evaluation concerns
  - Enhance IncomingDataManager to perform more up-front async work
    before passing data on to the engine.
  - Perform partition data filtering/forwarding outside of the engine
- Help object creation/GC by using TreeSet pervasively
- Fix to make sure Data comparison involves tenantId
- Fix synchronization in AlertsEngineImpl to avoid downstream
  ConcurrentModification problems.
- don't do unnecessary .clear()s
- Improvements to DataDrivenGroupCacheManager
  - increase timeout when updating cache so that waiting calls
    don't timeout.
  - apply READ locks where possible, to reduce waiting
- add some better debug logging, and some cleanup
- Add a bit more more warm up time for cleaner test results


jshaughn:
@lucasponce These changes seem to run pretty well.  I spent most of the day running PerfSendITest. Speeds for this PR and master are nearly identical but in the PR we actually do a bit more before forwarding to the engine.  Both master (with fixes, see below) and this PR outperformed the 1.3 release by about 18%.

The code is now in line with your PR from Monday, and the code in master, it forwards the pre-processed data to the engine.  I'm still promoting the idea of handling the data partitioning before the engine gets involved.  Please consider it.

I have another PR with just some fixes for master.  The same fixes are incorporated in this PR, but I thought it might be good to also PR them separately, for easier visibility.  It can be closed if we move forward with this.

lucasponce:
I am going to review it this morning.
I am reluctant to some changes.
I have tested this PR and the cluster test is basically broken on my machine.
But I haven't investigated yet where and why.

lucasponce:
I have found why the cluster test is broken.
It is an issue in the PartitionManagerImpl as it is designed to have a single listener (I will prepare a fix for master to support several ones).

But the idea of promoting the PartitionManager listeners splitted into the AlertEngineImpl and the IncomingDataManagerImpl is not something I like.

Conceptually the distribution is an implementation detail of the AlertEngineImpl, splitting in other component is possible but it breaks the concept to extract this logic from their original component.

I have performed several perf test versus master and in the PR246 about this, and there is not perf advantage in clustering either. 
From runtime perspective those threads are executed from the same internal pool managed from ISPN, so they are still async.

I am still thinking that PartitionManagerListener should remain into the AlertsEngineImpl as same as AlertsEngineCache work as we centralize in the AlertsEngine implementation how to deal when single node/distributed mode.

lucasponce:
I have prepared the fix on #248.

What I can propose you is that I can rebase this PR branch with master and I can try to re-organize the aspect I don't see clear.

My major concern is the separation of the PartitionManager listeners.
I see that from IncomingDataManager perspective, distribution should be transparent, something hidden in the implementation of the AlertsEngine.

The rest of the ideas are good to me.

(There are a couple of minor things to comment, but I guess that the main discussion is on this topic).

jshaughn:
Yes, at the end of the day yesterday I ran the ClusterITest and saw a failure, but did not have a chance to investigate.  Thanks for pursuing that failure.  I will move forward with your recommendations and then we can meet.

jshaughn:
> But the idea of promoting the PartitionManager listeners splitted into the AlertEngineImpl and the IncomingDataManagerImpl is not something I like. Conceptually the distribution is an implementation detail of the AlertEngineImpl, splitting in other component is possible but it breaks the concept to encapsulate this logic on its original component. I have performed several perf test versus master and in the PR246 about this, and there is not perf advantage in clustering either.  From runtime perspective those threads are executed from the same internal pool managed from ISPN, so they are still async. I am still thinking that PartitionManagerListener should remain into the AlertsEngineImpl as same as AlertsEngineCache work as we centralize in the AlertsEngine implementation how to deal when single node/distributed mode.

How about this:  What if we go back to having AlertsEngineImpl do the consolidated partition management.  But also to make the AlertsEngineCache available via a getter.  And let IncomingDataManager continue to do the data *handling*.  Meaning, the forwarding of necessary data to the cache, and the node-specific filtering?   This may be a "best of both worlds" approach.  It allows IncomingDataManager to:
- full encompass data concerns
- continue to perform incoming data requests quickly
- perform more work before forwarding the incoming data

And it consolidates partition management in AlertsEngineImpl.

lucasponce:
This is the code of AlertsEngineImpl:

sendData() is invoked into the async thread, so in terms of where the node filtering is performed has not impact in the computational aspect. It is the same thread on the same stack.

But exposing this into the IncomingDataManager class at the end breaks the "hidden" feature of maintaining the distribution as part of the engine (even there is a filtering data performed).

```java
    public void sendData(Collection<Data> data) {
        if (data == null) {
            throw new IllegalArgumentException("Data must be not null");
        }
        addPendingData(data);
        if (distributed) {
            partitionManager.notifyData(data);
        }
    }
    private void addPendingData(Collection<Data> data) {
        Collection<Data> filteredData = data;
        if (distributed) {
            filteredData = new ArrayList<>(data);
            for (Iterator<Data> i = filteredData.iterator(); i.hasNext(); ) {
                Data d = i.next();
                if (!alertsEngineCache.isDataIdActive(d.getId())) {
                    i.remove();
                }
            }
        }
        synchronized (pendingData) {
            pendingData.addAll(filteredData);
        }
    }
```

I know that what you are proposing is to move this 

```java
            filteredData = new ArrayList<>(data);
            for (Iterator<Data> i = filteredData.iterator(); i.hasNext(); ) {
                Data d = i.next();
                if (!alertsEngineCache.isDataIdActive(d.getId())) {
                    i.remove();
                }
```

into the IncomingDataManagerImpl but for that we need to offer visibility of the distribution status to this component.

I don't see benefits on that:
- Computational speaking, it is using the same thread.
- Code wise, is exposing the IncomingDataManagerImpl to deal with "distributed" status, instead to leave that into the AlertingEngineImpl and make it transparent.

But I am not blocking this, if you want to perform the node filtering on the IncomingDataManagerImpl I'd ask to expose the distributed flag from the AlertsEngine in a similar as the cache.

But my eyes still tell me that this is an implementation detail of the AlertsEngineImpl class :-).

jshaughn:
I've pushed changes base on our off-PR IRC conversation.  I've re-encapsulated partition-awareness into AlertsEngineImpl, as you prefer.  And I've allowed concurrency in AlertsEngine.sendData and .sendEvents now that we're on the same page with Singleton behavior.   It seems perhaps even my original PR was being blocked somewhat by sendData being write-locked,  if I'm correct in my observations, I'm seeing about a 15% faster throughput with the PR now, as compared with master.  On the down side, the cluster itest is failing and I'm not yet sure what the issues is.

lucasponce:
Thanks Jay.
I have started to investigate the cluster scenario failures.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -720,29 +759,38 @@ private int checkMissingStates() {
...
+        It does not need to be re-propagated.
+
+        We allow concurrent threads to make this call in order to process distributed data in parallel. We
+        use synchronized blocks to protect pendingData.
      */
     @Override
+    @Lock(LockType.READ)
     public void onNewData(Data data) {
-        addPendingData(data);
+        sendData(data);
```
lucasponce: Here it is the problem.
`onNewData` receives the notification of the distribution of the data.

But if we invoke to
```java
    public void sendData(TreeSet<Data> data) {
        if (data == null) {
            throw new IllegalArgumentException("Data must be not null");
        }
        if (data.isEmpty()) {
            return;
        }

        addData(data);

        if (distributed) {
            partitionManager.notifyData(new ArrayList<>(data));
        }
    }
```
Then we are looping again the data into the cluster, which is not good.

So, `onNewData` should just add the received data into the pendingData for this node.

The cluster error was generated due duplicated data in a node coming from others nodes.

lucasponce: Confirmed.
A patch like
```
    public void onNewData(Data data) {
-        sendData(data);
+        TreeSet<Data> newData = new TreeSet<>();
+        newData.add(data);
+        addData(newData);
     }
 
     @Override
     public void onNewData(Collection<Data> data) {
-        sendData(new TreeSet<>(data));
+        addData(new TreeSet<>(data));
     }
 
     /*
@@ -785,12 +787,14 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
      */
     @Override
     public void onNewEvent(Event event) {
-        sendEvent(event);
+        TreeSet<Event> newEvents = new TreeSet<>();
+        newEvents.add(event);
+        addEvents(newEvents);
     }
 
     @Override
     public void onNewEvents(Collection<Event> events) {
-        sendEvents(new TreeSet<>(events));
+        addEvents(new TreeSet<>(events));
     }
```
Fixes the cluster env again.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -469,92 +477,122 @@ private void removeTrigger(Trigger trigger) {
         }
     }
 
+    // We allow concurrent threads to make this call in order to process distributed data in parallel. We
+    // use synchronized blocks to protect pendingData.
     @Override
-    public void sendData(Collection<Data> data) {
+    @Lock(LockType.READ)
+    public void sendData(TreeSet<Data> data) {
```
lucasponce: Minor.
I wonder if we can leave the Collection<Data> as the signature of the method.
As we have an @Override, this will fail on the IDE checking that it is not overriding the same method.
Or change the signature on the API.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -119,7 +119,7 @@ public void addData(Data data) {
     }
 
     @Override
-    public void addData(Collection<Data> data) {
+    public void addData(TreeSet<Data> data) {
```
lucasponce: Minor.
Same here.
I don't have any preference if we maintain Collection<Data> or the TreeSet<Data> in the signature, but as it is `@Override` it should be changed in the interface also.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -129,7 +129,7 @@ public void addEvent(Event event) {
     }
 
     @Override
-    public void addEvents(Collection<Event> events) {
+    public void addEvents(TreeSet<Event> events) {
```
lucasponce: Minor.
Same as commented about signatures.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/IncomingDataManagerImpl.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    private Collection<Event> filterIncomingEvents(IncomingEvents incomingEvents) {
+        Collection<Event> events = incomingEvents.getIncomingEvents();
+        events = incomingEvents.isRaw() ? dataIdCache.filterEvents(events) : events;
+
+        return events;
+    }
+
+    private void enforceMinReportingInterval(TreeSet<Data> orderedData) {
```
lucasponce: I am testing this one.
I think there is something not working fine.
After the fix proposed, the cluster is working but the enforceMinReportingInterval is not filtering as expected.
I am debugging it, I will post the insights.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
jshaughn: Arg!  Thanks!  I was tired, I guess.  That is the whole point of having the private "addData" method :)

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
jshaughn: Hmm, I did already change the signature of the API, not sure why you see an IDE issue.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/IncomingDataManagerImpl.java
lucasponce: Yes.
There is a gap in the design.
As commented in the IRC, this design works great when we have a single call for sendData() with close datapoints.
But if we have multiple datapoints in multiple calls, we might need to apply same filter at consumer side.
Both filtering make sense, as one filters in the producers side, it saves time on the consumers one.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -164,77 +168,115 @@ public void fire() {
...
+            log.warnf("Batching Data [%s] took [%s] ms exceeding [%s] ms",
+                    batchData.size(), batchingTime, PERF_BATCHING_THRESHOLD);
+        }
+    }
+
+    private void batchEvents() {
+        TreeSet<Event> batchEvents;
+        long startBatching;
+
+        synchronized (pendingEvents) {
```
lucasponce: Same comment.
This should be executed from a single thread, no need to `synchronized` here.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/RulesEngineTest.java
```diff
@@ -1769,14 +1776,14 @@ public void autoResolveTest() {
...
 
-        datums.add(Data.forAvailability("tenant", "AvailData-01", 1, AvailabilityType.DOWN));
-        datums.add(Data.forAvailability("tenant", "AvailData-01", 2, AvailabilityType.UNAVAILABLE));
-        datums.add(Data.forAvailability("tenant", "AvailData-01", 3, AvailabilityType.UP));
-        datums.add(Data.forAvailability("tenant", "AvailData-01", 4, AvailabilityType.DOWN));
-        datums.add(Data.forAvailability("tenant", "AvailData-01", 5, AvailabilityType.DOWN));
-        datums.add(Data.forAvailability("tenant", "AvailData-01", 6, AvailabilityType.DOWN));
-        datums.add(Data.forAvailability("tenant", "AvailData-01", 7, AvailabilityType.DOWN));
-        datums.add(Data.forAvailability("tenant", "AvailData-01", 8, AvailabilityType.UP));
+        datums.add(Data.forAvailability("tenant", "AvailData-01", 1000, AvailabilityType.DOWN));
```
lucasponce: Wow. Not sure if we have a test without sequential timestamp :)

lucasponce: I admit that those sequential were useful for debugging :-)

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/IncomingDataManagerImpl.java
jshaughn: +1, as we discussed, there is an issue.  I've pushed changes to address the problem.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/DroolsRulesEngineImpl.java
```diff
@@ -114,23 +122,19 @@ public void addFacts(Collection facts) {
     }
 
     @Override
-    public void addData(Data data) {
-        pendingData.add(data);
-    }
-
-    @Override
+    @Lock(LockType.READ)
```
jshaughn: OK, perhaps I am getting synchronized-crazy :)  Let me review...

jshaughn:
Hola @lucasponce, I think this is now ready for final consideration.  I believe this satisfies all of the previous review comments and discussions and I have no more pending work unless there are further comments.

I've run the perf and cluster i-tests and both seem good.  As for PerfSendItest, this PR had a final average send time 45% faster than 1.3.x and 16% faster that the enhancements we already added to master. (based on single runs, so not hugely scientific, but encouraging).

lucasponce:
Thanks Jay.
I am investigating some local errors when I ran the -Pitest profile:
Failed tests: 
  GroupITest.t01_dataDrivenGroupTest:188 expected:<2> but was:<0>

Travis seemed correct, so I'm restarting it to validate if this is local to my environment related of why I am having this.

lucasponce:
Yeap.
Travis now is failing in the same test, with identical result:
```
Failed tests: 
  GroupITest.t01_dataDrivenGroupTest:188 expected:<2> but was:<0>
```
I am still investigating.

lucasponce:
Ok, a missing timestamp triggered this error.
This patch works locally:
```
--- a/hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/GroupITest.groovy
+++ b/hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/GroupITest.groovy
@@ -159,12 +159,12 @@ class GroupITest extends AbstractITestBase {
         assertEquals("contextValue2", mt2.context.get("contextName2"));
 
         // Send in avail data to create Source-1 alert
-        jsonData = "[{\"source\":\"Source-1\",\"id\":\"test-ddgroup-avail\",\"timestamp\":" + System.currentTimeMillis() + ",\"value\":\"DOWN\"}]";
+        jsonData = "[{\"source\":\"Source-1\",\"id\":\"test-ddgroup-avail\",\"timestamp\":" + (System.currentTimeMillis() + 1000) + ",\"value\":\"DOWN\"}]";
         resp = client.post(path: "data", body: jsonData);
         assertEquals(200, resp.status)
 
         // Send in avail data to generate Source-2 alert
-        jsonData = "[{\"source\":\"Source-2\",\"id\":\"test-ddgroup-avail\",\"timestamp\":" + System.currentTimeMillis() + ",\"value\":\"DOWN\"}]";
+        jsonData = "[{\"source\":\"Source-2\",\"id\":\"test-ddgroup-avail\",\"timestamp\":" + (System.currentTimeMillis() + 1000) + ",\"value\":\"DOWN\"}]";
         resp = client.post(path: "data", body: jsonData);
         assertEquals(200, resp.status)
```

lucasponce:
I am going to send a small PR to your branch with two things:
- Patch for the GroupITest.
- Introduce an ENV variable for min-reporting-interval.
I guess with these two things work is almost done, locally things seems correct.

lucasponce:
Here it is:
https://github.com/jshaughn/hawkular-alerts/pull/2

jshaughn:
> Ok, a missing timestamp triggered this error.

Good catch, I guess I just got lucky (maybe I should say unlucky) last night!  Fixing...

PR #248, opened 2016-11-02
Support for multiple listeners in PartitionManagerImpl

lucasponce:
The PartitionManager implementation was designed with a single listener per type.
But re-registering overwrite the previous listener and could create unexpected behaviors.

This should be addressed allowing multiple listeners as it is done on this PR.


jshaughn:
LGTM. I restarted travis due to a failure, but it seems good, I will merge when green.

On a tangential note, it might be nice id we simplify our API and get rid of the single Data and single Event "convenience" variants, and just always use Collections.

PR #251, opened 2016-11-04
HWKALERTS-191 Prevent excessive cache updates

jshaughn:
- batch update requests if the cache is already updating
- add more context to definition change events
  - note, unused now but I did the work and it doesn't
    hurt anything. So, I think we may as well keep it.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
@@ -16,7 +16,7 @@
  */
 package org.hawkular.alerts.engine.cache;
 
-import static org.hawkular.alerts.api.services.DefinitionsEvent.Type.CONDITION_CHANGE;
+import static org.hawkular.alerts.api.services.DefinitionsEvent.Type.TRIGGER_CONDITION_CHANGE;
 import static org.hawkular.alerts.api.services.DefinitionsEvent.Type.TRIGGER_REMOVE;
 
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


jshaughn:
@lucasponce Please review for potantial inclusion in the next release.

lucasponce:
@jshaughn minor one, but could you add the HWKALERTS-191 in the commit message as well ?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -82,64 +85,91 @@ public void init() {
...
-    private synchronized void updateActiveIds() {
+    // Just run updateCache one time if multiple requests come in while an update is already in progress...
+    private void requestCacheUpdate() {
+        log.debug("Cache update requested");
+        if (updateRequested) {
+            log.debug("Cache update, redundant request ignored.");
+            return;
+        }
+
+        updateRequested = true;
```
lucasponce: I have one minor doubt, non blocker about if as we pass this object as a listener, if the object is invoked in the tx context of the EJB.
Basically, the listeners are invoked on CassDefinitionsServiceImpl, so they are using a ejb pool for that.
But as the object is passed in "raw" (without container visibility) not sure if this method is invoked with the Lock(WRITE) semantic or not.
In case it is not, several threads might invoke it and modify the updateRequested variable, so on that case perhaps it should be marked as volatile.
But, I'm not 100% sure.
Just leaving this note here.

jshaughn: I think it is a good idea to declare volatile on both 'updateRequested' and 'updating', to keep the most recent values in memory at all times. I will change it.

PR #252, opened 2016-11-07
Move to commons 0.8.0 and match parent pom version. 

jshaughn:
This transitively updates the c* and driver version's to 3.0.9 and 3.0.5, respectively.

@lucasponce It looks like we'll need to update to the latest commons to match Hservices infrastructure (c* driver version).  We'll probably need a release fairly soon, to allow for Hmetrics to do the same.  Either a 1.3.x or a 1.4.0 version, we can discuss which one we prefer.

cc @stefannegrea

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,7 +23,7 @@
   <parent>
     <groupId>org.hawkular</groupId>
     <artifactId>hawkular-parent</artifactId>
-    <version>43</version>
+    <version>49</version>
   </parent>
 
   <groupId>org.hawkular.alerts</groupId>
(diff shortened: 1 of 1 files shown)


lucasponce:
+1

> Either a 1.3.x or a 1.4.0 version, we can discuss which one we prefer.

Our original plan was to reach 1.0.0 with a feature completed and from 1.x to include small enhancements and focus on performance improvements which is where we are focus these days.

I'm more to bump into a 1.4.x version. I should consider performance enhancements as relevant as new features, we can leave the micro version for bugfixes only (I think).

PR #254, opened 2016-11-07
HWKALERTS-193 Split min-reporting-interval config

lucasponce:
Splitting min-reporting-interval config as Data and Events scenarios could be very different.
So it makes sense to configure them by separate.


lucasponce:
@jshaughn  please, can you review it ? Thanks.

PR #256, opened 2016-11-09
HWKALERTS-195 Import should defer notifications to the end

jshaughn:
- add ability to defer a set of notifications
- change listener callback to accept a set of notifications

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
@@ -95,6 +95,15 @@ public class CassDefinitionsServiceImpl implements DefinitionsService {
     private final MsgLogger msgLog = MsgLogger.LOGGER;
     private final Logger log = Logger.getLogger(CassDefinitionsServiceImpl.class);
 
+    // Note, the next two variables hold 'state' only for the duration of a single EJB method. This is
+    // a stateless EJB, different method invocations can happen in different bean instances.  The idea here is
+    // that a method (e.g. importDefinitions()) may be an umbrella for multiple updates.  In that case it is
+    // desirable to only send a listener update one time, at the end, because we want listeners to be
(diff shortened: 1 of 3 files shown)


jshaughn:
@lucasponce please review

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
```diff
@@ -95,6 +95,9 @@
     private final MsgLogger msgLog = MsgLogger.LOGGER;
     private final Logger log = Logger.getLogger(CassDefinitionsServiceImpl.class);
 
+    private Set<DefinitionsEvent> deferredNotifications = new HashSet<>();
```
lucasponce: I see a small problem on this.
CassDefinitionsServiceImpl is stateless, so this state for deferredNotifications is not guaranteed.
As same as deferNotifications flag.
Perhaps just extracting these two states in an auxiliary bean with application scope is just enough for this.

Another implementation alternative is to scope this state into the importDefinitions method, if there is not other potential use.
As a local variable and passing into notifyListenersDeferred(Set events).

jshaughn: The idea was not to carry state between multiple calls into the EJB, but to provide a deferNotification status through one call to import (or, feasibly, any single EJB method that wanted to perform multiple updates).   So, I think this approach is OK.

jshaughn: If you agree with the above, I could certainly add some inline comments to make this more clear in the code.

lucasponce: Well, I think it could create confusion as this state is used internally in importDefinitions() method who performs some calls, but the shared variable is managed at EJB stateless instance.
So yes, if you can add some comments I guess it will help in the future just if we wonder why we put it there.
Thanks.

PR #258, opened 2016-11-10
HWKALERTS-198 custom hawkular-alerts.properties file not recognized

jshaughn:
- it looks like maybe an unintended '.' prefix was leftover from the past


lucasponce:
The "." is on purpose.
Linux/Unix convention is to prefix with "." application configuration files that are set up on $HOME.
Also, metrics used this one:
https://github.com/hawkular/hawkular-metrics/blob/master/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/config/ConfigurableProducer.java#L90
Note the ".metrics.conf" file.
Also it was even documented :)
https://github.com/hawkular/hawkular.github.io/blob/pages/src/main/jbake/content/docs/components/alerts/index.adoc#external-configuration-file
But this "Installation page" was moved in one of the hawkular.org re-organizations.

jshaughn:
Sorry for the confusion. Closing...

PR #259, opened 2016-11-10
HWKALERTS-197 move hawkular-alerts.properties

jshaughn:
- move from engine jar to war dists such that defaults can be
  modified more easily for different distributions.
- disable data-driven group triggers for metrics and services
  distributions, because the primary use case is data forwarded
  from metrics, and that Data will not have 'source' field values.


jshaughn:
@lucasponce Please review.  Note that PR #258 is for a related issue.   But with that other fix I confirmed that without overrides the defaults were applied as expected.  Command line -D overrides worked.  And file override via <user.home>/hawkular-alerts.properties also worked.

lucasponce:
LGTM

PR #260, opened 2016-11-14
HWKALERTS-199 Improve CacheManager updates

lucasponce:
HWKALERTS-196 Improve perf examples for CRUD scenarios

This PR implement some enhancements on the CacheManager algorithm.
In an initial step, it fetches from backend the triggersIds/dataIds relation.
Using listeners, it maintains a triggerId -> dataIds relation used to update the published cache.
Now, the published cache holds a dataIds -> triggerIds relation.
On this way, we can maintain multiple references on N:M dataIds:triggerIds.

Improvements are good, PerfCrudITest shows that penalties queries have been minimized.

Please @jshaughn, let's start looking at this PR.

Note that I focus on the CacheManager only.
I will the DataDrivenGroupCacheManager in an following PR.

But I think we can review this one.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
@@ -19,6 +19,7 @@ package org.hawkular.alerts.engine.cache;
 import static org.hawkular.alerts.api.services.DefinitionsEvent.Type.TRIGGER_CONDITION_CHANGE;
 import static org.hawkular.alerts.api.services.DefinitionsEvent.Type.TRIGGER_REMOVE;
 
+import java.io.Serializable;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
@@ -38,12 +39,17 @@ import org.hawkular.alerts.api.services.PropertiesService;
 import org.hawkular.alerts.engine.log.MsgLogger;
 import org.hawkular.alerts.filter.CacheKey;
 import org.infinispan.Cache;
-import org.infinispan.CacheSet;
 import org.jboss.logging.Logger;
 
 /**
  * Manages the cache of globally active dataIds. Incoming Data and Events with Ids not in the cache will be filtered
  * away.  Only Data and Events with active Ids will be forwarded (published) to the engine for evaluation.
+ *
+ * This implementation design initialize the cache on each new node.
+ * Definitions events are not propagated into the cluster as there is not a real use case for them and
+ * it can overload the clustering traffic.
+ * A coordinator strategy to initialize will not help either as each new node can became coordinator.
+ * So, it is tradeoff to maintain an extra state or let each node initialize the publish* caches.
  *
  * @author Jay Shaughnessy
  * @author Lucas Ponce
@@ -60,18 +66,20 @@ public class CacheManager {
     private static final String RESET_PUBLISH_CACHE_PROP = "hawkular-alerts.reset-publish-cache";
     private static final String RESET_PUBLISH_CACHE_ENV = "RESET_PUBLISH_CACHE";
 
-    private volatile boolean updateRequested = false;
-    private volatile boolean updating = false;
-
     @EJB
     PropertiesService properties;
 
     @EJB
     DefinitionsService definitions;
 
-    // Note, Only the cache key is being used to today. The values are currently set to "" and reserved for future use.
+    // It stores a list of dataIds used per key (tenantId, triggerId).
+    @Resource(lookup = "java:jboss/infinispan/cache/hawkular-alerts/dataIds")
+    private Cache<TriggerKey, Set<String>> publishDataIdsCache;
+
+    // It stores a list of triggerIds used per key (tenantId, dataId).
+    // This cache is used by CacheClient to check wich dataIds are published and forwarded from metrics.
     @Resource(lookup = "java:jboss/infinispan/cache/hawkular-alerts/publish")
-    private Cache<CacheKey, String> publishCache;
+    private Cache<CacheKey, Set<String>> publishCache;
 
     @PostConstruct
     public void init() {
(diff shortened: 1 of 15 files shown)


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -60,18 +60,18 @@
...
     @EJB
     PropertiesService properties;
 
     @EJB
     DefinitionsService definitions;
 
+    @Resource(lookup = "java:jboss/infinispan/cache/hawkular-alerts/dataIds")
+    private Cache<TriggerKey, Set<String>> dataIdsCache;
+
     // Note, Only the cache key is being used to today. The values are currently set to "" and reserved for future use.
```
jshaughn: Please update this comment with the updated definition.  Maybe mention that the tenantId in the key is applicable to the value (so the values can just be a Set of triggerIds).

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -60,18 +60,18 @@
...
-    private volatile boolean updateRequested = false;
-    private volatile boolean updating = false;
-
     @EJB
     PropertiesService properties;
 
     @EJB
     DefinitionsService definitions;
 
+    @Resource(lookup = "java:jboss/infinispan/cache/hawkular-alerts/dataIds")
```
jshaughn: Maybe add a comment here explaining how this cache is used.  Maybe mention that the tenantId in the key is applicable to the value (so the values can just be a Set of dataIdsIds).

Suggestion: dataIdsCache -> publishDataIdsCache, to better reflect it's part of the publishing mechanism.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -60,18 +60,18 @@
     private static final String RESET_PUBLISH_CACHE_PROP = "hawkular-alerts.reset-publish-cache";
     private static final String RESET_PUBLISH_CACHE_ENV = "RESET_PUBLISH_CACHE";
 
-    private volatile boolean updateRequested = false;
-    private volatile boolean updating = false;
```
jshaughn: If I understand this correctly, these can go away because every update needs to perform its work, whereas before we tried not to repeat the same work.  So, this makes sense to me.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -85,141 +85,150 @@ public void init() {
                 publishCache.clear();
```
jshaughn: Should we also clean the dataIdsCache?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -85,141 +85,150 @@ public void init() {
...
-    private void requestCacheUpdate() {
-        log.debug("Cache update requested");
-        if (updateRequested) {
-            log.debug("Cache update, redundant request ignored.");
-            return;
+    private void removePublishCache(String tenantId, String triggerId, Set<String> dataIds) {
+        if (!isEmpty(dataIds)) {
+            dataIds.stream().forEach(dataId -> {
+                CacheKey cacheKey = new CacheKey(tenantId, dataId);
+                Set<String> triggers = publishCache.get(cacheKey);
```
jshaughn: suggestion: triggers ->triggerIds

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -85,141 +85,150 @@ public void init() {
...
+                if (!isEmpty(triggers)) {
+                    triggers.remove(triggerId);
+                    if (triggers.isEmpty()) {
+                        publishCache.remove(cacheKey);
+                    } else {
+                        publishCache.put(cacheKey, triggers);
```
jshaughn: Is this necessary?  The Set reference is unchanged, only the Set contents changed.  Perhaps it is required to let ispn know for distribution purposes.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -85,141 +85,150 @@ public void init() {
...
-        if (!updating) {
-            updateCache();
+    private void addPublishCache(String tenantId, String triggerId, Set<String> dataIds) {
+        if (!isEmpty(dataIds)) {
+            dataIds.stream().forEach(dataId -> {
+                CacheKey cacheKey = new CacheKey(tenantId, dataId);
+                Set<String> triggers = publishCache.get(cacheKey);
+                Set<String> newTriggers = triggers == null ? new HashSet<>() : new HashSet<>(triggers);
+                newTriggers.add(triggerId);
+                publishCache.put(cacheKey, newTriggers);
```
jshaughn: Why wrap an existing Set in a new HashSet?

Is it necessary to do the put if there existed a Set already? The Set reference is unchanged, only the Set contents changed.  Perhaps it is required to let ispn know for distribution purposes.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
```diff
@@ -85,141 +85,150 @@ public void init() {
...
+            for (Condition c : conditions) {
+                String triggerId = c.getTriggerId();
+                CacheKey cacheKey = new CacheKey(c.getTenantId(), c.getDataId());
+                initialPublish(cacheKey, triggerId);
+                if (c instanceof CompareCondition) {
+                    String data2Id = ((CompareCondition) c).getData2Id();
+                    CacheKey cacheKey2 = new CacheKey(c.getTenantId(), data2Id);
+                    initialPublish(cacheKey2, triggerId);
                 }
             }
```
jshaughn: I guess my big question here is why we don't have to populate dataIdsCache in this initialCacheUpdate.  Don't we need both caches fully defined for the mechanism to work?

hawkular-alerts-rest-tests/src/test/resources/standalone-ha.xsl
```diff
@@ -67,6 +67,9 @@
       <replicated-cache name="publish" mode="ASYNC">
         <transaction mode="BATCH"/>
       </replicated-cache>
+      <replicated-cache name="dataIds" mode="ASYNC">
+        <transaction mode="BATCH"/>
+      </replicated-cache>
       <replicated-cache name="schema" mode="SYNC">
```
jshaughn: The publish and dataIds caches are replicated and that makes sense because it is unknown which node will process a definitions update (right?).   But I think as it is written the CacheManager will perform the initialInitialization on every node.  That does not seem correct, should it be limited to the coordinator?

I suppose an alternative would be to make these local caches but to propogate DefinitionEvents across the cluster.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
lucasponce: Yes, and possible congestion on same key is managed at ISPN level.
So I don't see the need to use them on the new design.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
lucasponce: +1
Additional comments will help to follow better the design.
I'm ok with publishDataIdsCache suggestion.
I will update it.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
lucasponce: Yes.
It is key of the design.
It shows that dataId1 can be referenced on triggerId1 and triggerId2.
When dataId1 references to an empty set it means that the dataId is not longer active so it should be removed.
Also if we delete the reference from triggerId2, the triggersIds should be updated to count down the references of the dataId.
(If I don't have introduce any issue in the logic).

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
lucasponce: Yes, I wanted to force the put() call to avoid side effects on distribution scenarios.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
jshaughn: I believe I understand the design, this was just a question of whether the cacheKey needs to be "re-put", because the value has not changed, the Set is still the same reference.  It's the Set membership that has changed.

hawkular-alerts-rest-tests/src/test/resources/standalone-ha.xsl
lucasponce: Yes.
This design has the price to initialize the cache on each new node.
Propagating the events across the cluster is something I want to avoid, to not load the traffic too much.
I think that Definitions listeners can be maintained at local node, as definitinos have visibility per all nodes and it's unlikely we may need to propagate such events yet.
Coordinator won't help too much here, because normally jgroups makes the new node the new coordinator, so everytime a new node is joining probability to perform this is high.
In the other hand, we might have some kind of flags but it's the same, it's a new state to maintain.
So I think that the tradeoff could be to update the cache at initialization per node.
It's a cheap price to pay comparing to the previous design.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
jshaughn: OK, it is the distribution issue, which is what I was suspecting.

Can you answer the other question, why the new HashSet?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
lucasponce: If I am not wrong, if we don't invoke the put, ISPN is not aware of collections changes.

hawkular-alerts-rest-tests/src/test/resources/standalone-ha.xsl
jshaughn: It's fine to do it one time on each node, I think.  Perhaps there is some possible vulnerability if changes are processed on one node when another is initializing, I'm not sure.

The design decisions as described in your response would be great as comments in the class.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
lucasponce: > Can you answer the other question, why the new HashSet?

Probably we can reuse the existing object.
I had the doubt as I don't remember if the implementation of ISPN detects that is the same object and doesn't trigger the change or if it performs a new hashCode on the value.
I will review it.

hawkular-alerts-rest-tests/src/test/resources/standalone-ha.xsl
lucasponce: I think one thing we can do is to protect this load in a serialized way.
As it is one-step we can lock the entire cache for this operation.
So, no matter when a node joins, no inconsistencies as this cassandra -> cache will be done in atomic bases.

+1
I will update the comments with this.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/CacheManager.java
lucasponce: Good catch.
Yes, initialization is needed.

lucasponce:
Applied most of the comments.
Also added some specific perf scenarios for HWKALERTS-200.

Some local results:

```
[main] INFO groovy.org.hawkular.alerts.rest.PerfImportITest - Import 1000 took 7371 ms
[main] INFO groovy.org.hawkular.alerts.rest.PerfImportITest - Import 2000 took 8099 ms
[main] INFO groovy.org.hawkular.alerts.rest.PerfImportITest - Import 3000 took 12408 ms
[main] INFO groovy.org.hawkular.alerts.rest.PerfImportITest - Import 4000 took 15340 ms
[main] INFO groovy.org.hawkular.alerts.rest.PerfImportITest - Import 5000 took 22122 ms
[main] INFO groovy.org.hawkular.alerts.rest.PerfImportITest - Import 6000 took 29908 ms
```

/cc @FilipB

lucasponce:
Worth to comment that PerfImportITest checks the import endpoints.
These are slower than the individual /trigger endpoints, as an import performs additional validation tasks with conditional deleting/overwritting existing triggers.
PerfCrudITest shows a constant delay using individual endpoints (as on this way import validations are not performed).

jshaughn:
@lucasponce Looks good, ready to merge.  Please squash and add a detailed commit message (like the description in the PR).

lucasponce:
@jshaughn squashed

jshaughn:
I was hoping for a more descriptive comment, I could have otherwise used 'squash and merge' from github.    But let's get this good stuff in!

PR #261, opened 2016-11-15
HWKALERTS-201 Improve scale for getAlerts/getEvents

jshaughn:
For getAlerts:
- remove indexing on status and severity
  - the low cardinality for these fields make them bad for
    indexing as the number alerts grows.  And for filtering the
    number of returned alertIds will be very large.
  - prior to filtering on these fields, fetch the alerts (typically
    already filtered significantly by time) and then filter them
    directly.
For getEvents
  - although category may have low cardinality it is unknown because
    categories are set by the client. So, maintain the index.
  - filter on category either via query+intersection, or manually,
    based on the size of the event set already filtered.
  - make the CriteriaNoQuerySize configurable and also optiponally
    overridden in the criteria.
- Fix issue when initial filter returns 0 entities
- Add some tests


hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/schema/updates/schema-1.4.0.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+schemaChange {
+  version '3.0'
+  author 'jshaughn'
+  tags '1.4.x'
+  cql "DROP TABLE alerts_statuses"
+}
+
+schemaChange {
+    version '3.0'
```
lucasponce: Version on schema changes should be something sequencial (3.0, 3.1).
If not it might overwrite the internal log.
I.e. https://github.com/hawkular/hawkular-metrics/blob/master/core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.21.0.groovy#L26
(5.1, 5.2, 5.3...)

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/PersistenceTest.java
```diff
@@ -685,19 +686,17 @@ public void test0022GroupDampening() throws Exception {
...
     public void test0030BasicTags() throws Exception {
-        logger.info("test0030BasicTags...");
+        log.info("test0030BasicTags...");
 
         Trigger t = definitionsService.getTrigger(TENANT, "trigger-1");
         assertNotNull(t);
 
         Collection<Condition> cs = definitionsService.getTriggerConditions(TENANT, t.getId(), null);
         assertTrue(cs.toString(), cs.size() == 1);
-        Condition c = cs.iterator().next();
```
lucasponce: This is for cleaning unused code ?

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/PersistenceTest.java
```diff
@@ -2380,4 +2379,108 @@ public void test0140EventTags() throws Exception {
         assertEquals("EventTag2Value", event.getTags().get("EventTag2Name"));
     }
 
+    // These tests would be nice in a separate clas but I couldn't figure how to get multiple test classes
+    // to run embedded Cassandra without stepping on each other.
```
lucasponce: Note that to run these tests -Dtest-env=perf is needed.
This is automatically ran by travis CI.

lucasponce:
The changes look good to me.
I have several comments that are minor, nothing important.
So, if you can address them I can merge and rebase as for HWKALERTS-188 I'm touching same methods and I need to reuse these changes.
Thanks.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/PersistenceTest.java
jshaughn: Yes, it was just a cleanup.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/PersistenceTest.java
```diff
@@ -685,19 +686,17 @@ public void test0022GroupDampening() throws Exception {
...
         assertNotNull(t);
 
         Collection<Condition> cs = definitionsService.getTriggerConditions(TENANT, t.getId(), null);
         assertTrue(cs.toString(), cs.size() == 1);
-        Condition c = cs.iterator().next();
 
         Map<String, String> tags = new HashMap<>(t.getTags());
         tags.put("testname", "testvalue");
         t.setTags(tags);
-        Map<String, String> newTag = new HashMap<>(1);
```
jshaughn: Yes, also just a cleanup.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/PersistenceTest.java
jshaughn: Yes, I ran the -Dtest-end=perf many times to execute these tests.   Originally I had introduced two new classes. CassPerfGetTest and PerfGetTest, modeled after CassPersistenceTest and PersistenceTest.   I failed to get these two 'Cass' tests to run together successfully because of issues I had with Cassandra life-cycle.  I feel there is probably a way, but I wasted too much time, and so I merged the new tests into PersistenceTest and removed the new test classes.

hawkular-alerts-engine/src/test/java/org/hawkular/alerts/engine/PersistenceTest.java
```diff
@@ -2380,4 +2379,108 @@ public void test0140EventTags() throws Exception {
...
+    // to run embedded Cassandra without stepping on each other.
+
+    private long nqs100Time = 0L;
+
+    @Test
+    public void test0200GetEventsNQS100() throws Exception {
+        EventsCriteria criteria = new EventsCriteria();
+        criteria.setEndTime(50L);
+        criteria.setCategory("category-a");
+        criteria.setCriteriaNoQuerySize(100);
```
jshaughn: Good comment.  At the moment it is an internal helper (good for testing) because I did not add it to the REST API, but it perhaps could be added to give users control if they were to need it.  There is also the new sysprop for global override, exposing this would also allow for fine-grained override.  We could add later to REST if it seems like a good idea.

lucasponce:
LGTM

PR #264, opened 2016-11-18
HWKALERTS-202 Introduce batch in several Insert/Update/Delete scenarios

lucasponce:
This is still a POC not sure if it is ready for merging.
Testing and tweaking parameters are still in WIP.
But code can be reviewed.
Strategy is to batch those things that make sense, basically within same partition, and prioritizing aspects close to real use cases.
With this setup, -Dtest-env=perf and -Pperf-itest shows a 5% of improvements.

Anyway, I am still studying these changes and testing a good parametrization before to confirm any good result.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
@@ -67,6 +67,7 @@ import org.hawkular.alerts.engine.service.AlertsEngine;
 import org.hawkular.alerts.engine.service.IncomingDataManager;
 import org.jboss.logging.Logger;
 
+import com.datastax.driver.core.BatchStatement;
 import com.datastax.driver.core.BoundStatement;
 import com.datastax.driver.core.PreparedStatement;
 import com.datastax.driver.core.ResultSet;
@@ -90,10 +91,22 @@ public class CassAlertsServiceImpl implements AlertsService {
     private static final String CRITERIA_NO_QUERY_SIZE_ENV = "CRITERIA_NO_QUERY_SIZE";
     private static final String CRITERIA_NO_QUERY_SIZE_DEFAULT = "200";
 
+    /*
+        These parameters control the number of statements used for a batch.
+        Batches are used for queries on same partition only.
+        The real size of the batch may vary as it depends of the size in bytes of the whole query.
+        But on our case and for simplifying we are counting the number of statements.
+     */
+    private static final String BATCH_SIZE = "hawkular-alerts.batch-size";
+    private static final String BATCH_SIZE_ENV = "BATCH_SIZE";
+    private static final String BATCH_SIZE_DEFAULT = "10";
+
     private final MsgLogger msgLog = MsgLogger.LOGGER;
     private final Logger log = Logger.getLogger(CassAlertsServiceImpl.class);
 
     private int criteriaNoQuerySize;
+    private int batchSize;
+    private final BatchStatement.Type batchType = BatchStatement.Type.LOGGED;
 
     @EJB
     AlertsEngine alertsEngine;
@@ -117,13 +130,15 @@ public class CassAlertsServiceImpl implements AlertsService {
     @Resource
     private ManagedExecutorService executor;
 
+    public CassAlertsServiceImpl() {
+    }
+
     @PostConstruct
     public void init() {
         criteriaNoQuerySize = Integer.valueOf(properties.getProperty(CRITERIA_NO_QUERY_SIZE,
                 CRITERIA_NO_QUERY_SIZE_ENV, CRITERIA_NO_QUERY_SIZE_DEFAULT));
-    }
-
-    public CassAlertsServiceImpl() {
+        batchSize = Integer.valueOf(properties.getProperty(BATCH_SIZE,
+                BATCH_SIZE_ENV, BATCH_SIZE_DEFAULT));
     }
 
     public void setSession(Session session) {
@@ -132,6 +147,10 @@ public class CassAlertsServiceImpl implements AlertsService {
 
     public void setExecutor(ManagedExecutorService executor) {
         this.executor = executor;
+    }
+
+    public void setProperties(PropertiesService properties) {
+        this.properties = properties;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


jsanda:
In general it is a bad idea to do unlogged batches across multiple partitions. A single node cluster may be a bit of an exception to that though.

lucasponce:
The batches are performed on queries over single partition. Thus to test
unlogged first.
Queries for multiple partitions are unbatched yet for performance reasons.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -1007,28 +1069,32 @@ private void filterByStatuses(String tenantId, AlertsCriteria criteria, Collecti
...
-        orderedTenantsIds.stream().forEach(tenantId -> {
-            futures.add(executor.submit(() -> {
-                try {
-                    List<Event> tenantEvents = getEvents(tenantId, criteria);
-                    synchronized (events) {
-                        events.addAll(tenantEvents);
+        if (tenantIds.size() == 1) {
+            events.addAll(getEvents(tenantIds.iterator().next(), criteria));
+        } else {
+            TreeSet<String> orderedTenantsIds = new TreeSet<>(tenantIds);
```
jshaughn: trivial: orderedTenantsIds -> orderedTenantIds

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
```diff
@@ -164,6 +170,8 @@ private void readProperties() {
         keyspace = AlertProperties.getProperty(ALERTS_CASSANDRA_KEYSPACE, "hawkular_alerts");
         cassandraUseSSL = Boolean.parseBoolean(AlertProperties.getProperty(ALERTS_CASSANDRA_USESSL,
                 ALERTS_CASSANDRA_USESSL_ENV, "false"));
+        maxQueue = Integer.parseInt(AlertProperties.getProperty(ALERTS_CASSANDRA_MAX_QUEUE,
+                ALERTS_CASSANDRA_MAX_QUEUE_ENV, "9182"));
```
jshaughn: What is this setting for?

jshaughn:
I am not against the driver-based batching, I think it could help our response times.  I think even with the batching introduced here we may still run into the thread issues that have shown up with the change to the 3.0.5 driver.  To avoid that we may still need to limit the number of futures we submit concurrently.  So we may still need to chunk large collections.  For example, if storing 1000 events, we may need to chunk by 100 and then batch those into 10s.

jsanda:
@jshaughn one thing to keep in mind is that batch inserts to the same partition are very efficient. In fact, they are performed as an atomic operation, and this is the best use case for unlogged batches. The problem with unlogged batches across multiple partitions is that the coordinator might timeout waiting for responses from one or more nodes, and you have no way of knowing which writes failed, which means you have to retry the entire batch.

lucasponce:
> @jshaughn one thing to keep in mind is that batch inserts to the same partition are very efficient. In fact, they are performed as an atomic operation, and this is the best use case for unlogged batches. The problem with unlogged batches across multiple partitions is that the coordinator might timeout waiting for responses from one or more nodes, and you have no way of knowing which writes failed, which means you have to retry the entire batch.

@jsanda, as commented earlier. We are not using batching across multiple partitions. Indeed the use introduced here is exactly as you describe.

jsanda:
That's good. In hawkular-metrics, we always batch multiple writes that are for the same partition.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
lucasponce: This setting is related to your previous comment:

> I am not against the driver-based batching, I think it could help our response times. I think even with the batching introduced here we may still run into the thread issues that have shown up with the change to the 3.0.5 driver. To avoid that we may still need to limit the number of futures we submit concurrently. So we may still need to chunk large collections. For example, if storing 1000 events, we may need to chunk by 100 and then batch those into 10s.

I see your point here, and I guess that at some point we should not discard that. But in general, this setting is doing basically the same at driver level. It queues all the queries that driver can not process.

If we batch/cut the max number of parallel async queries will help, but I guess that driver is doing the same.
Point is that controlling this for the application on an efficient way could be costly, as we need to store another singleton state with the max queries per all request (not only per a single request).
And also, reading all best practices with C* for large writtings the recommendation is basically what we are doing, batch in unlogged manner for same partition, and make async queries in general.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
Sorry, I always translate bad all plural expressions :-).

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
jshaughn: I wouldn't suggest that we control this in an "efficient: or complicated way via singleton, etc..  I don't think it's worth the effort.   I think simple facades that can chunk via a configurable chunk size would probably be sufficient, despite not being a guarantee.  Filip is still seeing the busy pool issue despite @jsanda's introduction of the driver settings, so that is still a worry.  But, our use case is different than metrics, and I am OK if we don't pursue further at this time. Also, there may be improved driver handling in the future.

lucasponce: Yeap.
I am not worried, as according of Cassandra driver docs, they are benchmarking 80k request/seconds, so assuming that we can count with a 10% of that and queue-ing efficiently, we can manage the thresholds for single node and scale horizontally.
This PR is not ready finish but my goal is that: to find an acceptable threshold and config that could works as an starting point.

lucasponce:
@jshaughn I addressed some minor feedback.
Can you take a final look ?
Basically, this PR does two things:
- Batches what it makes sense for same partition.
- Adjust the queue of the driver to support overflows of requests.

The recommended way for C* is to send async calls to the driver, and the driver is able to route them into the C* cluster nodes.
I think that doing some adjustments of the queue related alerting usescase (more heavy writes, in general) is good.
C* official docs stands about 80k req/second per node, so even this is quite subjective, it can give us an orientation of orders. 
So I think these improvements will generally help.

We can let QA to run more perf tests and we can try to run future optimizations accordingly.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -117,13 +128,26 @@
...
+        String propTypeValue = properties.getProperty(BATCH_TYPE, BATCH_TYPE_ENV, BATCH_TYPE_DEFAULT);
+        switch (propTypeValue) {
+            case "LOGGED":
+                batchType = BatchStatement.Type.LOGGED;
+                break;
+            case "COUNTER":
+                batchType = BatchStatement.Type.COUNTER;
+                break;
+            default:
+                batchType = BatchStatement.Type.UNLOGGED;
```
These code was for testing and it makes sense to remove it in the final revision.
Good catch.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java
```diff
@@ -90,10 +91,20 @@
     private static final String CRITERIA_NO_QUERY_SIZE_ENV = "CRITERIA_NO_QUERY_SIZE";
     private static final String CRITERIA_NO_QUERY_SIZE_DEFAULT = "200";
 
+    private static final String BATCH_SIZE = "hawkular-alerts.batch-size";
```
jshaughn: Maybe a comment about batching and only using it for single-partition queries ?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassCluster.java
```diff
@@ -118,6 +119,10 @@
     private static final String ALERTS_CASSANDRA_USESSL = "hawkular-alerts.cassandra-use-ssl";
     private static final String ALERTS_CASSANDRA_USESSL_ENV = "CASSANDRA_USESSL";
 
+    private static final String ALERTS_CASSANDRA_MAX_QUEUE = "hawkular-alerts.cassandra-max-queue";
+    private static final String ALERTS_CASSANDRA_MAX_QUEUE_ENV = "CASSANDRA_MAX_QUEUE";
+
```
jshaughn: maybe define an ALERTS_CASSANDRA_MAX_QUEUE_DEFAULT and use that below?

jshaughn:
Just a couple of final comments for you to review.  Please make any final changes, squash with final commit comment, and I will merge. Thanks!

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassAlertsServiceImpl.java

PR #265, opened 2016-11-21
HWKALERTS-204 notifyListenersDeferred should filter correct events

lucasponce:
- publishCache and publishDataIdsCache should work in atomic way


jshaughn:
Good find/fix.

PR #267, opened 2016-11-24
Unify email properties

lucasponce:
- Fixing bug to let properties with suffixes (used in email plugin for properties on lifecycle state or localization).
- Unifying all properties used on email at plugin level. Originally email server was configured at system properties only, now both approaches are merged taking priority configuration at plugin level.
- Adding security support.


jshaughn:
LGTM, @lucasponce, anything pending or is it good to merge?

lucasponce:
No, I think this one is finished from my side.

PR #268, opened 2016-11-28
HWKALERTS-206 Improve global plugins

lucasponce:
This PR is a refactoring of global plugins.
Global plugins were designed to listen for any action and route in a wire tap pattern to a global plugin.
This was a interesting concept as it lets to not define specifically actions on triggers in scenario where we want to be notified for all actions fired.
Previous implementation have two main drawbacks:
- Global plugin required a new plugin implementation.
- Configuration was global and not all flexible as desired.
- Only explicit actions could be "listened" on the wire tap pattern.

With this new approach we maintain the same concept but we introduce several interesting enhancements:
- Global plugin is removed from plugin level and moved to action definition.
- No need to write an special global plugin, any plugin can perform global actions.
- Global actions can be configured with tenants and with all features (filtering by status or using time constraints).
- Global actions can be queried from history.


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/ActionDefinition.java
```diff
@@ -82,44 +82,51 @@
     @JsonInclude
     private String actionId;
 
+    @ApiModelProperty(value = "Flag to indicate this is a global plugin.",
```
jshaughn: "global plugin" -> "global action"

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/action/ActionDefinition.java
```diff
@@ -82,44 +82,51 @@
...
+        this(tenantId, actionPlugin, actionId, false, properties, null, null);
     }
 
-    public ActionDefinition(String tenantId, String actionPlugin, String actionId,
+    public ActionDefinition(String tenantId, String actionPlugin, String actionId, boolean global,
                             Map<String, String> properties, Set<String> states, TimeConstraint calendar) {
         this.tenantId = tenantId;
         this.actionPlugin = actionPlugin;
         this.actionId = actionId;
+        this.global = false;
```
jshaughn: this.global = global;

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/ActionsService.java
```diff
@@ -35,10 +35,10 @@
     /**
      * Generate and send an action to be processed by the plugins architecture.
      *
-     * @param triggerAction Link between a trigger and an action used to generate the action
+     * @param trigger Trigger where it will be generated the action
```
jshaughn: suggestion: "Trigger generating the action"

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -636,6 +636,14 @@ void updateActionPlugin(String actionPlugin, Map<String, String> defaultProperti
     Map<String, Map<String, Set<String>>> getAllActionDefinitionIds() throws Exception;
 
     /**
+     * Get all actions definitions configured in the system.
+     *
+     * @return The existing actions definitions stored in the system. Not null.
```
jshaughn: trivial:  "existing actions definitions" -> "existing action definitions"

jshaughn:
It's great to remove the global plugin modules, much cleaner.  And I think also easier to understand the semantics of a global action.

A couple of comments to respond to, including one in the constructors that looks like a bug.

Note also that the cache addition will affect (minimally) metrics and agent and QE, so we'll need PRs for the two components and also an e-mail in dev-list for Filip.  I can generate the PRs today so they will be ready, and we can send mail after the merge.

lucasponce:
Thanks for the review @jshaughn . Commit amended with the feedback and bug fixes (thanks).
Great if you can prepare the PR to the other components.

PR #269, opened 2016-11-29
Replace srcdeps-maven-plugin with srcdeps core extension

ppalaga:


ppalaga:
@jshaughn or @lucasponce could you please review?

lucasponce:
A preliminar question.
The new script mvnw is only necessary when we are using SRCDEPS links.
Am I right ?

ppalaga:
> The new script mvnw is only necessary when we are using SRCDEPS links.

The only requirement of srcdeps core extension is that the Maven version has to be 3.3.1+.

`mvnw` is just a handy way to ensure that all parties (devs, CI, aliens, ....) are using a Maven version that is guaranteed to work with the present project. `mvnw` script is able to download and install the configured mvn version locally if needed (see [1] for details ). The version can be configured in .mvn/wrapper/maven-wrapper.properties https://github.com/hawkular/hawkular-alerts/pull/269/commits/c1a0df7d6c4569f2c6b9e11b06b1575c73b753f0#diff-de2bfeb3d4c0f4a0b3af8d261a26a21aR18

The idea comes from Gradle where they have a similar `gradlew` script.

That said, nobody is forced to use `mvnw` instead of just `mvn` as long as the version is 3.3.1+.

[1] https://github.com/takari/maven-wrapper#maven-wrapper

lucasponce:
LGTM
Thanks for the explanation @ppalaga

PR #270, opened 2016-12-07
HWKALERTS-209 NelsonRules

jshaughn:
- Introduce a new native condition type, NelsonCondition. This
condition uses "Nelson Rules" to detect "out-of-control" metric
behavior.  The condition establishes baseline metric behavior
based on a configurable metric sample (mean, stdDev) and uses the
baseline to perform analysis on the incoming measurements.
- Introduces apache math3 dep for engine
- Introduce new activeRules field in conditions table

Also:
- fixes for HWKALERTS-208 because I was already adding some  new condition support...
- Fix bug with MISSING conditions in group trigger support
- improve the toString for Alert
- add some missing test support for a couple of condition types

@lucasponce, please review at your leisure. We can talk about it off-PR as well, if it makes sense.

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/NelsonCondition.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/NelsonCondition.java
@@ -1,0 +1,249 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.alerts.api.model.condition;
+
+import java.util.Collection;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+
+import org.hawkular.alerts.api.model.data.Data;
+import org.hawkular.alerts.api.model.trigger.Mode;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+/**
+ * A condition that uses historically collected data to perform a variety of tests for value instability. See:
+ * https://en.wikipedia.org/wiki/Nelson_rules.
+ *
+ * @author Jay Shaughnessy
+ * @author Lucas Ponce
+ */
+@ApiModel(description = "A condition to detect instability based on historical data. + \n" +
+        " + \n" +
+        "From one to all of the defined Nelson rules can be evaluated. See + \n" +
+        "https://en.wikipedia.org/wiki/Nelson_rules for a description of the rules.")
+public class NelsonCondition extends Condition {
+    private static final long serialVersionUID = 1L;
+
+    public enum NelsonRule {
+        Rule1("One point is more than 3 standard deviations from the mean"), //
+        Rule2("Nine (or more) points in a row are on the same side of the mean."), //
+        Rule3("Six (or more) points in a row are continually increasing (or decreasing)."), //
+        Rule4("Fourteen (or more) points in a row alternate in direction, increasing then decreasing."), //
+        Rule5("At least 2 of 3 points in a row are > 2 standard deviations from the mean in the same direction."), //
+        Rule6("At least 4 of 5 points in a row are > 1 standard deviation from the mean in the same direction."), //
+        Rule7("Fifteen points in a row are all within 1 standard deviation of the mean on either side of the mean."), //
+        Rule8("Eight points in a row exist, but none within 1 standard deviation of the mean, "
+                + "and the points are in both directions from the mean.");
+
+        private String description;
+
+        NelsonRule(String description) {
+            this.description = description;
+        }
+
+        public String getDescription() {
+            return description;
+        }
+    }
+
+    private static final int DEFAULT_SAMPLE_SIZE = 50;
+    private static final Set<NelsonRule> DEFAULT_ACTIVE_RULES = EnumSet.allOf(NelsonRule.class);
+
+    @JsonInclude(Include.NON_NULL)
+    private String dataId;
+
+    @ApiModelProperty(value = "Set of NelsonRule to evaluate.",
+            position = 1,
+            example = "All Rules",
+            required = false)
+    @JsonInclude(Include.NON_NULL)
+    private Set<NelsonRule> activeRules;
+
+    @ApiModelProperty(value = "Number of samples used to establish baseline information (mean, standard deviation).",
+            position = 2,
+            example = "50",
+            required = false)
+    @JsonInclude(Include.NON_NULL)
+    private int sampleSize;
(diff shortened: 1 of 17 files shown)


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/NelsonConditionEval.java
```diff
@@ -0,0 +1,181 @@
...
+
+    @ApiModelProperty(value = "Standard Deviation applied to NelsonRules.", position = 2)
+    @JsonInclude(Include.NON_NULL)
+    private Double standardDeviation;
+
+    @ApiModelProperty(value = "Data used to determine violations.", position = 3)
+    @JsonInclude(Include.NON_NULL)
+    private List<Data> violationsData;
+
+    @ApiModelProperty(value = "NelsonRule violations for the data.", position = 5)
```
lucasponce: position = 4 ?

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
```diff
@@ -230,13 +230,8 @@ public Long getLastResolvedTime() {
...
     @Override
     public String toString() {
-        return "Alert{" +
-                "severity=" + severity +
-                ", status=" + status +
-                ", notes=" + notes +
-                ", lifecycle=" + lifecycle +
-                ", resolvedEvalSets=" + resolvedEvalSets +
-                '}';
+        return "Alert [triggerId=" + getTriggerId() + ", severity=" + severity + ", status=" + status + ", tenantId="
```
lucasponce: This change is on purpose ? or just automatically re-create the toString() ?
I remember those fields helped a lot in debugging scenarios (and IIRC this method is invoked on tracing/debugging only contexts).

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/util/NelsonData.java
```diff
@@ -0,0 +1,442 @@
...
+ * It can be overridden for a specific dataId via hawkular-alerts.nelson-rules-sample-size.<dataId> or
+ * HAWKULAR_NELSON_RULES_SAMPLE_SIZE_<dataId>.
+ *
+ * @author Jay Shaughnessy
+ * @author Lucas Ponce
+ */
+public class NelsonData {
+    String NELSON_RULES_SAMPLE_SIZE = "hawkular-alerts.nelson-rules-sample-size";
+    String NELSON_RULES_SAMPLE_SIZE_ENV = "HAWKULAR_NELSON_RULES_SAMPLE_SIZE";
+    String NELSON_RULES_SAMPLE_SIZE_DEFAULT = "50";
```
lucasponce: I am not familiar with the Nelson rules.
One question I might have is about the population number.
I see it is configurable but I was wondering if this constant is something fixed and rarely modified (then the current impl is good) or if it is something worth to be part of the condition definition.
i.e. standard agent collects metrics per second, so this means that means are calculated per minute, perhaps letting the user to define the population size per metric/condition would be more useful (for example, for my CPU I'm interested in peaks per hour not per minute, so then for a dataId would like to increase this to a different range, like 3600).

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -263,6 +266,41 @@ rule Rate
...
+        insert( nelsonData );
+end
+
+// Given current data and a NelsonData for a given dataId, perform a NelsonCondition evaluation.
+rule Nelson
+    when
+        $t  : Trigger( $tenantId : tenantId, $tid : id, $tmode : mode, $tsource : source )
+        $c  : NelsonCondition ( tenantId == $tenantId, triggerId == $tid, triggerMode == $tmode, $did : dataId )
+        $d  : Data( tenantId == $tenantId, source == $tsource, id == $did )
+        $nd : NelsonData( data.tenantId == $tenantId, data.source == $tsource, data.id == $did )
```
lucasponce: Not sure if I have seen that in the code, but the NelsonData is removed when a condition is deleted ?
Just thinking on potential memory left on the engine.

lucasponce:
The PR is interesting and I like this feature.
I have pointed some comments/suggestions that I think it could be worth for the feature.
Also other minor comments on the code to address.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/NelsonConditionEval.java
jshaughn: +1, will fix.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
jshaughn: It was on purpose but it could be revised.  I was actually debugging and was annoyed that the triggerId for the alert was not being shown.  TenantId is important, severity and status were more fields I thought should be shown.  Lifecycle is a little verbose and resolvedEvalSets can be quite verbose.  I can add them back, but perhaps we should actually use log level to decide about the verbosity.  If log.isDebugEnabled { add on those two }.  What do you think?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/util/NelsonData.java
jshaughn: Actually, I started with the sampleSize as being part of the NelsonCondition definition.  But that is a problem because *all* NelsonConditions for the same dataId share one NelsonData tracking fact/object.  So it is not possible to have condition-specific sample size (unless we make a significant change and have one NelsonData per Condition. but this is less efficient and at the moment we have no evidence that we need this level of granularity on sample size).  It is obviously not as convenient, but using sysprops we can set sample size per dataId.  I did take a look on the internet about what was a "proper" sample size for mean and standard deviation and in general I saw the size of 20-50 being recommended, so I used 50 as a default.  Collecting 50 can take some time, so I didn't want to go higher.

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
jshaughn: Good catch, there should be a removal rule if there are no conditions using the NelsonData. I thought about that but looks like I forgot to add it.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/event/Alert.java
jshaughn: Hmm, actually, we don't expose any loggers in the API classes, so perhaps that's not a good approach.  Let me  know if you want me to add those fields back...

jshaughn:
@lucasponce New commit with PR feedback. Take a look.  I can squash if you desire.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/NelsonCondition.java
```diff
@@ -0,0 +1,244 @@
...
+
+    private static final int DEFAULT_SAMPLE_SIZE = 50;
+    private static final Set<NelsonRule> DEFAULT_ACTIVE_RULES = EnumSet.allOf(NelsonRule.class);
+
+    @JsonInclude(Include.NON_NULL)
+    private String dataId;
+
+    @ApiModelProperty(value = "Set of NelsonRule to evaluate. Default = All Rules.", position = 1, required = false)
+    @JsonInclude(Include.NON_NULL)
+    private Set<NelsonRule> activeRules;
```
lucasponce: I see activeRules field has been mapped as a set<text> in cassandra.
This can not be null as C* could throw unexpected errors on that, perhaps some IllegalArgumentException if some null on constructor or setter would be good.

lucasponce:
I put a minor comments. But LGTM.
If you can adress them I can merge it in a while.

PR #272, opened 2016-12-09
HWKALERTS-207 Allow ExternalCondition to be fired on Event submission

lucasponce:
Extending the ExternalCondition to events.
Please @jshaughn, can you review it ?
Thanks.

### DIFF
--- a/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
+++ b/hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
@@ -18,21 +18,23 @@ package org.hawkular.alerts.api.model.condition;
 
 import org.hawkular.alerts.api.model.condition.Condition.Type;
 import org.hawkular.alerts.api.model.data.Data;
+import org.hawkular.alerts.api.model.event.Event;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
+
 /**
- * An evaluation state for an external condition.  Note that external conditions report values as
- * <code>StringData</code>.
+ * An evaluation state for an external condition.  Note that external conditions may report a <code>Data</code> value
+ * or an <code>Event</code>.
  *
  * @author Jay Shaughnessy
  * @author Lucas Ponce
  */
 @ApiModel(description = "An evaluation state for an external condition. + \n" +
-        "Note that external conditions report values as StringData.")
+        "Note that external conditions may report a Data value or an Event.")
 public class ExternalConditionEval extends ConditionEval {
 
     private static final long serialVersionUID = 1L;
@@ -47,10 +49,22 @@ public class ExternalConditionEval extends ConditionEval {
     @JsonInclude(Include.NON_NULL)
     private String value;
 
+    @ApiModelProperty(value = "Event value used for dataId.",
+            position = 2)
+    @JsonInclude(Include.NON_NULL)
+    private Event event;
+
     public ExternalConditionEval() {
         super(Type.EXTERNAL, false, 0, null);
         this.condition = null;
         this.value = null;
+        this.event = null;
+    }
+
+    public ExternalConditionEval(ExternalCondition condition, Event event) {
+        super(Type.EXTERNAL, condition.match(event.getText()), event.getCtime(), event.getContext());
+        this.condition = condition;
+        this.event = event;
     }
 
     public ExternalConditionEval(ExternalCondition condition, Data data) {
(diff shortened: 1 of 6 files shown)


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
```diff
@@ -37,6 +40,10 @@
 
     private static final long serialVersionUID = 1L;
 
+    private enum TypeValue {
+        DATA, EVENT
+    }
```
jshaughn: Do we really need the new enum or to store the Event in a field?  It may be sufficient to just add the new constructor for Event?  You have a String value using Data or Event.

If you really want to keep it, perhaps change from 'TypeValue' to 'ValueType'.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
```diff
@@ -18,12 +18,15 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
+
 /**
  * An evaluation state for an external condition.  Note that external conditions report values as
  * <code>StringData</code>.
```
jshaughn: You can remove "  Note that external conditions report values as <code>StringData</code>."   That is an obsolete comment.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
```diff
@@ -47,16 +54,33 @@
...
         this.condition = null;
         this.value = null;
+        this.typeValue = TypeValue.DATA;
+    }
+
+    public ExternalConditionEval(ExternalCondition condition, Event event) {
+        super(Type.EXTERNAL, condition.match(event.getText()), event.getCtime(), event.getContext());
+        this.condition = condition;
+        this.event = event;
+        this.typeValue = TypeValue.DATA;
```
jshaughn: If you keep the typeValue field I think this should be EVENT.  But as I said above, I don't think you need it.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
```diff
@@ -47,16 +54,33 @@
...
     }
 
     public ExternalConditionEval(ExternalCondition condition, Data data) {
         super(Type.EXTERNAL, condition.match(data.getValue()), data.getTimestamp(), data.getContext());
         this.condition = condition;
         this.value = data.getValue();
+        this.typeValue = TypeValue.EVENT;
```
jshaughn: And if you keep this field, I think this one should be DATA.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
```diff
@@ -75,6 +99,14 @@ public void setValue(String value) {
         this.value = value;
     }
 
+    public Event getEvent() {
+        return event;
+    }
+
+    public void setEvent(Event event) {
+        this.event = event;
+    }
```
jshaughn: If you keep the 'event'  and 'valueType' fields, this setter should probably set the valueType to EVENT for safe deserialization.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
```diff
@@ -97,42 +129,48 @@ public int getConditionSetIndex() {
 
     @Override
     public String getLog() {
-        return condition.getLog(value) + ", evalTimestamp=" + evalTimestamp + ", dataTimestamp=" + dataTimestamp;
+        switch (typeValue) {
```
jshaughn: If you keep the event you could just use its existence here as a discriminator.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
lucasponce: +1

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/condition/ExternalConditionEval.java
lucasponce: I'm not sure. I'm +1 to change to ValueType.
As you suggest, we can use the presence of the String/Event as implicit discriminator about it.
I'm not against that either, just wondering in some corner cases (if any) what happen if there is a null value on both (then perhaps is an IllegalArgumentException).
I'm for maintain temporarily this enum changing to ValueType and we can clean it later once the events alerter is putting in place (we can review it).

lucasponce:
Finally I removed the enum.
I think we have room to introduce it later if there is some use case where is needed.
Added a missing json deserializer task for new event field on ExternalConditionEval.

PR #279, opened 2016-12-23
Aggregate Events Extension/Alerter

lucasponce:
- Defines a new API to add extensions to the Alerting engine
- Make extensions enable/disable by configuration
- Extensions can filter Data/Events before they are processed by the Alerting engine
- Define a new AggregatedEvents Extension that defines a new Alerter
- New Alerter defines a DSL expressions to evaluate aggregated Events defining constraints on aggregated values

@jshaughn this is a preliminar PR for this new feature. I want to work more on tests coverage but the main architecture is coded and I think is good to take a review *after* Christmas break :-)


lucasponce:
Travis is failing with some unrelated issues to this PR.
Some problem to download from nexus .zip artifacts.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -517,6 +534,16 @@ private void addData(TreeSet<Data> data) {
         return filteredData;
     }
 
+    private TreeSet<Data> filterDataExtensions(TreeSet<Data> data) {
```
jshaughn: maybe rename to processDataExtensions()?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -557,6 +588,16 @@ private void addEvents(TreeSet<Event> events) {
         return filteredEvents;
     }
 
+    private TreeSet<Event> filterEventsExtensions(TreeSet<Event> events) {
```
jshaughn: maybe rename to processEventExtensions()?

lucasponce:
> I do find some terminology a little confusing (I know, I have a problem, I always get caught up in terminology :-) ).

:-) I know. 

> Engine "Extensions" is not bad, I thought about some alternatives but none seemed better; "Inteceptors" is clunky. "Filters" is not quite right. "Preprocessors" is pretty accurate but not enough better to merit a change. 

Well, it is some sort of optional "add-on", I guess that any term will need any description. "Extension" sounded reasonable readable in programming context, I guess. 

> From what I see, the fundamental approach is that we can optionally have a set of extensions, potentially ordered (in the future), such that incoming data (or events) are processed by extension 1..N before finally reaching the core engine. Each extension can act on, and then forward data as it sees fit. The extension can alter the data, forward it unchanged, or filter it completely (no forwarding).

Correct. Excellent description.


> One thing that may be more clear for me is to use the term "Extension" a bit more pervasively. For example, in ExtensionsService, instead of"
void addListener(ExtensionEventListener listener);
void addListener(ExtensionDataListener listener);
maybe:
void addExtension(EventExtension extension);
void addExtension(DataExtension extension);
I guess I sort of don't like "Listener" in this use case, it seems unnecessary and not exactly the right fit. And I don't really like the word "filter" in a few places it is used. Anyway, see the more detailed comments and we can go from there...

I guess "Listener"/"Filter" could fit here, but I agree that perhaps this use case is special so I agree with the proposal, I will update the name of the interfaces and the API in a following commit.

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/CepEngineImpl.java
```diff
@@ -0,0 +1,203 @@
...
+    List<Event> results;
+    KieSession kieSession;
+
+    @EJB
+    private AlertsService alertsService;
+
+    @Resource
+    private ManagedExecutorService executor;
+
+    @Lock(LockType.READ)
```
lucasponce: Yes. It is how drools propose it.
Multiple threads can add objects into the KieSession. The KieSession maintains an internal lock for stateful session.

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/CepEngineImpl.java
```diff
@@ -0,0 +1,203 @@
...
+    public void processEvents(Collection<Event> events) {
+        if (kieSession != null) {
+            events.stream().forEach(e -> kieSession.insert(e));
```
lucasponce: Indeed, the caller will use always a TreeSet<Event> as a parameter.
I don't think that it might have a future usecase with a different type, so perhaps here we should propagate the TreeSet as the input parameter.

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/CepEngineImpl.java
```diff
@@ -0,0 +1,203 @@
...
+                alertsService.sendEvents(Arrays.asList(event));
+            } catch (Exception e) {
+                log.error("Error sending result events.", e);
+            }
+        });
+    }
+
+    public void updateConditions(Collection<FullTrigger> activeTriggers) {
+        expression = new Expression(activeTriggers);
+        log.debugf("Rules: \n %s", expression.getDrl());
```
lucasponce: +1

I would maintain the getDrl() as it is more explicit and informative, but I will add the toString() calling the getDrl() method too.

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/CepEngineImpl.java
```diff
@@ -0,0 +1,203 @@
...
+        KieBase kieBase = new KieHelper().addContent(expression.getDrl(), ResourceType.DRL).build(kieBaseConfiguration);
+        KieSessionConfiguration kieSessionConf = KnowledgeBaseFactory.newKnowledgeSessionConfiguration();
+        kieSessionConf.setOption( ClockTypeOption.get( "realtime" ) );
+
+        kieSession = kieBase.newKieSession(kieSessionConf, null);
+        results = new ArrayList<>();
+        kieSession.setGlobal("results", this);
+        kieSession.setGlobal("log", log);
+        kieSession.addEventListener(new CepAgendaEventListener());
+        kieSession.addEventListener(new CepRuleRuntimeEventListener());
```
lucasponce: +1

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/ExpressionManager.java
```diff
@@ -0,0 +1,181 @@
...
+            definitions.registerListener(events -> refresh(events), TRIGGER_CREATE, TRIGGER_UPDATE, TRIGGER_REMOVE);
+            extensions.addListener(this);
+        }
+    }
+
+    private void refresh(Set<DefinitionsEvent> defEvents) {
+        boolean refresh = false;
+        for (DefinitionsEvent defEvent : defEvents) {
+            if (defEvent.getTags() != null && defEvent.getTags().get(TAG_NAME) != null
+                    && defEvent.getTags().get(TAG_NAME).equals(TAG_VALUE)) {
```
lucasponce: +1

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/ExpressionManager.java
```diff
@@ -0,0 +1,181 @@
...
+                // get all of the triggers tagged for this alerter
+                Collection<Trigger> triggers = definitions.getAllTriggersByTag(TAG_NAME, TAG_VALUE);
+                log.infof("Found [%s] Tag [%s, %s] Triggers!", triggers.size(), TAG_NAME, TAG_VALUE);
+
+                Collection<FullTrigger> activeTriggers = new ArrayList<>();
+                for (Trigger trigger : triggers) {
+                    Collection<Condition> conditions = null;
+                    List<Condition> activeConditions = new ArrayList<>();
+                    try {
+                        if (trigger.isEnabled()) {
```
lucasponce: I agree there is not an easy way for this. As there might be some cornercases in both directions.
But I think is better to eliminate this filter and let all triggers to be part of the extension, and deal with the resulting events in the downstream, as it is designed for it.
So, I will update this with the suggested comments.

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/ExpressionManager.java
```diff
@@ -0,0 +1,181 @@
...
+    @Override
+    public TreeSet<Event> processEvents(TreeSet<Event> events) {
+        if (isEmpty(events)) {
+            return events;
+        }
+        List<Event> retained = new ArrayList<>();
+        TreeSet<Event> filtered = new TreeSet<>();
+        for (Event event : events) {
+            if (event.getTags().get(TAG_NAME) != null && event.getTags().get(TAG_NAME).equals(TAG_VALUE)
+                    && event.getContext().get(CONTEXT_PROCESSED) == null) {
```
lucasponce: Extensions might generate events as a result of the pre-processing inference.
These events are marked in the context to filter out of the Extension for potential loops.

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/ExpressionManager.java
jshaughn: +1, I feel we must handle it downstream as we really can't rebuild the rule-base on an enablement change.  If you want to bounce around ideas let me know.

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/ExpressionManager.java
```diff
@@ -0,0 +1,181 @@
...
+                retained.add(event);
+            } else {
+                filtered.add(event);
+            }
+        }
+        if (!retained.isEmpty()) {
+            executor.submit(() -> cep.processEvents(retained));
+        }
+        return filtered;
+    }
```
lucasponce: Well, perhaps the javadoc is pretty basic and needs more description.

Javadoc stands:
```
    /**
     * Process Events received in the engine before it is processed by the Alerting Engine.
     *
     * @param events the Events received by the engine before to be processed
     * @return the resulting Events processed by the extension and sent to the engine
     */
    TreeSet<Event> processEvents(TreeSet<Event> events);
```

So, technically it is correct (from the generic interface perspective). 
The nuances come for this specific extension implementation: it filters a type of events from the input, so the filtered events are correct.

What is missing in the implementation is the description that the filtered events of the extension are used in an inference process and the outcome of that computation might generate additional events that it will be forwarded into the engine again.

I guess the filtered events terminology is ok here (at least for me) but I agree that without this explanation the whole process might be confusing, I will update the comments to make it better readable.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -119,6 +122,11 @@
     private AlertsEngineCache alertsEngineCache = null;
     boolean distributed = false;
 
+    private static final String ENGINE_EXTENSIONS = "hawkular-alerts.engine-extensions";
+    private static final String ENGINE_EXTENSIONS_ENV = "ENGINE_EXTENSIONS";
+    private static final String ENGINE_EXTENSIONS_DEFAULT = "true";
+    boolean engineExtensions;
```

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
We can change the terminology here to process*.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/ExtensionEventListener.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public interface ExtensionEventListener {
+
+    /**
+     * Process Events received in the engine before it is processed by the Alerting Engine.
+     *
+     * @param events the Events received by the engine before to be processed
+     * @return the resulting Events processed by the extension and sent to the engine
+     */
+    TreeSet<Event> processEvents(TreeSet<Event> events);
```
lucasponce: The suggestions for the javadoc are good.
I will incorporate them into the review, perhaps trying to unify both proposals.

lucasponce:
Thanks for your review and suggestions @jshaughn. I will prepare a follow-up commit with the changes.

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/ExpressionManager.java
jshaughn: The fact is that given the jdoc and the impl I still found it quite confusing so I do think something needs to done for clarity.   In general extensions are going to generate some set of forwarded data/events, and it may make it more clear (certainly for me, maybe for others) to make that more explicit in the jdoc and/or impl.

Also, CONTEXT_PROCESSED could use some short inline comment, I think.  Will an event with that context ever be forwarded/persisted?  I don't think so, but if so, it would be nice if that context was scrubbed.

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/AggregatedEventsExtension.java
```diff
@@ -42,21 +42,29 @@
...
 import org.hawkular.alerts.api.services.ExtensionsService;
 import org.hawkular.alerts.api.services.PropertiesService;
 import org.jboss.logging.Logger;
 
 /**
- * Manages the APM expression evaluations and interacts with the Alerts system.
+ * This EventExtension is responsible of the following tasks:
+ *
+ * - It register a DefinitionsListener into the Alerting engine.
+ *   It will process ExternalConditions tagged with HawkularExtension=AggregatedEvents into CEP rules.
```
jshaughn: Very minor and totally optional: AggregatedEvents -> EventAggregation

Just a change from using past tense to something more active.  This may make it easier to talk about, "If enabled, incoming events are first processed by the EventAggregation extension..."

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/AggregatedEventsExtension.java
```diff
@@ -42,21 +42,29 @@
...
+ *
+ * - All events tagged with HawkularExtension=AggregatedEvents will be filtered out and processed asynchronously by
+ *   the extension applying aggregated rules defined in the ExternalCondition expressions.
+ *
+ * - The result of the processing might generate new events that are send into the Alerting engine.
```
jshaughn: send -> sent

hawkular-alerts-engine-extensions/hawkular-alerts-aggregated-events-alerter/src/main/java/org/hawkular/alerts/extensions/AggregatedEventsExtension.java
```diff
@@ -159,13 +170,16 @@ private void refresh(Set<DefinitionsEvent> defEvents) {
         if (isEmpty(events)) {
             return events;
         }
-        List<Event> retained = new ArrayList<>();
+        TreeSet<Event> retained = new TreeSet<>();
         TreeSet<Event> filtered = new TreeSet<>();
         for (Event event : events) {
             if (event.getTags().get(TAG_NAME) != null && event.getTags().get(TAG_NAME).equals(TAG_VALUE)
```
jshaughn: minor: could use that same trick to simplify condition (and add null check for tags):
if ( event.getTags() != null && TAG_VALUE.equals( event.getTags().get(TAG_NAME) && event.getContext().get(CONTEXT_PROCESSED) == null)

lucasponce:
> These changes look good, thanks. So now we still have the issue of handling the disabled triggers downstream, or did I miss that?

I don't think it is necessary. It is done already as if the trigger is disabled the events will be ignored.

lucasponce:
I've sent a final commit.
Let me know if there is more comments, otherwise I can squash and set a proper JIRA code with description before to merge.

jshaughn:
> I don't think it is necessary. It is done already as if the trigger is disabled the events will be ignored.

OK, if you are satisfied then we can keep it as is.  I was afraid maybe we would end up persisting an Event that should not really have been generated.  But maybe that is not actually possible.

I had recommended EventAggregation as opposed to EventsAggregation but the plural works fine as well, I am ready to merge if you want to squash and update the final commit comment.

lucasponce:
> OK, if you are satisfied then we can keep it as is. I was afraid maybe we would end up persisting an Event that should not really have been generated. But maybe that is not actually possible.

The new events generated by the extension use `alertsService.sendEvents(Arrays.asList(event));` instead to add, so none of the incoming or generated events are persisted, only the generated by the trigger itself.

> I had recommended EventAggregation as opposed to EventsAggregation but the plural works fine as well, I am ready to merge if you want to squash and update the final commit comment.

Yeah, I know, I took the license to use the plural, as it is wide used in the terminology.

Let me to prepare a commit with the three commits and the proper code.

lucasponce:
Renaming a module triggered some weird issues with the files and license plugin but it is not fixed.

From my side I think this PR is ready to merge.

PR #283, opened 2017-01-24
Dampening update: set correct TriggerId to dampening object

dkorn:
When updating trigger dampening (also for a group trigger) with a wrong triggerId, the `getCleanDampening()` would generate a wrong dampening.

In addition, in the case of missing DampeningId or TriggerId, only the DampeningId is mentioned in the Bad Argument HTTP error message. This change also suggests a missing TriggerId.

fixes: https://issues.jboss.org/browse/HWKALERTS-215


lucasponce:
Thanks for the fix @dkorn.
It looks good to me.
As soon as Travis CI is fine with it, I will merge it.

PR #285, opened 2017-01-26
HWKALERTS-216 Ensure disabled triggers are removed from the rulebase

jshaughn:
@lucasponce Please review.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
@@ -332,6 +332,7 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
         Trigger trigger = null;
         try {
             trigger = definitions.getTrigger(tenantId, triggerId);
+
         } catch (Exception e) {
             log.debug(e.getMessage(), e);
             msgLog.errorDefinitionsService("Trigger", e.getMessage());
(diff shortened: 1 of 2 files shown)


lucasponce:
There are some problems with this patch and the cluster profile.
I have checked that cluster profile is working good in current master but with this one the test is not passing.
I am investigating root cause.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -372,9 +373,15 @@ private void reloadTrigger(Trigger trigger) {
         if (log.isDebugEnabled()) {
             log.debug("Reloading " + trigger);
         }
+
         // Look for the Trigger in the rules engine, if it is there then remove everything about it
         removeTrigger(trigger);
 
+        // If the trigger is not enabled then just stop after ensuring it's no longer loaded
```
lucasponce: Ok, this validation can be removed.
It is not really necessary and it makes the cluster fails.

In a couple of lines below, there is protection for this:
```
if (trigger != null && trigger.isLoadable())
```
In distributed scenarios, the full trigger is not propagated, only ids, so the refresh to get the full data of the backend is necessary. 

Anyway, I have tested locally this PR without this validation and itest and cluster pass.

jshaughn: +1, I should have run the cluster scenario.  The cluster fails because the enabled flag would always be false before being fetched, if the trigger was set to only tenantId+triggerId.

jshaughn:
@lucasponce The travis log shows a successful build, the failure was in travis cleanup.  I think we can merge.

PR #288, opened 2017-02-01
HWKALERTS-221 Fix Events propagation on cluster environments

lucasponce:
@jshaughn I have splitted the HWKALERTS-218 in order to introduce the priorities changes in 1.5.1.Final.

This one is important as some events tests on cluster were broken due this bug.


jshaughn:
Yes, definitely a problem, definitely merging and backporting...

PR #289, opened 2017-02-01
[HWKALERTS-222] add cassalog verification checks

jsanda:
When we are good with this PR I will open another one for the 1.5 branch assuming we want the changes there as well.


lucasponce:
LGTM

PR #291, opened 2017-02-02
Extend cluster itest

lucasponce:
This PR extend the ITests coverage on cluster profile.

There is a known issue about listeners that is being addressed on HWKALERTS-223 but that will be sent in a separate PR.

This PR have the cluster profile stable.

Please @jshaughn, take a look. Thx.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineCache.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineCache.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -42,7 +42,7 @@ public class AlertsEngineCache {
     /**
      * A cache of the dataIds hold. Used to filter if a data has a dataId on this node or not.
      */
-    private Set<String> activeDataIds;
+    private Set<DataId> activeDataIds;
 
     public AlertsEngineCache() {
         activeDataEntries = new HashSet<>();
@@ -52,12 +52,13 @@ public class AlertsEngineCache {
     /**
      * Check if a specific dataId is active on this node
      *
+     * @param tenantId to check if has triggers deployed on this node
      * @param dataId to check if it has triggers deployed on this node
      * @return true if it is active
      *         false otherwise
      */
-    public boolean isDataIdActive(String dataId) {
-        return dataId != null && activeDataIds.contains(dataId);
+    public boolean isDataIdActive(String tenantId, String dataId) {
+        return tenantId != null && dataId != null && activeDataIds.contains(new DataId(tenantId, dataId));
     }
 
     /**
@@ -67,8 +68,9 @@ public class AlertsEngineCache {
      */
     public void add(DataEntry dataEntry) {
         activeDataEntries.add(dataEntry);
-        if (!activeDataIds.contains(dataEntry.dataId)) {
-            activeDataIds.add(dataEntry.dataId);
+        DataId newDataId = new DataId(dataEntry.getTenantId(), dataEntry.getDataId());
+        if (!activeDataIds.contains(newDataId)) {
+            activeDataIds.add(newDataId);
         }
     }
 
@@ -91,15 +93,16 @@ public class AlertsEngineCache {
             }
         });
         activeDataEntries.removeAll(dataEntriesToRemove);
-        Set<String> dataIdToCheck = new HashSet<>();
+        Set<DataId> dataIdToCheck = new HashSet<>();
         dataEntriesToRemove.stream().forEach(e -> {
-            dataIdToCheck.add(e.getDataId());
+            dataIdToCheck.add(new DataId(e.getTenantId(), e.getDataId()));
         });
-        Set<String> dataIdToRemove = new HashSet<>();
+        Set<DataId> dataIdToRemove = new HashSet<>();
         dataIdToCheck.stream().forEach(dataId -> {
             boolean found = false;
             for (DataEntry entry : activeDataEntries) {
-                if (entry.getDataId().equals(dataId)) {
+                DataId currentDataId = new DataId(entry.getTenantId(), entry.getDataId());
+                if (currentDataId.equals(dataId)) {
                     found = true;
                     break;
                 }
@@ -117,6 +120,58 @@ public class AlertsEngineCache {
     public void clear() {
         activeDataEntries.clear();
         activeDataIds.clear();
+    }
+
+    public static class DataId {
+        String tenantId;
+        String dataId;
+
+        public DataId(String tenantId, String dataId) {
+            this.tenantId = tenantId;
+            this.dataId = dataId;
+        }
+
+        public String getTenantId() {
+            return tenantId;
+        }
+
+        public void setTenantId(String tenantId) {
+            this.tenantId = tenantId;
+        }
+
+        public String getDataId() {
+            return dataId;
+        }
+
+        public void setDataId(String dataId) {
+            this.dataId = dataId;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            DataId dataId1 = (DataId) o;
+
+            if (tenantId != null ? !tenantId.equals(dataId1.tenantId) : dataId1.tenantId != null) return false;
+            return dataId != null ? dataId.equals(dataId1.dataId) : dataId1.dataId == null;
+        }
+
+        @Override
+        public int hashCode() {
+            int result = tenantId != null ? tenantId.hashCode() : 0;
+            result = 31 * result + (dataId != null ? dataId.hashCode() : 0);
+            return result;
+        }
+
+        @Override
+        public String toString() {
+            return "DataId{" +
+                    "tenantId='" + tenantId + '\'' +
+                    ", dataId='" + dataId + '\'' +
+                    '}';
+        }
     }
 
     public static class DataEntry {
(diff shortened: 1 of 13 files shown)


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineCache.java
```diff
@@ -52,12 +52,13 @@ public AlertsEngineCache() {
...
      *
+     * @param tenantId to check if has triggers deployed on this node
      * @param dataId to check if it has triggers deployed on this node
      * @return true if it is active
      *         false otherwise
      */
-    public boolean isDataIdActive(String dataId) {
-        return dataId != null && activeDataIds.contains(dataId);
+    public boolean isDataIdActive(String tenantId, String dataId) {
+        return tenantId != null && dataId != null && activeDataIds.contains(new DataId(tenantId, dataId));
```
jshaughn: Was adding the tenant fixing a bug or added as some sort of formality?  If I recall, ignoring tenant was a conscious decision to avoid the extra object creation and to consolidate the cache entries. Was there an issue with remove being too aggressive?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -467,22 +467,27 @@ private void removeTrigger(Trigger trigger) {
...
             rules.removeFacts(t -> {
                 if (t instanceof Dampening) {
-                    return ((Dampening) t).getTriggerId().equals(triggerId);
+                    return ((Dampening) t).getTenantId().equals(tenantId)
+                            && ((Dampening) t).getTriggerId().equals(triggerId);
                 } else if (t instanceof Condition) {
-                    return ((Condition) t).getTriggerId().equals(triggerId);
+                    return ((Condition) t).getTenantId().equals(tenantId) &&
+                            ((Condition) t).getTriggerId().equals(triggerId);
                 }
```
jshaughn: Wow, yes, this was definitely a problem. Good catch.  I still wonder whether we should do this in rules to simplify the java side.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -521,6 +526,9 @@ private void addData(TreeSet<Data> data) {
         }
 
         synchronized (pendingData) {
+            if (log.isDebugEnabled()) {
+                log.debugf("Adding [%s] to pendingData [%s]", data, pendingData);
+            }
```
jshaughn: trivial: you can remove the isDebugEnabled wrapper

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -575,6 +583,9 @@ private void addEvents(TreeSet<Event> events) {
         }
 
         synchronized (pendingEvents) {
+            if (log.isDebugEnabled()) {
+                log.debugf("Adding [%s] to pendingEvents [%s]", events, pendingEvents);
+            }
```
jshaughn: Same, can remove the wrapper.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -762,19 +773,21 @@ private void handleAutoResolvedTriggers() {
 
     private int checkMissingStates() {
         int numMatchingEvals = 0;
-        for (MissingState missingState : missingStates) {
+        Iterator<MissingState> itMissingStates = missingStates.iterator();
```
jshaughn: It's fine, but why did you change this to use an Iterator?

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
```diff
@@ -270,6 +270,9 @@ public void notifyData(Collection<Data> data) {
         if (distributed) {
             NotifyData nData = new NotifyData(currentNode, data, Data.class);
             Integer key = nData.hashCode();
+            if (log.isDebugEnabled()) {
+                log.debugf("Sending data [%s]", nData);
+            }
```
jshaughn: same, can remove wrapper.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
```diff
@@ -280,6 +283,9 @@ public void notifyEvents(Collection<Event> events) {
         if (distributed) {
             NotifyData nEvent = new NotifyData(currentNode, events, Event.class);
             Integer key = nEvent.hashCode();
+            if (log.isDebugEnabled()) {
+                log.debugf("Sending events [%s]", nEvent);
+            }
```
jshaughn: sorry for the repetition, same...

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
```diff
@@ -740,6 +746,9 @@ private void processNotifyData(NotifyData notifyData) {
             if (!dataListeners.isEmpty() && notifyData.getFromNode() != currentNode) {
                 if (notifyData.getDataCollection() != null) {
                     dataListeners.stream().forEach(dataListener -> {
+                        if (log.isDebugEnabled()) {
+                            log.debugf("processNotifyData [%s]", notifyData);
+                        }
```
jshaughn: same.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineCache.java
lucasponce: If this function doesn't have a tenant filter, it might send wrong data/events to the core engine.
At the end, data/events are discarded but the engine itself, but this limits the flow.
Yes, it is a bug, it doesn't show up in the final result, but it makes the engine to process wrong data.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
lucasponce: I think rules side is safe.
For the moment, this fix is the only needed to fix the problem.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
lucasponce: Because missingStates can be updated concurrently by other thread.
So using a for is not safe on this case.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
lucasponce: +1

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineCache.java
jshaughn: OK, so if in a distributed environment triggers with the same ids and different tenants can be processed on different nodes, then this makes sense.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
jshaughn: Perhaps someday I will create a PR for this to be rule based, I am fine with the change here, though.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
jshaughn: Are you sure Iterator will help this situation?  It's true that Iterator offers the ability to remove the current item but I'm not sure it is really thread safe.

lucasponce: It's handled in two ways.
The updated part is synchronized by the missingStates, and the read part is moved to use an iterator.
So, the iterator itself doesn't solve, that's protected via the synchronized, but moving to an iterator, it check on each element, which then is safe on potential modifications.

jshaughn: I will believe you.  Although, I always thought the for-each construct was just syntactic sugar on top of an underlying Iterator, which would make this change a no-op.  Whatever is happening, it seems subtle in the code and may merit an inline comment about the concurrency issue and how this is safe.

lucasponce: Yes you are right.
The bytecode for each vs iterator is the same.
I first moved to an iterator, and when I discover the problem I protected the update with a synchronize in the other side.
Well, I would maintain it, at least for me it helps to remind that a collection managed by iterator can be updated outside of the loop concurrently.

PR #294, opened 2017-02-08
HWKALERTS-223 Introduce distributed DefinitionsListener

lucasponce:
This PR needs a little bit of context.
Problem:
- DefinitionsListener objects are executed on the local node that performs the definitions update.
- Some alerters/extensions have used the DefinitionsListener mechanism to react when a new definitions is on place.
- As triggers are distributed across nodes, it might happen than the node that update a definitions is not responsible for the execution, and the alerter/extension can miss an update.
- MetricsAlerter and EventsAggregationExtension are affected by this.

Previous work:
To address this problem I have tried several strategies before:

- Strategy 1: Make all DefinitionsListener distributed
Pros: simple at first sight 
Cons: current DefinitionsListener used for Publishing or DataGroups process unnecessary operations, causing pressure on ISPN (and even hitting some Wildfly 10.0.0.Final bugs).

- Strategy 2: Distribute events aligned with distributed triggers
Pros: in theory good idea, the node owner of the trigger is responsible to handle definitions changes
Cons: Needs additional refactoring of the current design, as definitions are transparent across nodes (all nodes can handle definitions updates), this strategy requires some changes in the design, for example, to distribute the node at definitions time, and coordinate in a different way all the listeners in place. So, discarded for the moment. Also, current logic for local listener needs to be revisit.

- Strategy 3: Allow local and distributed DefinitionsListener registration.
Pros: Maintain local listener logic intact. Reduce unnecessary traffic marking which listeners need distribution. 
Cons: The only cons is to decide at registration time when a definitions listener might require distribution (mostly on alerters/extensions).

So, after spent some time with the above, I think Strategy 3 is the better for this problem and tests looks the less overhead in distribution scenarios.

The code basically adds a new cache used for definitions events distribution and extend the registration flag for distributed modes.

Please @jshaughn take a look and let me know any doubt.

Thanks.


hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
```diff
@@ -654,9 +654,30 @@ void updateActionPlugin(String actionPlugin, Map<String, String> defaultProperti
...
+     * A local listener is only invoked in the node that performs the definitions operation.
+     * A distributed listener is invoked in all the nodes of the cluster.
+     *
+     * @param listener the listener to add
+     * @param distributed true indicates listener is distributed
+     *                    false indicates listener is local
+     * @param eventType specify a type to filter when the listener will be invoked
+     * @param eventTypes optional types to filter when the listener will be invoked
+     */
+    void registerListener(DefinitionsListener listener, boolean distributed, Type eventType, Type... eventTypes);
```
jshaughn: Maybe for clarity we should offer an explicit registerDistributedListener() as opposed to using a flag.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/PartitionManagerImpl.java
```diff
@@ -79,6 +81,7 @@
  *          <local-cache name="partition"/>
  *          <local-cache name="triggers"/>
  *          <local-cache name="data"/>
+ *          <local-cache name="definitionsEvents"/>
```
jshaughn: sorry, but maybe lose the double plural and make this "definitionEvents".   Adding caches causes a lot of PRs to update things in various places, but I guess it's unavoidable.

lucasponce:
> This seems to solve the issue of missing notifications for changes on other nodes, but does it help an alerter to run on multiple nodes? I'm not sure it does, right?

Well, today, an alerter can run in multiple times, as the engine filters the resulting data.
For the EventsAggregation Alerter works fine, but the MetricsAlerter would require additional design if we want to scale it and run on every node.

Your comments gave me another idea, I'm revisiting my local branch for the Strategy 2 proposed. 
I think we can introduce a new type of event/listener for the main API DistributedEvent/DistributedListener.

So, in that way, we can unify alerters, they can run on all nodes with the security that only specific triggers will be executed on their node, so, no collision. DefinitionsEvent can still be local and it shouldn't be necessary to add a new cache.

Please, don't merge this PR yet. I am going to re-arrange the strategy 2 local branch to test this idea.

PR #295, opened 2017-02-09
HWKALERTS-223 Propagate DistributedEvents

lucasponce:
This is a development of strategy 2 defined  of #294.

This approach has the benefit that all alerters/extensions can be co-located with the alerting engine and be notified on distribution scenarios.

So, on this way, any alerter can handle only the triggers of its node.

Main change is Alerter should register a DistributedListener instead a DefinitionsListener one.

DefinitionsListener remains local into the node. This design doesn't need additional caches.

Please @jshaughn take a look.


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsContext.java
```diff
@@ -69,4 +129,58 @@ public void registerActionListener(ActionListener listener) {
...
+                .filter(e -> shouldNotify(e.getValue(), notificationTypes))
+                .forEach(e -> {
+                    log.debugf("Notified Listener %s of %s", e.getKey(), notificationTypes);
+                    e.getKey().onChange(notifications.stream()
+                            .filter(de -> e.getValue().contains(de.getType()))
+                            .collect(Collectors.toSet()));
+                });
+        if (!distributed) {
+            distributedListener.stream().forEach(listener -> listener.onChange(mapDistributedEvents(notifications)));
+        }
```
jshaughn: If I understand, this treats a distributedListener like a definitionsListener if we are not in distributed mode?  If so, that is good.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsContext.java
```diff
@@ -69,4 +129,58 @@ public void registerActionListener(ActionListener listener) {
...
+                case TRIGGER_CONDITION_CHANGE:
+                    op = DistributedEvent.Operation.UPDATE;
+                    break;
+            }
+            if (op != null) {
+                events.add(new DistributedEvent(op, tenantId, triggerId));
+            }
+        }
+        return events;
+    }
```
jshaughn: Should we consider just merging DefinitionsEvent and DistributedEvent into one thing?

lucasponce:
> A DefinitionsListener is notified of all changes performed on the registration node.

Yes, DefinitionsListener is local at registration node and it takes fine granularity, triggers, conditions, dampening and actions.

> A DistributedListener is notified of all changes relevant to the registration node.

Slight difference here, meanwhile DefinitionsListener is target for registration operations, DistributedListener is focus on the load/unload operations, like it happens in the core engine.
So, it targets at triggers only and granularity of trigger (conditions/dampenings are part of a trigger update) as in term of flow the reload operation is similar.
Actions are not distributed so, no need to include them on this context.

> Given the changes here it seems an alerter should always be co-located with the alerting nodes because I don't think there is an option to successfully be notified of cluster-wide changes with a single listener.

Yes, this designs approaches to co-located alerter for better scalability. And I think it what it makes more sense. When we grow in nodes of a topology, load should be partitioned and distributed.
For example, for the MetricsAlerter means that no need to have only one active in singleton mode, as when the topology scales up, each node is able to focus on their own triggers.

> If this observation is correct it makes me wonder whether we really should just offer one type of listener, basically the equivalent of the DistributedListener, because a DefinitionsListener in a distributed environment is not useful (I think).

Yes, for alerters, I definitely think that only DistributedListener makes sense, the DefinitionsListener is still needed for several local use cases as global actions registration, published dataId flow and data driven group.

I think both listeners are necessary. I perform some preliminar tasks to converge into one (as it was something I though as well) but there is an unnecessary complexity there. As current tasks for DefinitionsListeners are local and it doesn't makes sense to move them into a DistributedListener.
But on the contrary, Alerters needs a different type of notification and an utility to scale in distributed scenarios.

Perhaps, DefinitionsListener could be something more "internal" or restricted into the engine also, but as they are part for the DefinitionsService and I think it can go on the same interface indicating the local/distributed behavior.

hawkular-alerts-engine-extensions/hawkular-alerts-events-aggregation/src/main/java/org/hawkular/alerts/extensions/EventsAggregationExtension.java
```diff
@@ -106,68 +109,125 @@ public void init() {
...
+
+    private Set<DistributedEvent> optimizeEvents(Set<DistributedEvent> distEvents) {
+        Map<TriggerKey, Operation> map = new HashMap<>();
+        distEvents.stream().forEach(event -> map.put(new TriggerKey(event.getTenantId(), event.getTriggerId()),
+                event.getOperation()));
+        Set<DistributedEvent> optimizedEvents = new HashSet<>();
+        map.entrySet().stream().forEach(entry -> optimizedEvents.add(new DistributedEvent(entry.getValue(),
+                entry.getKey().getTenantId(), entry.getKey().getTriggerId())));
+        return optimizedEvents;
+    }
```
lucasponce: The idea is that in a call of refresh() we might have several events for the same trigger.
For example, first creates the trigger, second update trigger with conditions, another one enables to true, and finally it is removed.
So, effectively the last one is what is needed (on this case for the alerter, others might have a different behaviour, so this logic is placed here for this specific alerter), so for same trigger last operation should prevail.
I am reviewing that there is a pending corner case is that original Set is based on HashSet which doesn't guarantee order, a small change here to use a LinkedHashSet will guarantee this behaviour.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsContext.java
lucasponce: Yes, it is correct, in local modes, DistributedListener are invoked always.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsContext.java
lucasponce: As commented, they are not really the same concept.
DefinitionsEvent has an extended target (actions) and granularity (conditions, dampening) meanwhile DistributedEvent is focus only on triggers and whole granularity as it is designed to load/unload mostly (in the core engine or alerters).

hawkular-alerts-engine-extensions/hawkular-alerts-events-aggregation/src/main/java/org/hawkular/alerts/extensions/EventsAggregationExtension.java
jshaughn: These subtleties merit a method comment, coming back in a year this would be hard to understand.

jshaughn:
OK, thanks for the explanations.  I agree we do need the two listener types.  I would like to see the excellent information presented in your PR responses disseminated into inline doc and jdoc ( DistributedListener, etc).

We may also need to add some best practice information into the the actual user doc for external alerters.  Our two alerters (events and metrics) will use the new listeners, they are co-located by default.  I suppose a 3rd party extension alerter could be co-located if it were to be deployed alongside alerting (or ham), and manage to make use of our java api layer (is it feasible, I can't recall the reach of the recent extensions stuff).  In which case it too could use the DistributedListener.   But there could also be non-co-located 3rd party alerters and for them I guess the best approach is one that just drives off the rest api, probably with intermittent polling for updates and home-grown methods of distributing work, if necessary.  Anyway, my point is only that we need both internal and external doco to help us understand the complexities of our code for distributed envs and also for end users.

lucasponce:
I added additional comments in the doc explaining the new changes introduced.

jshaughn:
@lucasponce Looks good, thanks for the clarifying comments.  I'll merge when green.

jshaughn:
The build succeeded despite travis exiting with code 1, merging...

PR #297, opened 2017-02-22
HWKALERTS-229 Filter unmatching triggerMode conditions on set*Conditions

lucasponce:
A filter to ignore those conditions that does not match with the triggerMode specified as parameter.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/CassDefinitionsServiceImpl.java
@@ -2126,7 +2126,6 @@ public class CassDefinitionsServiceImpl implements DefinitionsService {
         // Get rid of the prior condition set
         removeConditions(tenantId, triggerId, triggerMode);
 
-
         Set<String> dataIds = new HashSet<>();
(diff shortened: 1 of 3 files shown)


jshaughn:
I'm not totally sure I agree with this approach.   I could just as easily think that the path parameter should be applied to the specified condition set.  In fact, I think in other places we do ensure the path params are applied to entities as needed.  Or, I think this should probably be an IllegalArgumentException, it seems to me it could catch a client issue, it seems odd to supply a condition set potentially scattered with conditions of the wrong trigger mode.

As for the filter, perhaps this should be done in the rest handler code as opposed to here.  At this point I would think we'd either apply the requested trigger-mode or throw an IllegalArgumentException.   If you feel strongly the filtering should be done as is, then we should probably filter first and ensure the remaining set is not empty before doing any work.

lucasponce:
I like the suggestion to throw a 400 error in the rest handler if some condition with a different triggerMode is added.
I will re-arrange it in that way.

lucasponce:
PR addressed.
I don't touch the Java API as it might be internal scenarios which might need this as discussed in JIRA.
So, instead an IllegalArgumentException it is responding a HTTP bad request.
Please, @jshaughn feel free to take a look.

jshaughn:
Thanks @lucasponce .

PR #298, opened 2017-02-22
HWKALERTS-230 New endpoint for set all conditions with multiple modes

lucasponce:


lucasponce:
Added two additional endpoints to allow use cases of setting firing and autoresolve conditions in the same request without specifying trigger mode.
Please @jshaughn take a look, thanks.

jshaughn:
These endpoints are good but it may be worth adding to the notes. maybe something like: "This sets the conditions for all trigger modes, replacing existing conditions for all trigger modes."

hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java
```diff
@@ -1050,6 +1051,57 @@ public Response getTriggerConditions(
...
+            Collection<Condition> firingConditions = conditions.stream()
+                    .filter(c -> c.getTriggerMode() == null || c.getTriggerMode().equals(Mode.FIRING))
+                    .collect(Collectors.toList());
+            if (!isEmpty(firingConditions)) {
+                updatedConditions.addAll(definitions.setConditions(tenantId, triggerId, Mode.FIRING, firingConditions));
+            }
+            Collection<Condition> autoResolveConditions = conditions.stream()
+                    .filter(c -> c.getTriggerMode().equals(Mode.AUTORESOLVE))
+                    .collect(Collectors.toList());
+            if (!isEmpty(autoResolveConditions)) {
```
jshaughn: I think we need to still perform the update if the set is empty, otherwise you can't remove all conditions for a particular trigger mode.

lucasponce:
Updated the PR allowing empty conditions sets and modified the documentation.

PR #299, opened 2017-02-27
HWKALERTS-231 Integration with ElasticSearch

lucasponce:
- New ElasticSearch Alerter to schedule queries and fetch documents as events
- New ElasticSearch Plugin to write back actions as documents

I think this PR is in good shape for a review, please @jshaughn when you have some time, take a look on it.
Thanks.


hawkular-alerters/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+
+    <profile>
+      <id>openshift</id>
+      <modules>
+        <module>hawkular-elasticsearch-alerter</module>
+      </modules>
+    </profile>
+  </profiles>
+
+</project>
```
jshaughn: I've been trying to decide whether this should really be part of hAlerting or whether it should be a separate project. It looks to be a generic extension for interacting with an ES db, and that is cool, but it makes me wonder it should be versioned with alerting or on its own schedule.  I suppose this way is easiest, because it interacts with the alerting Java API, it should be deployed against the same version of hAlerting.  Just thinking out loud, I think I am OK with keeping it in the hAlerts umbrella.

hawkular-alerters/hawkular-elasticsearch-alerter/hawkular-elasticsearch-alerter-lib/src/main/java/org/hawkular/alerter/elasticsearch/ElasticSearchAlerter.java
```diff
@@ -0,0 +1,365 @@
...
+
+import org.hawkular.alerts.api.model.trigger.Trigger;
+import org.hawkular.alerts.api.services.AlertsService;
+import org.hawkular.alerts.api.services.DefinitionsService;
+import org.hawkular.alerts.api.services.DistributedEvent;
+import org.hawkular.alerts.api.services.PropertiesService;
+import org.jboss.logging.Logger;
+
+/**
+ * This is the main class of the ElasticSearch Alerter.
```
lucasponce: +1

hawkular-alerters/hawkular-elasticsearch-alerter/hawkular-elasticsearch-alerter-lib/src/main/java/org/hawkular/alerter/elasticsearch/ElasticSearchAlerter.java
```diff
@@ -0,0 +1,365 @@
...
+ *
+ * The ElasticSearch Alerter will listen for triggers tagged with "ElasticSearch" tag. The Alerter will schedule a
+ * periodically query to an ElasticSearch system with the info provided from the tagged trigger context. The Alerter
```
lucasponce: +1

hawkular-alerters/hawkular-elasticsearch-alerter/hawkular-elasticsearch-alerter-lib/src/main/java/org/hawkular/alerter/elasticsearch/ElasticSearchAlerter.java
```diff
@@ -0,0 +1,365 @@
...
+ * will convert ElasticSearch documents into Hawkular Alerting Events and send them into the Alerting engine.
+ *
+ * The ElasticSearch Alerter uses the following conventions for trigger tags and context:
+ *
+ * <pre>
+ *
+ * - [Required]    trigger.tags["ElasticSearch"] = "<timestamp field>"
+ *
+ *   Documents fetched from ElasticSearch needs a date field to indicate the timestamp.
```
lucasponce: +1

hawkular-alerters/hawkular-elasticsearch-alerter/hawkular-elasticsearch-alerter-lib/src/main/java/org/hawkular/alerter/elasticsearch/ElasticSearchAlerter.java
```diff
@@ -0,0 +1,365 @@
...
+ *   This timestamp will be used in the queries to fetch documents in interval basis.
+ *
+ *   Timestamps are expected to follow patterns:
+ *
+ *      "yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ"
+ *      "yyyy-MM-dd'T'HH:mm:ssZ"
+ *
+ *   i.e.   trigger.tags["ElasticSearch"] = "@timestamp"
```
lucasponce: Ok, let's move the timestamps field into the context.
Perhaps in the future we might want to allow others alerters and the tag value can be reserved to discriminate other alerters or even some special functions.

I am not a fan of "Elasticsearch_" or "es_" prefixes, I think that is reasonable to not use them as when a Elasticsearch tag is present is clear that context is related to the alerter.

hawkular-alerters/pom.xml
lucasponce: I think alerters/externsions/plugins should be under the hAlerts repository unless there is a good reason to split them.
Metrics alerter is good to have it on the metrics repo because it introduces complexity on dependencies, and perhaps the semantic and operations are more related to the metrics context rather than alerting.
But on this alerter (and future ones) I think that they are like plugins, so, some library dependency is expected but logic is generic enough to maintain them here.

It is a good point to think if this require a separate lifecycle, well, I expect once is addressed and more mature it will follow the regular lifecyce of alerting core or plugins.

We can revisit this in the future if we see that it has a different requeriments rate rather than alerting, but in any case, as it has to be packaged on a higher metrics.ear, we will have same issues, so I don't see a worth simplification of the process.

lucasponce:
Thanks for your review @jshaughn I am going to address the feedback in a following commit.

hawkular-alerters/hawkular-elasticsearch-alerter/hawkular-elasticsearch-alerter-lib/src/main/java/org/hawkular/alerter/elasticsearch/ElasticSearchAlerter.java
jshaughn: I'm not sure I agree that all context values for the trigger would be those required by the alerter.  It may be a trigger tied to the alerter but the context is still passed on to its generated alerts and events.  Are we sure the developer would not need or want to add domain-specific context outside of the alerter fields? For example: how about context that provides a link to resolution steps, or really, anything else.

lucasponce: I agree. Not the whole context is reserved for the alerter.
I mean that in an alerter trigger is reasonable that the "index" is related to the alerter instead to need tag "es_index".
Adding extra prefixes just reduces the collision probability.

jshaughn: yes, exactly, it identifies and groups the required alerter context from domain-specific context.  You don't see any value in having that convention?

lucasponce: I see the convention, but for the alerter case I have the impression it adds more larger naming harder to deal with.
I wouldn't pursue it on the alerters unless there is collision, personally I think larger names in json make the impression that things are more complex.

Disclaimer: Naming conventions discussions are hard :-).

PR #300, opened 2017-03-01
HWKALERTS-232 Allow dot characters for names in the tag query language

lucasponce:
This is a minor bug fix in the tag query language with additional tests.
Please @jshaughn, take a look.


jshaughn:
Important fix.

PR #301, opened 2017-03-02
HWKALERTS-233 : Add trigger enable/disable to REST API

jshaughn:
- new endpoints triggers/enabled and triggers/groups/enabled
- update existing itests to use new endpoints where applicable
- add new test

### DIFF
--- a/hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/ActionsITest.groovy
+++ b/hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/ActionsITest.groovy
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


jshaughn:
@lucasponce Please review as time permits.

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/ActionsITest.groovy
```diff
@@ -333,9 +333,7 @@ class ActionsITest extends AbstractITestBase {
         assertEquals(1, resp.data.size())
 
         // ENABLE Trigger
-        testTrigger.setEnabled(true);
-
-        resp = client.put(path: "triggers/test-email-threshold", body: testTrigger)
+        resp = client.put(path: "triggers/enabled", query:[triggerIds:"test-email-threshold",enabled:true] )
```
lucasponce: We should maintain tests with the old way to disable (using the whole trigger update) to not miss possible regressions.
I see there were several modifications in tests, I couldn't see if all of them are updated with this one.
Let's maintain explicit tests for non-groups and groups triggers.

jshaughn: I'm not sure I understand the regression risk?  The point of the new endpoints is to reduce risk (and add convenience) when just performing an enable or disable.  I have updated all of these in the existing tests.  I can go back and undo those changes but can you better explain why that is necessary?

lucasponce: Yes. If we eliminate all tests when we enable/disable a trigger using the normal update endpoint, we might miss a explicit test for testing the previous (and currently used) behaviour.
So, those changes are not bad, but we should maintain some tests to enable/disable triggers (non-group and groups) using the current and previous whole trigger update.

jshaughn: I'll change a few back to have a mix although I had updated only the instances where the only change to the already stored trigger was the enabled flag, just a put of that one change.

lucasponce:
Thanks @jshaughn

PR #302, opened 2017-03-07
HWKALERTS-234 Apply current trigger mode to MissingState eval

jshaughn:
@lucasponce review por favor, gracias.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
@@ -402,10 +402,14 @@ public class AlertsEngineImpl implements AlertsEngine, PartitionTriggerListener,
                         }
                     }
                     if (c instanceof MissingCondition) {
+                        // MissingState keeps a reference to the Trigger fact to check active trigger mode
                         MissingState missingState = new MissingState(trigger, (MissingCondition) c);
                         // MissingStates are modified inside the rules engine
-                        missingStates.add(missingState);
-                        rules.addFact(missingState);
+                        synchronized (missingStates) {
+                            missingStates.remove(missingState);
+                            missingStates.add(missingState);
+                            rules.addFact(missingState);
+                        }
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/LifecycleITest.groovy
```diff
@@ -1203,8 +1204,87 @@ class LifecycleITest extends AbstractITestBase {
...
+            // FETCH recent alerts for trigger, there should be 1
+            resp = client.get(path: "", query: [startTime:start,triggerIds:"test-HWKALERTS234-trigger"] )
+            if ( resp.status == 200 && resp.data.size() >= 2 ) {
+                break;
+            }
+            assertEquals(200, resp.status)
+        }
+        assertEquals(200, resp.status)
+        assertEquals(1, resp.data.size())
+        assertEquals("OPEN", resp.data[0].status)
```
lucasponce: I think test can be completed with a couple of steps (low hanging fruit).
- Wait another interval, and validate that there is not additional alerts created (as triggers is in autoresolve mode).
- Send an availability and check that the same alert is now resolved.

lucasponce:
Thanks for the fix @jshaughn it looks good and certainly it was a missing validation.
I suggest to do a couple of cheap additions into the test to ensure this funcionality.
Please, if you can add it, I will merge the PR.
Thx.

hawkular-alerts-rest-tests/src/test/groovy/org/hawkular/alerts/rest/LifecycleITest.groovy
jshaughn: This is not a test for autoresolve, it is a test for the bug case.  After it validates the bug test case it stops.  We have a lot of tests for autoresolve so I didn't want to extend the test further.  I can add it though, no big deal.  Commit coming.

lucasponce: Then perhaps I suggest to increase the 10s to 15s (as condition threshold is 5s). 
That would be enough to ensure that only 1 alert is generated.
Autoresolve is optional, I was thinking not to test autoresolve itself but that the trigger backs correctly to firing mode.

jshaughn: @lucasponce OK, so while adding a little more the test I stumbled over a few more tricky things that ended up taking me most of the day to figure out.  I think things are in good shape now, but please take a good look at the changes.

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
```diff
@@ -455,44 +459,47 @@ public void removeTrigger(String tenantId, String triggerId) {
     }
 
     private void removeTrigger(Trigger trigger) {
+        final String tenantId = trigger.getTenantId();
+        final String triggerId = trigger.getId();
+
+        // If necessary, clean up working memory
```
lucasponce: I see that alertsEngineCache and missingStates are update out of this edge.
Probably is safe, but I wonder motivation, is there any potential case where a trigger to remove from the alerts engine is not found on the drools engine ?.
I ask for this, because if you found a case that this is true, probably this is something related to other problem.

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
```diff
@@ -621,6 +621,156 @@ rule SetFiringModeOnSatisfiedDampening
...
+        not  Trigger( tenantId == $tenantId, id == $tid )
+    then
+        retract( $d )
+
+        if (log != null && log.isDebugEnabled()) {
+            log.debugf("Retract Dampening with no Trigger. %s", $d);
+        }
+end
+
+rule retractAvailabilityCondition
```
lucasponce: I think all conditions rules can be simplified into one:

```
rule retracCondition
    when
        $c : Condition( $tenantId : tenantId, $tid : triggerId )
        not Trigger( tenantId == $tenantId, id == $tid )
    then
        retract( $c )

        if (log != null && log.isDebugEnabled()) {
            log.debugf("Retract Condition with no Trigger. %s", $c);
        }
end
```

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
jshaughn: We had an issue basing this work on the engine state.  In ConditionMatch.drl#SetFiringModeOnSatisfiedDampening we retract the trigger.  Now that the trigger fact is gone we would not go into this conditional and it would leave orphaned facts in the rulebase as well as not execute the cache work of the update of 'missingStates'.

So what I've done in this PR is to treat Trigger removal (via api) and Trigger retraction (via rulebase) independently.  Now, with trigger retraction there are the new rules to clean up the orphaned dampening, condition, and missingState facts, so e don't need to do this in the engine code.  With trigger removal, if the rulebase has the trigger fact, we do need to remove the rulebase facts via code, that is what this conditional is for.

Regardless of whether we are removing the trigger due to the rule-base (like handleAutoResolveTriggers) or API (e.g. REST),  I am *pretty sure* we want to do the alertsEngineCache  and missingStates work.  That is why it is now outside the 'if' block.  Does it make sense?

hawkular-alerts-engine/src/main/resources/org/hawkular/alerts/engine/rules/ConditionMatch.drl
jshaughn: Yes of course!  I am an idiot, for whatever reason I was thinking Drools wouldn't deal with the abstract class. Fixing this...

hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/impl/AlertsEngineImpl.java
lucasponce: Ok, then I see a potential problem
In ConditionMatch.drl#AlertOnSatisfiedDampening 

```
        if ($t.isAutoResolve()) {
            log.debugf("Setting Trigger to AutoResolve Mode! %s", $t);
            retract( $t );
            $t.setMode(Mode.AUTORESOLVE);
            insert( $t );
        }
```

this should be modified to something like

```
        if ($t.isAutoResolve()) {
            log.debugf("Setting Trigger to AutoResolve Mode! %s", $t);
            modify($t) {
               setMode(Mode.AUTORESOLVE)
            }
        }
```
To avoid conflicts with the new rules, as potentially, an autoresolve change could wipe out conditions and dampenings.

As we didn't have explicit `not Trigger` rules tjos was safe, but at the moment of use them, this could be impacted.

jshaughn: I saw that as well, but I don't think so, RHS actions complete before another rule recognition cycle.  I'll check but this is typically a fundamental aspect of the rule cycle.

jshaughn: although, it may be more efficient to use modify($t), I'm not sure.

jshaughn: Although I am sure it is not a problem from the perspective of rule firings, I will make the change to modify() because that seems the appropriate construct.

lucasponce:
Thx @jshaughn

PR #304, opened 2017-03-15
HWKALERTS-237 TimeConstraint needs to support the target time zone

jshaughn:
Let clients specify a time zone for the TimeConstraint such that
timestamp checks are performed against a predictable TZ as opposed to
the server default TZ.

@lucasponce Please review.


lucasponce:
The PR looks fine.
I miss just a couple of lines in the description of the TimeConstraint class where there are a couple of examples in JSON about how to use.
I think just an example with TZ would be great.

lucasponce:
Thanks @jshaughn

PR #309, opened 2017-03-26
HWKALERTS-236 Watchers for events and alerts

lucasponce:
- Add support for filtering alerts by stime field without requiring status
- Introduce a REST stream watchers feature for alerts and events
- Make this feature available for crosstenant operations

I have started two approaches for this feature, a preliminar one (available here https://github.com/lucasponce/hawkular-alerts/tree/watchers-support-without-polling) uses a listener directly from the engine. This one has the advantages that does not require quering the backend, and that seemed a nice approach at the beginning. After some development, I realized that ensuring an ordered timeline could be more complex, as the asynchornous nature of our distributed design. So, I started this second approach which I think it is also interesting, as it offers to watch alerts and events in a timeline.
Guaranteeing that all alerts and events are watched in a strict timeline.

So, with this feature in place, no need to add additional mechanism as tagging to mark alerts as "seen", when a watcher session is open, all alerts and lifecycle states can be observed and client can use timestamps in a reliable way, as when an alert is observed, system guarantees that there is not possibility (*) of observation of an alert before.

(*) Disclaimer: to fully stand this, we should need a strong transactional system between several components, which is something we don't handle, but algorithm design should guarantee this under normal circumstances (i.e. no corrupted scenarios or similar unhandled situations).

### DIFF
--- a/hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/CrossTenantHandler.java
+++ b/hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/CrossTenantHandler.java
@@ -68,6 +68,9 @@ public class CrossTenantHandler {
     @EJB
     AlertsService alertsService;
 
+    @EJB
+    StreamWatcher streamWatcher;
+
     public CrossTenantHandler() {
         log.debug("Creating instance.");
     }
@@ -155,6 +158,14 @@ public class CrossTenantHandler {
                     allowableValues = "Timestamp in millisecond since epoch.")
             @QueryParam("endAckTime")
             final Long endAckTime,
+            @ApiParam(required = false, value = "Filter out alerts with some lifecycle state before this time.",
+                    allowableValues = "Timestamp in millisecond since epoch.")
+            @QueryParam("startStatusTime")
+            final Long startStatusTime,
+            @ApiParam(required = false, value = "Filter out alerts with some lifecycle after this time.",
+                    allowableValues = "Timestamp in millisecond since epoch.")
+            @QueryParam("endStatusTime")
+            final Long endStatusTime,
             @ApiParam(required = false, value = "Return only thin alerts, do not include: evalSets, resolvedEvalSets.")
             @QueryParam("thin")
             final Boolean thin,
@@ -174,7 +185,8 @@ public class CrossTenantHandler {
                 unifiedTagQuery = tagQuery;
             }
             AlertsCriteria criteria = new AlertsCriteria(startTime, endTime, alertIds, triggerIds, statuses,
-                    severities, unifiedTagQuery, startResolvedTime, endResolvedTime, startAckTime, endAckTime, thin);
+                    severities, unifiedTagQuery, startResolvedTime, endResolvedTime, startAckTime, endAckTime,
+                    startStatusTime, endStatusTime, thin);
             Page<Alert> alertPage = alertsService.getAlerts(tenantIds, criteria, pager);
             if (log.isDebugEnabled()) {
                 log.debug("Alerts: " + alertPage);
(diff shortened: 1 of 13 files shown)


hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/StreamWatcher.java
```diff
@@ -0,0 +1,290 @@
...
+ *
+ * @author Jay Shaughnessy
+ * @author Lucas Ponce
+ */
+@Stateless
+public class StreamWatcher {
+    private static final Logger log = Logger.getLogger(StreamWatcher.class);
+    private static final Pager ctimePager;
+    private static final Pager stimePager;
+    private static final long WATCHER_INTERVAL_DEFAULT = 5000;
```
jshaughn: Did you really mean 5000 seconds?

hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/EventsHandler.java
```diff
@@ -344,6 +347,90 @@ We maintain old tags criteria as deprecated (it can be removed in a next major v
         }
     }
 
+    @GET
+    @Path("/watch")
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Watch events with optional filtering.",
+            notes = "Return a stream of events ordered by ctime. + \n" +
+                    " + \n" +
+                    "If not criteria defined, it fetches all evemts stored in the system. + \n" +
```
jshaughn: evemts -> events

hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/CrossTenantHandler.java
```diff
@@ -291,6 +303,215 @@ We maintain old tags criteria as deprecated (it can be removed in a next major v
...
+        return Response.ok(streamWatcher.watchAlerts(tenantIds, criteria, watchInterval)).build();
+    }
+
+    @GET
+    @Path("/watch/events")
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Watch events with optional filtering from multiple tenants.",
+            notes = "Return a stream of events ordered by ctime. + \n" +
+                    " + \n" +
+                    "If not criteria defined, it fetches all evemts stored in the system. + \n" +
```
jshaughn: evemts -> events

hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/StreamWatcher.java
```diff
@@ -0,0 +1,290 @@
...
+        ordering.add(Order.by("stime", Order.Direction.ASCENDING));
+        stimePager = new Pager(0, PageContext.UNLIMITED_PAGE_SIZE, ordering);
+        ordering = new ArrayList<>();
+        ordering.add(Order.by("ctime", Order.Direction.ASCENDING));
+        ctimePager = new Pager(0, PageContext.UNLIMITED_PAGE_SIZE, ordering);
+    }
+
+    public StreamingOutput watchAlerts(Set<String> tenantIds, AlertsCriteria criteria, Long watchInterval) {
+        return output -> {
+            Writer writer = new BufferedWriter(new OutputStreamWriter(output));
```
jshaughn: Does this writer need to be closed? Also below.

jshaughn:
@lucasponce Looks good, Just a couple of comments, I think this is very close to merging.

hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/EventsHandler.java
lucasponce: +1 typo and copy/paste

hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/CrossTenantHandler.java
lucasponce: +1

hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/StreamWatcher.java
lucasponce: The output stream is managed externally and closed by the StreamOutput user.
A close is harmless, anyway, so I will add it.

lucasponce:
@jshaughn updated with feedback.

hawkular-alerts-rest/hawkular-alerts-rest-api/src/main/java/org/hawkular/alerts/rest/StreamWatcher.java
```diff
@@ -0,0 +1,306 @@
...
+                    log.debug("Watcher client disconnected");
+                    try {
+                        writer.close();
+                    } catch (Exception ignored) {}
+                    return;
+                }
+                watchedIds.add(new WatchedId(alert.getId(), alert.getCurrentLifecycle().getStime()));
+            });
+            startWatchTime = startWatchTime == null ? System.currentTimeMillis() : startWatchTime;
+            long sleepWatcher = watchInterval == null ? WATCHER_INTERVAL_DEFAULT : watchInterval * 1000;
```
jshaughn: The reason I questioned WATCHER_INTERVAL_DEFAULT=5000 in the original comment is because I misread this line.  I intepreted it as:
`watchInterval == (null ? WATCHER_INTERVAL_DEFAULT : watchInterval) * 1000`
But actually it is OK, because it is parsed as:
`watchInterval == null ? WATCHER_INTERVAL_DEFAULT : (watchInterval * 1000)`

I think some parentheses would have been more clear, but this is minor.

PR #312, opened 2017-03-29
HWKALERTS-236 Fix issue when having spaces in the install path.

jshaughn:


jshaughn:
This change was already tested by gbaufake, merging...

PR #314, opened 2017-03-30
For the elasticsearch alerter and plugin:

jshaughn:
- defer to metrics ear and remove some config files from the wars
- remove security-constrain from web.xml

@lucasponce Please review.  I'm fairly sure the security constraint decls in web.xml are not supposed to be there but let me know if you think differently.


lucasponce:
It looks good.
Probably I re-use some old layout.
Yes, no need to security constraints on the alerter as it doesn't expose any endpoint.

PR #315, opened 2017-03-31
[HWKALERTS-249] Wait for C* cluster to be up before installing schema

jsanda:
The amount of time to wait is determined by the ```hawkular-alerts.cassandra-retry-timeout``` system property. The number of times to check that the cluster is up is determined by the ```hawkular-alerts.cassandra-retry-attempts``` system property.


lucasponce:
LGTM

jshaughn:
The logic looks OK, I do have one question.  Does this mean we will have to wait for all Cassandra nodes to be up just to run, even if we have no schema updates?  That would be bad, I think, as it would defeat the purpose of replication, I think.

lucasponce:
The logic only happens at initialization phase.
So, yes, is before to initialization even if no changes but happens only at startup.
Perhaps in the future this can be fine tuned in C* configuration but I don't see a bad workaround meanwhile.

jsanda:
> The logic looks OK, I do have one question. Does this mean we will have to wait for all Cassandra nodes to be up just to run, even if we have no schema updates? 

If we do not have schema changes to apply, then I do not think we should wait; however, after Cassalog runs the sys_config table is updated with the version number of Hawkular Alerts. That write could still fail. More importantly, there currently is no API in Cassalog to let us know what if any  updates need to be applied. I am not sure there is a way around this, at least as things exist right now.

> That would be bad, I think, as it would defeat the purpose of replication, I think.

We do not have any replication.

jshaughn:
I'm OK with the PR given the current state of things, but my understanding of larger clusters is that sometimes nodes are down.

lucasponce:
> I'm OK with the PR given the current state of things, but my understanding of larger clusters is that sometimes nodes are down.
+1
I think this is a good comment.
Well, in *runtime* this might happen and it is transparent as logic doesn't wait for all nodes.
But yes, at initialization this demands all nodes up and running for initialization.

PR #317, opened 2017-04-04
HWKALERTS-252 Fix wrong nanoseconds parsed time

lucasponce:
HWKALERTS-251 Fix wrong dot character on Elasticsearch alerter
HWKALERTS-247 Add an Elasticsearch intergation example

Please @jshaughn, could you review ?

This would be necessary for a potential 1.6.x. and metrics 0.26.x.

### DIFF
--- a/hawkular-alerters/hawkular-elasticsearch-alerter/hawkular-elasticsearch-alerter-lib/src/main/java/org/hawkular/alerter/elasticsearch/ElasticsearchAlerter.java
+++ b/hawkular-alerters/hawkular-elasticsearch-alerter/hawkular-elasticsearch-alerter-lib/src/main/java/org/hawkular/alerter/elasticsearch/ElasticsearchAlerter.java
@@ -69,7 +69,7 @@ import org.jboss.logging.Logger;
  *   Documents fetched from Elasticsearch need a date field to indicate the timestamp.
  *   This timestamp will be used in the queries to fetch documents in interval basis.
  *
- *   If there is not defined a specific pattern under the trigger.context["timestamp.pattern"] it will follow the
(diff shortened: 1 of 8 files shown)


hawkular-alerters/hawkular-elasticsearch-alerter/hawkular-elasticsearch-alerter-lib/src/main/java/org/hawkular/alerter/elasticsearch/ElasticsearchAlerter.java
```diff
@@ -69,7 +69,7 @@
  *   Documents fetched from Elasticsearch need a date field to indicate the timestamp.
  *   This timestamp will be used in the queries to fetch documents in interval basis.
  *
- *   If there is not defined a specific pattern under the trigger.context["timestamp.pattern"] it will follow the
+ *   If there is not defined a specific pattern under the trigger.context["timestamp_pattern"] it will follow the
```
jshaughn: Why the change from the dot to the underscore?  I think ES does not allow a dot but this is an hAlerting contect name, no?

lucasponce: Because Elasticsearch 2.x version doesn't accept dot as part of field name.
And when storing a full alert, it might fail with context with dots.
It is fixed in upper versions but 2.x is the target used for OS deployments, so that's motivation.

jshaughn:
Thanks, Lucas!

PR #318, opened 2017-04-05
Update hawkular-commons 0.9.1 -> 0.9.2

pgier:


lucasponce:
Travis looks good.
Is this one necessary for 1.6.x branch ?
(1.6.x versions would be used for metrics 0.26.x).

lucasponce:
I think yes, but please @pgier or @jshaughn just confirm and I can merge and move the commit to the 1.6.x branch.

jshaughn:
@pgier I think I must have missed the conversation about this.  I see that H Services is at 0.9.2 so I guess you are trying to consolidate around that?  As I understand the H Metrics release plan H Services will use the Metrics 0.26.x line until further notice.  And that means the 1.6.x line for H Alerting.

So this PR for H A master, and the other for H M master are OK but will actually just keep things up to date going forward.  What we really do need are PRs for 0.26.0 and 1.6.x branches for M and A, respectively, because that is what is going to be in H Services for the foreseeable future, Correct me if I am wrong @stefannegrea.

pgier:
@lucasponce @jshaughn Thanks! I submitted pr #319 for the 1.6.x branch.

PR #327, opened 2017-05-23
HWKALERTS-262 Fix notification issues when setting conditions

jshaughn:
A couple of related problems fixed here:
- remove deprecated java API add/remove/updateCondition
  - internal use caused some problems, just kill them
- avoid unnecessary cache updates in publishCacheManager
- add setAllConditions() to java API
  - avoids multiple notifications
  - ensures all dataids for a trigger (both modes) are correctly cached

@lucasponce please review. The problems here were subtle, they hit me only intermittently after I had made a change to the metrics alerter's notification listener and "refresh" code.

### DIFF
--- a/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/PublishCacheManager.java
+++ b/hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/PublishCacheManager.java
@@ -21,6 +21,7 @@ import static org.hawkular.alerts.api.services.DefinitionsEvent.Type.TRIGGER_REM
 
 import java.io.Serializable;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -77,7 +78,7 @@ public class PublishCacheManager {
     private Cache<TriggerKey, Set<String>> publishDataIdsCache;
 
     // It stores a list of triggerIds used per key (tenantId, dataId).
-    // This cache is used by CacheClient to check wich dataIds are published and forwarded from metrics.
+    // This cache is used by CacheClient to check which dataIds are published and forwarded from metrics.
     @Resource(lookup = "java:jboss/infinispan/cache/hawkular-alerts/publish")
     private Cache<CacheKey, Set<String>> publishCache;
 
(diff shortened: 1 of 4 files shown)


hawkular-alerts-engine/src/main/java/org/hawkular/alerts/engine/cache/PublishCacheManager.java
```diff
@@ -99,26 +100,35 @@ public void init() {
...
-                    if (e.getType().equals(TRIGGER_CONDITION_CHANGE)) {
-                        Set<String> newDataIds = e.getDataIds();
-                        publishDataIdsCache.put(triggerKey, newDataIds);
-                        addPublishCache(tenantId, triggerId, newDataIds);
-                    }
-                    publishDataIdsCache.endBatch(true);
-                    publishCache.endBatch(true);
-                });
+                events.stream()
+                        .sorted((e1, e2) -> e1.getType() == TRIGGER_REMOVE ? 1 : -1) // process removals last
```
lucasponce: I think I see where this is coming.
Unless I miss something I see risky to order the events and process the TRIGGER_REMOVE at last.
I have the feeling it could break things.

For example, in the code we might have:
[1] Delete trigger A
[2] Create new trigger A
[3] Update conditions on trigger A

In previous commit the events type were moved to a list to preserve the order of processing.
With the current code I guess that [1] can be executed at last and then break the order of notifications.

Another thing that it can impact, is that in the tests we don't count that a modification needs time in the engine to process the changes. Sometimes we do the [1][2][3] steps in a test plus [4] send data, and results are not expected because async processing didn't finish.

We can comment off line, perhaps if you have an example we can take a look, I can check if I am understanding the issue correctly.

jshaughn: Right, good catch.  I can revert the sort, it was intended just as an optimization.  I think the rest of the optimization is fine, although it too is not related to the core fix, so let me know if you think there is still an issue.

lucasponce: I think is fine as the order is preserved.
Let's see if tests are passing and I will merge it.

PR #329, opened 2017-05-30
Extract log, properties, rest-api and server into commons

lucasponce:
This one is pending on https://github.com/hawkular/hawkular-commons/pull/104
To run this branch it is needed to build locally a commons package with the pointed artifacts dependencies.


lucasponce:
So, it is expected that travis will fail until commons is consolidated.

PR #330, opened 2017-05-31
ActionPlugins and ActionDefinitions on Ispn backend

lucasponce:
@jshaughn, I have started to add support to both C* and Ispn backend.
In an iterative way I've started with easy ones.
Plan to continue tomorrow, but I wanted to consolidate this instead to send an uber-PR.


jshaughn:
I'll merge this in. I'm still getting familiar.  From what I can see we're limiting our use of Hibernate Search annotations to just those used for indexing.  I assume that it's by design that we're not using JPA annotations like @Entity, or the various relationship annotations like *To*, so that we're not bound to a JPA datastore.  Is that right?

lucasponce:
Correct. With Hibernate Search and Ispn user can control what to index in fine granularity.
So, no need to use JPA or similar, just a high level API to Lucene (backed engine for indexes).

PR #332, opened 2017-06-02
Ispn condition

jshaughn:
@lucasponce This isn't quite done, I need to add tests but also there is a test failure that points out an issue that maybe you are aware of but caught me a little by surprise.  When doing something like:

backend.put(pk(trigger), trigger);
trigger.setId("flubber");

The Trigger object in the cache is actually modified.  This maybe should not be surprising, but it is dangerous and tripped me up.  We should discuss whether we want this behavior or whether we want to actually store copies.  As is we need to be very careful.

Anyway, if you want to merge this go ahead, and we can just fix things as we move forward.

### DIFF
--- a/api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
+++ b/api/src/main/java/org/hawkular/alerts/api/services/DefinitionsService.java
@@ -369,66 +369,31 @@ public interface DefinitionsService {
      */
 
     /**
-     * A convenience method that adds a new Condition to the existing condition set for the specified
-     * Trigger and trigger mode.  The new condition will be assigned the highest conditionSetIndex for the
-     * updated conditionSet.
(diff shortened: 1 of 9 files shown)


lucasponce:
> backend.put(pk(trigger), trigger);
trigger.setId("flubber");

> The Trigger object in the cache is actually modified. This maybe should not be surprising, but it is dangerous and tripped me up. We should discuss whether we want this behavior or whether we want to actually store copies. As is we need to be very careful.

Yes, at the end we are managing a memory map, so object is stored with reference.
I have also thought about to store/return copies to be safe, but I wanted to progress more and see once we are familiar with the impl this is still a risk or not, having in mind that the user interacts with the REST endpoint, it is more a internal behaviour rather something exposed to end users.

Anyway, good point, I'm going to merge this PR, fix the problem and continue with the work.

PR #336, opened 2017-06-02
Condition support

jshaughn:
- remove deprecated getCondition()
- add test
- add indexing

WIP - This is not yet working, there seems to be issues with inheritance

Getting this to work is not going well. May need to get some ISPN or HS help internally... I hit different problems using ispn dsl or hs dsl.

### DIFF
--- a/api/src/main/java/org/hawkular/alerts/api/model/condition/AvailabilityCondition.java
+++ b/api/src/main/java/org/hawkular/alerts/api/model/condition/AvailabilityCondition.java
@@ -99,6 +99,14 @@ public class AvailabilityCondition extends Condition {
         this.operator = operator;
     }
 
+    public AvailabilityCondition(AvailabilityCondition condition) {
+        super(condition);
+
+        this.dataId = condition.getDataId();
+        this.operator = condition.getOperator();
+    }
+
+    @Override
     public String getDataId() {
         return dataId;
     }
@@ -161,7 +169,7 @@ public class AvailabilityCondition extends Condition {
 
     @Override
     public String toString() {
-        return "AvailabilityCondition [triggerId='" + triggerId + "', " +
+        return "AvailabilityCondition [tenantId='" + tenantId + "', triggerId='" + triggerId + "', " +
                 "triggerMode=" + triggerMode + ", " +
                 "dataId=" + (dataId == null ? null : '\'' + dataId + '\'') + ", " +
                 "operator=" + (operator == null ? null : '\'' + operator.toString() + '\'') + "]";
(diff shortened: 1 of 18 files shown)


lucasponce:
> WIP - This is not yet working, there seems to be issues with inheritance

> Getting this to work is not going well. May need to get some ISPN or HS help internally... I hit different problems using ispn dsl or hs dsl.

Polymorphic is supported in ISPN and HS but needs
- Add @Indexed annotation on parent and child classes (under same or different name, that's a performance consideration that can be reviewed later).
- Also, I think you are hitting the NPE that I found and I reported to ISPN

```
<lponce> anistor, I think I spot a potential bug here https://github.com/infinispan/infinispan/blob/9.0.0.Final/object-filter/src/main/java/org/infinispan/objectfilter/impl/util/ReflectionHelper.java#L261 if I am not wrong the getDeclaredMethod() only look into the class but not take into consideration parent classes, so I reach that in an example using the above discussed.
<anistor> lponce: so you do have a failing test :)
<anistor> lponce: share it pls
<lponce> sure, a second and I can push it
<lponce> anistor, https://github.com/lucasponce/hawkular-alerts/tree/ispn-backend, you can mvn clean install -DskipTests ; cd engine ;  mvn test -Dtest=IspnAlertsServiceImplTest
<lponce> anistor, it's easy to trace the call with a debugger
<anistor> lponce: thanks
```
So, workaround for this until it is merged is just to bring the getXXX method on the child class over is indexing (until we have a patch and we can bump version), I'm using that strategy in Alert/Event and queries work fine.

Next week I'm going to be some time off, so I'm planning to work on weekend on this, I suggest I can merge it and take a look to this.

lucasponce:
You are right.
When storing the child, parent is not well indexed.
I think we can re-use the workaround for the ActionPlugin.
If you look into the org.hawkular.alerts.engine.impl.model I put a ActionPlugin class which is just a private for the implementation, only used for storing in the cache and put the current indexes there.
I guess we can put a generic "Condition" class there and make the same strategy, put the indexes fields there, so internally.
Let me know what do you think, I won't merge this one, I will continue the work about Alert/Event and I suspect I will need same trick, I found a related issue but focus on fields from parent.
Anyway I guess that anistor from ISPN is aware of this with my chat.

lucasponce:
Or thinking in general, perhaps we can extend this and use the org.hawkular.alerts.engine.impl.model for all entities and apply the indexes there. I think it is a clear approach, so we can hide the indexes (which in theory is an implementation detail). And we can complete the discussion about clonning objects, as we now need to create the wrapper objects.
I think this can resolve questions commented before.
I am going to apply that to the existing getTriggers to see how it looks like.

jshaughn:
@lucasponce Updated to use the new pattern.  Removed DNM flag.

PR #337, opened 2017-06-05
Move Indexes to impl details

lucasponce:
As follow up of https://github.com/hawkular/hawkular-alerts/pull/336 I prepared this PR.
It moves the indexes annotation into the impl package, so api does not need to have visibility of them.
Also, continue with the discussion, we can store copies on the backend, two reasons: make idempotent the objects stored in the backend cache and simplify the indexes on polimorphics classes to avoid workarounds of ispn-objectfilter package.

@jshaughn I think this is in good shape to merge and it will simplify the work on the Conditions side.
From me I will continue with the queries for Alerts/Events.

### DIFF
--- a/api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
+++ b/api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
@@ -26,12 +26,6 @@ import java.util.UUID;
 import org.hawkular.alerts.api.model.Severity;
 import org.hawkular.alerts.api.model.data.Data;
 import org.hawkular.alerts.api.model.event.EventType;
-import org.hawkular.alerts.api.model.index.TagsBridge;
-import org.hibernate.search.annotations.Analyze;
-import org.hibernate.search.annotations.Field;
-import org.hibernate.search.annotations.FieldBridge;
-import org.hibernate.search.annotations.Indexed;
-import org.hibernate.search.annotations.Store;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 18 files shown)


api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
```diff
@@ -263,6 +253,40 @@ public Trigger(String tenantId, String id, String name, Map<String, String> cont
...
+        this.actions = new HashSet<>();
+        for (TriggerAction action : trigger.getActions()) {
+            this.actions.add(new TriggerAction(action));
+        }
+        this.autoDisable = trigger.isAutoDisable();
+        this.autoEnable = trigger.isAutoEnable();
+        this.autoResolve = trigger.isAutoResolve();
+        this.autoResolveAlerts = trigger.isAutoResolveAlerts();
+        this.autoResolveMatch = trigger.getAutoResolveMatch();
+        this.dataIdMap = new HashMap<>(trigger.getDataIdMap() != null ? trigger.getDataIdMap() : new HashMap<>());
```
jshaughn: Minor, I think you meant:
` this.dataIdMap = trigger.getDataIdMap() != null ? new HashMap<>(trigger.getDataIdMap()) : new HashMap<>());`

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
```diff
@@ -182,7 +184,8 @@ public Alert getAlert(String tenantId, String alertId, boolean thin) throws Exce
             log.debugf("getAlerts criteria: %s", criteria.toString());
         }
 
-        StringBuilder query = new StringBuilder("from org.hawkular.alerts.api.model.event.Alert where ");
+        StringBuilder query = new StringBuilder("from org.hawkular.alerts.engine.impl.ispn.model.IspnEvent where ");
+        query.append("eventType = 'ALERT' and ");
```
jshaughn: I'm not sure how Lucene does things and whether the order of criteria matters.  But to keep our approach of most restrictive criteria first we'd want to move this condition to the end, as the field has only 2 possible values.

PR #338, opened 2017-06-06
ISPN Dampening support

jshaughn:


engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnDefinitionsServiceImpl.java
```diff
@@ -419,43 +460,147 @@ public Dampening addGroupDampening(String tenantId, Dampening groupDampening) th
...
+            throw new IllegalArgumentException("Trigger [" + tenantId + "/" + triggerId
+                    + "] is a member trigger and must be managed via the group.");
+        }
+
+        return updateDampening(dampening);
+    }
+
+    private Dampening updateDampening(Dampening dampening) throws Exception {
+        try {
+            backend.put(pk(dampening), dampening);
```
lucasponce: perhaps here is an IspnDampening instead ?

jshaughn: Ha, good eyes.  Of course the bug was in the method for which I forgot to add a test.  Fixing...

lucasponce:
Thx ! merging...

PR #339, opened 2017-06-06
Finish TODO section of RemoveTrigger

jshaughn:
@lucasponce I added batch support for the backend cache config, for the remove.  If you think it's a bad idea we can talk about it.  I noticed I hadn't finished the TODO in removeTrigger after adding condition and dampening, so just wanted to finish that up.


lucasponce:
Batch support should be fine. In a future phase, when we compare, it is a point that can affect to performance as it forces sync communication and perhaps we don't want that in all operations.
But for now, it doesn't matter, so no problem to add it.

PR #340, opened 2017-06-08
Adapt tagQuery parser to Ispn syntax

lucasponce:
I am working in the tagQuery which is one of the difficult criterias.
Hopefully the backend supports all usecases and it is a change of design, just creating a proper query.
I am doing this in iterative steps, I want to consolidate this in .next.
@jshaughn can you merge ? thanks


jshaughn:
Looks good.

PR #341, opened 2017-06-08
ISPN Group Trigger Support

jshaughn:
@lucasponce migrated over the group trigger support.  added a new indexed trigger field to easily query member triggers of a group. Also ported over the trigger enablement services.

### DIFF
--- a/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnDefinitionsServiceImpl.java
+++ b/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnDefinitionsServiceImpl.java
@@ -53,6 +53,7 @@ import org.hawkular.alerts.api.model.condition.StringCondition;
 import org.hawkular.alerts.api.model.condition.ThresholdCondition;
 import org.hawkular.alerts.api.model.condition.ThresholdRangeCondition;
 import org.hawkular.alerts.api.model.dampening.Dampening;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnDefinitionsServiceImpl.java
```diff
@@ -489,7 +917,48 @@ public void removeDampening(String tenantId, String dampeningId) throws Exceptio
...
+
+            Collection<Trigger> memberTriggers = getMemberTriggers(tenantId, groupId, false);
+
+            for (Trigger member : memberTriggers) {
+                Collection<Dampening> dampenings = getTriggerDampenings(tenantId, member.getId(),
+                        groupDampening.getTriggerMode());
+                if (dampenings.isEmpty()) {
+                    continue;
+                }
+                removeDampening(dampenings.iterator().next());
```
lucasponce: I guess a comment for future maintainers "getTriggerDampenings() is filtered with TriggerMode, so only one dampening is expected"

jshaughn: It can return multiple if the 'triggerMode' param is null: From jdoc:
Parameters:
  triggerMode Return only dampenings for the given trigger mode. Return all if null.

PR #348, opened 2017-06-20
ISPN ActionsService support

jshaughn:
@lucasponce Please take a look.  I removed the unused executor from ispnalertsserviceimpl, we can put it back if you think we need it.  I also removed use of an executor in ispnactionsserviceimpl.  My understanding is the executor was needed only for c* concurrent writes.

Also, I wasted a lot of time trying to use the "in" operator in an ispn query.  I don't see that I was doing anything wrong but it seems not to work.  Instead I'm using an exploded "or" approach.

### DIFF
--- a/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnActionsServiceImpl.java
+++ b/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnActionsServiceImpl.java
@@ -16,21 +16,41 @@
  */
 package org.hawkular.alerts.engine.impl.ispn;
 
-import java.util.concurrent.ExecutorService;
+import static org.hawkular.alerts.engine.util.Utils.isEmpty;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.hawkular.alerts.api.model.action.Action;
+import org.hawkular.alerts.api.model.action.ActionDefinition;
 import org.hawkular.alerts.api.model.event.Event;
+import org.hawkular.alerts.api.model.paging.ActionComparator;
+import org.hawkular.alerts.api.model.paging.ActionComparator.Field;
+import org.hawkular.alerts.api.model.paging.Order;
 import org.hawkular.alerts.api.model.paging.Page;
 import org.hawkular.alerts.api.model.paging.Pager;
 import org.hawkular.alerts.api.model.trigger.Trigger;
+import org.hawkular.alerts.api.model.trigger.TriggerAction;
 import org.hawkular.alerts.api.services.ActionListener;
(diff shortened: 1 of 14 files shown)


lucasponce:
First comment is the executor is needed.
It helps to not block the engine when executing actions.
I will take a close look this morning with more comments.

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnActionsServiceImpl.java
```diff
@@ -61,32 +95,312 @@ public void setActionsCacheManager(ActionsCacheManager actionsCacheManager) {
...
+                query.append(inClause("result", results));
+            }
+        }
+
+        List<IspnAction> ispnActions = queryFactory.create(query.toString()).list();
+        return prepareActionsPage(ispnActions.stream().map(a -> a.getAction()).collect(Collectors.toList()), pager);
+    }
+
+    // An exploded "in" clause because the actual one seems not to work
+    private String inClause(String field, Set<String> values) {
```
lucasponce: I use the same pattern in alerts criteria.
I think you need an iterator here or you will have a " and ( or field = 'value' or field = 'value')" and this problably will throw a syntax error.
I used a 
```
Iterator<String> iter = values.iterator();
while (iter.hasNext()) {
      String value = iter.next();
      sb.append("field = '").append(value).append("'");
      if (iter.hasNext()) {
        sb.append(" or ");
      }
}
```

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnActionsServiceImpl.java
```diff
@@ -61,32 +95,312 @@ public void setActionsCacheManager(ActionsCacheManager actionsCacheManager) {
         this.actionsCacheManager = actionsCacheManager;
     }
 
-    public void setExecutor(ExecutorService executor) {
```
lucasponce: The executor has the meaning to not block the engine thread while writting in the backend or in actions, when processing actions.
So I would leave it.
Once that we start the round of itest on Ispn backend we can decide if we refactor it but I wouldn't introduce this change now.

lucasponce:
Merging, any comments can be fixed in upcoming PRs.

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnActionsServiceImpl.java
jshaughn: > I think you need an iterator here

No, this is OK as is.  Your approach is good, too.

jshaughn:
> First comment is the executor is needed

OK, we can add some executor use back in as needed.

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnActionsServiceImpl.java
lucasponce: true, the separator was initially empty. I didn't see it :-).

PR #351, opened 2017-06-22
Complete Events logic for Ispn backend

lucasponce:
PR to complete the Events part.
I am working putting all pieces together and enabling the itest for ispn backend.


engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
```diff
@@ -488,17 +543,110 @@ public Alert getAlert(String tenantId, String alertId, boolean thin) throws Exce
...
+            if (criteria.hasCTimeCriteria()) {
+                query.append("and (");
+                if (criteria.getStartTime() != null) {
+                    query.append("ctime >= ").append(criteria.getStartTime()).append(" ");
+                }
+                if (criteria.getEndTime() != null) {
+                    if (criteria.getStartTime() != null) {
+                        query.append("and ");
+                    }
+                    query.append("ctime <= ").append(criteria.getEndTime()).append(" ");
```
jshaughn: Just an FYI, the "between" operator seems to work and could be used when having start and end criteria.  I don't know if it has any perf impact.

jshaughn: It may be useful at some point to test whether the order of operators has any perf impact.  For example, it may make sense to apply time constraints before anything else.  Hopefully there is a decent optimizer and ordering is unnecessary, but it may be worth checking.

PR #353, opened 2017-06-26
Fix Ispn and Cass PersistenceTests on itest profile

lucasponce:


lucasponce:
With this PR, both Ispn and Cass PersistenceTest will be run by itest profile.

jshaughn:
LGTM

PR #354, opened 2017-06-26
Prometheus external alerter

jshaughn:
This is an initial alerter that can set up an ExternalCondition to execute
a Promql query and feed the results back as events for matching the
condition.
Some things to think about given this POC:
- Should we use tags or context for the prometheus labels?
- How well will this really work given that prometheus could return
  many metrics for the single expression?
  - should we think about setting source? Data-driven groups here?

Next: This can also work in conjunction with P8A Alerts.  Because P8S
generates a special "Alerts" time series for the alerts it generates, it
may be a good approach to actually query the Alerts TS as opposed to the
metric data directly.  I'm going to write a test to try this out...


hawkular-alerters/hawkular-prometheus-alerter/hawkular-prometheus-alerter-lib/src/main/java/org/hawkular/alerter/prometheus/BaseHttpClientGenerator.java
```diff
@@ -0,0 +1,370 @@
...
+
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
+
+import org.jboss.logging.Logger;
+
+import okhttp3.OkHttpClient;
```
lucasponce: Minor, is this one is for just http communication, we should try to unify them for the .next branch, as there are several libraries for the same use (I am thinking http-builder).

lucasponce:
> Should we use tags or context for the prometheus labels?

First thought is to use context, as it is a separated domain (p8s vs hawkular) but if we want to enable searches in hawkular for it, then we should use tags.
In general, context is good to place info that helps to explain the trigger.

> How well will this really work given that prometheus could return
many metrics for the single expression?
should we think about setting source? Data-driven groups here?

I don't think this is a problem or it should be, I don't think we should bring that granularity yet. In P8S there is a situation detected and it is brought to Hawkular to combine them with other situations and correlate.
I don't see data-driven groups yet, perhaps I can need a detailed example to understand it.

PR looks good to me, there is a travis issue that I think we should address (just to avoid noise in the CI), but once fixed I can merge it.

jshaughn:
This is now green, merging...

PR #355, opened 2017-06-27
Prom webhook

jshaughn:
This adds a REST endpoint to the Prometheus alerter making it a notification receiver.  It converts and persists alert notifications as events.


lucasponce:
LGTM

PR #357, opened 2017-06-29
Distribution for Alerting standalone

lucasponce:
This PR creates a standalone distribution with a simple layout ready to use and a .zip.
It depends on https://github.com/hawkular/hawkular-commons/pull/110 so, it is using SNAPSHOTS and it will fail in CI, as soon as commons is updated I will add a proper srcdeps and update it.


jshaughn:
You can now update to SRC-revision-bf3e241a07008978f949430b553ee5dba1d5056a

lucasponce:
Side comment.
I've executed the heavy PerfCrudITest with interesting results:
image
In a single 64Mb process it passes the tests without errors.
It is visible that it reaches the GC preassure and performance degradates (central curve of the graph), but passing this test with minimal configuration of memory is a good sign.

jshaughn:
Yes, this is good.  Also, it ran in less than 20 minutes, which iirc is pretty good.  Merging...

PR #358, opened 2017-06-29
Static content handler for UI

lucasponce:
Added a /ui endpoint for static content where we can place all UI AngularJS resources.


jshaughn:
We have a UI! :)

PR #359, opened 2017-06-30
Port Alerters architecture

lucasponce:
Consolidate Alerters around an API to make them easy to add.
Reusing same scheme for actions architecture which simplify the plugability.
Also porting elasticsearch and prometheus alerters into .next.


jshaughn:
LGTM

PR #361, opened 2017-07-03
Backbone for UI App

lucasponce:
This PR contains:
- Extract UI as a separate project.
- Setup all automation for UI based on npm, bower and grunt tool.
- Add UI build into pom.xml.
- Basic mockup based on AngularJS and PatternFly.


lucasponce:
The mock up has only intention to enable all PatternFly + AngularJS technology to speedup the learning curve and focus on features instead to spend time defining layouts and conventions.

image

image

image

ui/README.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+```
+    http://<host>:<port>/hawkular/alerts/ui
+```
+
+== Working on UI
+
+To develop in the UI Application you will need NodeJS tools.
+
+A standard strategy could be to work locally on your
```
jmazzitelli: You should make it clear that the developer should "cd" to the directory (.../src/main/ui) before running the npm/bower commands below.

ui/README.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+directory and prepare the dependencies you need to run all the UI elements
+
+```
+    npm install
+    sudo npm install -g bower
+    bower install
+    sudo npm install -g grunt-cli
+    sudo npm install -g compass
+    gem install sass
```
jmazzitelli: "gem" -- this requires Ruby / RVM to be installed I assume?

ui/README.adoc
```diff
@@ -0,0 +1,79 @@
...
+    gem install compass
+```
+
+These steps will create some folders that will be part of the .gitignore as they can't be pushed into the repository
```
jmazzitelli: After these instructions, I have the following that is not in .gitignore - git wants me to commit it. I think you will want to put this in .gitignore, too, correct?

`ui/src/main/ui/package-lock.json`

ui/README.adoc
lucasponce: +1

ui/README.adoc
lucasponce: Yes. I don't pretend to be comprehensive here, basically we are pointing to AngularJS + PatternFly and re-using the demo app, so it is possible that some additional note should be added in following iterations.

ui/README.adoc
lucasponce: Yes, it is weird, I didn't have it on my machine.
But yes, the lock shouldn't be pushed, so +1 to add it into the .gitignore.

lucasponce:
Travis worked fine, did you get able to have a local environment or is there any important step missed ?

jmazzitelli:
I have it running and I can see it in my browser when `grunt server` is run.

lucasponce:
Cool, thanks @jmazzitelli, going to add your comments and merge it.

PR #362, opened 2017-07-03
start in foreground logging in console

jmazzitelli:
be able to start alerting distro in foreground so logs show up in the console window


jmazzitelli:
This way you can run "hawkular.sh console" and not have to immediately run "tail -f logs/*" so you can see the output of the server. The logs shows up in the console. Additionally, this runs in foreground so you can "control-c" to kill the server rather than having to run the "hawkular.sh stop" to kill the background process.

lucasponce:
Note that we don't have yet captured the control-c signal.
So, perhaps we would need a PR for this on the server project that is hosted in commons:

https://github.com/hawkular/hawkular-commons/blob/master/vertx/server/src/main/java/org/hawkular/HawkularServer.java

Main reason this was not done before is I was looking to capture the kill SIGTERM <PID> signal from the JVM to invoke the stop, but there is not guarantee that is invoked from the JVM (It is from a control-c but not from an external kill SIGTERM as in some cases it depends of the platform).

So, potential side effect of killing the process is that some memory state couldn't be consolidated on disk (basically the indexes) and we should start the server with a flag to reindex the backend.
(This is a cornercase but I document it here to not forget).

PR #363, opened 2017-07-05
change grunt server IP

jmazzitelli:
set default to 0.0.0.0 so a developer can run the server on a dev machine while accessing the server from another client machine.


lucasponce:
LGTM
There are some options for local development vs server that we can prepare in following iterations.

PR #364, opened 2017-07-06
Hello World UI

lucasponce:
- Basic dashboard
- Support for listing triggers and actions


lucasponce:
The aim of this PR is to provide a basic structure for the UI application.
Also, one of the goals is to show features an functionalies in a visual way, instead to use only REST API.
With this one, we have started to show what a Hawkular Alerting Hello World does:

image

Now, definitions, actions and alerts generated can be visualized from the UI.

/cc @jmazzitelli @jotak

jshaughn:
Awesome Start!

jotak:
Nice work!
I wonder if we should align framework used (ie. AngularJs vs Angular) between here and metrics; Probably that, unless we want to merge them at some point, they can go independently

lucasponce:
@jotak Hi Jotak, I don't see any problem to confluence at some point in the time.
For the moment the idea is to provide a simple UI for "selling" the features, also, in a future metrics + alerting, most of the features can be added (functional speaking) into the unify solution.
The idea to choose patternfly + angularJS is just to not spend too much time in technologies but in something easy to show.
I would like to progress a little bit more and when we have something solid we can plan as you point, that probably we can import modules and offer them in a single solution (and perhaps we can deal with the technical details of unification of frameworks).

PR #365, opened 2017-07-06
be able to bind to different ip/port

jmazzitelli:
1. Get hawkular.sh to pass the correct sysprops so you can properly set the bind address and port

2. Change dashboard-service.js so you need only uncomment one line to change the IP (rather than change three lines total (comment 1, uncomment 2 and change one of them).


jshaughn:
LGTM

dist/src/main/resources/hawkular.sh
```diff
@@ -19,6 +19,8 @@ set_java_opts() {
     JAVA_OPTS="$JAVA_OPTS -Xmx64m -Xms64m"
     JAVA_OPTS="$JAVA_OPTS -Dhawkular.data=${HWK_DATA}"
     JAVA_OPTS="$JAVA_OPTS -Dhawkular.configuration=${HWK_CONFIG}"
+    JAVA_OPTS="$JAVA_OPTS -Dhawkular.bind-address=$(hwk_prop 'hawkular-alerts.bind-address')"
```
lucasponce: Ah, I brought some technical debt from the 1.x version.
In 2.x perhaps just changing hawkular.bind-address and hawkular.port in the ${HWK_CONFIG} is enough.
I forgot to change it.
Thanks for pointing out.

ui/src/main/ui/src/dashboard/dashboard-service.js
```diff
@@ -2,11 +2,12 @@ angular.module('hwk.dashboardModule').service('hwk.dashboardService', ['$resourc
   function ($resource) {
     'use strict';
 
+    var host = '';
```
lucasponce: This also would need to be changed on triggers-service.js and actions-service.js as well.
As this is for development mode, I was thinking in some global parameter passed via 'grunt server' and automatically detected.
I haven't played yet with it.
Feel free to propose something in that direction.

PR #366, opened 2017-07-06
POC : Make Standard Triggers natively work like DataDriven Triggers

jshaughn:
Marking as Do Not Merge and review is not yet necessary.  I just want travis to run and to keep in sync with the .next branch.  After other .next work is stable I'd like to talk about this work.  I spent a few days investigating this somewhat major change to trigger semantics.  In short it basically merges the data-driven group trigger feature into standard triggers.  Put another way, it lets a single trigger handle data/events from N sources.  The big difference is that it does it without using a backing group trigger.   It's not totally seamless, there are some behavioral differences, but I think the change is worth considering.


lucasponce:
ok, let's open this topic once the .next is in a good shape, as this could be a long review and discussion session.

jshaughn:
Here is a diagram of the approach:
native data-driven triggers

jshaughn:
@lucasponce This can wait until I get back, but just in case you want to look while I'm away, I added a diagram and rebased on next.

jshaughn:
Closing this and re-opening against master (because the next branch has been migrated to master).

PR #367, opened 2017-07-06
getEvents() should (I think) only return Events

jshaughn:


lucasponce:
This makes a non compatible change with 1.x.
getEvents() should return Events (plain Events and Alerts).
So, I don't think this change is good.
User could use a filter to get only one type of Events, but this breaks previous functionality.

lucasponce:
I start my shift this morning.
I guess I know where this is coming, basically in the UI we show redundant info in the Events and Open Alerts.
Instead to change the model (which is fundamental) I'm going to add an eventType criteria and apply it into the UI.
I am also porting a couple of examples from 1.x to 2.x and also thought on this (but in terms of visualization only).
I will send a PR for that.

jshaughn:
I'm not sure you are right about 1.x behavior, I will have to check.  But in cassandra events and alerts are held in separate tables, so I'm fairly sure getEvents did not return alerts.  Adding EventType criteria is OK, but without using it we may have a behavioral change.  Moreover, if we support two backends we'll see different default behavior.

You are right about the timeline, I don't think we should duplicate alerts on the event line.

lucasponce:
In C* implementation, when an alert is generated, it is also stored in the events table.
In the ISPN implementation there is not a duplication, so, both Alerts and Events are stored under the same IspnEvent. No need to store two objects as it happened in C*.
So, forcing to only fetch eventType = 'EVENT' misses the ALERT objects (which are also EVENTS).
That reason I added the eventType criteria in the following PR and I guess that is the behaviour you had in mind.
Let me know if I am wrong and perhaps you wanted to get a different result.

jshaughn:
OK, so that is the piece of information I did not remember, it's not that getEvents returned alerts, it's that alerts were also stored as events. So we were both right :)  I'll close this and yes, we do need the eventType filter, and we need to apply it to the UI for the events line.

PR #368, opened 2017-07-06
set cors props

jmazzitelli:


lucasponce:
LGTM
Note that all "hawkular.*" are read from the HawkularProperties framework, not directly from the System.getProperty().
This allows us to config them directly under config/hawkular.properties which reduce the properties to handle directly in the hawkular.sh script.

PR #369, opened 2017-07-07
AutoResolve example

lucasponce:
- Port auto-resolve example.
- Add eventType criteria to getEvents().
- Update UI to add eventType criteria and to not duplicate the ALERT and EVENT in the timeline.

@jshaughn I think adding the eventType criteria makes more flexible and less breaking the https://github.com/hawkular/hawkular-alerts/pull/367, so, now UI differenciates between ALERT/EVENT and API is still working as it was designed.


jshaughn:
Until I confirm the cassandra/1.x behavior I don't think we should merge this.  I'll let you know...

jshaughn:
OK, as discussed in the other PR, because with cassandra/1.x we stored alerts in both the alerts and events table the query behavior is the same.   To remove the redundancy in the UI we need to apply the filter to the events timeline to filter out alerts (we may have been able to do this with 'category' but eventType is more explicit).  merging this along with the other changes in this PR.

PR #372, opened 2017-07-07
fix the "console" option 

jmazzitelli:
fix the "console" option so it doesn't screw up the indexes of ispn on shutdown.
This now puts the server in background just like start_alerting, but it also immediately tails
the log file for you so it shows in the console, plus it adds a trap so when you control-c
the stop_alerting is called and the server is shutdown normally.


lucasponce:
LGTM, merging as travis doesn't test this one

PR #375, opened 2017-07-08
[WIP] Perfomance testing

gbaufake:
Hey Guys.

This is just a POC on performance testing on Hawkular Alerts. 

I am using [locustio](http://locust.io/) to generate the load  and it is pretty simple to configure with desired load.

The first "example" try to emulate the calls the are made from ManageIQ and this can used to simulate "one manage iq" or more than one.

Please give suggestions/feedback/critics on that PR.
Best Regards,
Guilherme Baufaker


lucasponce:
@gbaufake 
I would prefer if we align with Hawkular Metrics in tooling and design.
The goal should be to have a similar hook as Hawkular Metrics has and in each PR a CI job can be executed with the desired performance tests.
Also, it should give analysis about potential regressions and an explanation of the scenario.
About tooling and design, I suggest to contact with @FilipB to share their work and align in tooling.

The semantic of the test should take in place the semantics of the operations.
Create definitions is a expensive task and it does not expect to have an excessive load on it, the critical scenarios should be on detection.

Also, performance tests should have context and a well defined environment.
This means executing the load testing tool and the server in the same environment will introduce noise and probably you won't be able to get real info about where a bottleneck exists and why.

So, for 1.x we have a Cassandra process and a Wildfly process, each one should have its JVM limits and we need to have all this info referenced to know how react the software under different parameters, to know where are the limits, and whether a limit reach is a problem or not.

PR #376, opened 2017-07-10
HWKALERTS-270 TagQuery does not support empty spaces on regexp values

lucasponce:
This is a fix to support spaces inside tagQuery regexp.
It was spotted by @josejulio for 1.6.x version.
It is first fixed in master but I can port this into the 1.6.x branch and prepare a release there if necessary.
There is a workaround (not using spaces), to I let @josejulio decide which approach works better for his needs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
-sudo: false
+sudo: required
+dist: trusty
+
 language: java
 jdk:
 - oraclejdk8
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,4 +1,6 @@
-sudo: false
+sudo: required
+dist: trusty
```
jshaughn: Interesting, I guess this provides a more independent env for each run.

PR #378, opened 2017-07-10
CORS/Commons Update

jshaughn:
- update commons src-dep to use new cors props and * origin,
  making dist and grunt both usable by default.


lucasponce:
LGTM

PR #380, opened 2017-07-10
fix grunt clean so it really cleans

jmazzitelli:
There are times you have to clean out EVERYTHING that was generated to get back to a normal running state.

This allows "grunt clean" to clean everything. You then "bower install && npm install && grunt" to re-generate and build everything again.
I found this was the only way to get the UI to work under some condition I found myself in.


lucasponce:
Bower folder should not affect.
Also npm exec is first than bower.

jshaughn:
It's unclear to me as to the state of this PR.  Mazz, are you OK with just adding a new flavor of clean?

jmazzitelli:
We can adjust this PR to add, say, a "cleanall" or "purge" target or something like that. We can leave "clean" the way it was, but add a "cleanall" (or "purge" or whatever) target to do what clean does and then also clean these other things.

jmazzitelli:
@lucasponce @jshaughn I changed this PR. "clean" won't touch the install stuff so you won't be surprised when you clean. However, there is a new "purge" task - when you "grunt purge" it will remove all generated and installed files. It won't touch anything committed to git. So you can purge to clean the entire location of all generated and installed files.

jmazzitelli:
Here's what the output will look like when you run "grunt purge"

```
$ grunt purge
Running "clean:dist" (clean) task
Cleaning dist...OK

Running "clean:source" (clean) task
Cleaning src/templates...OK

Running "clean:server" (clean) task

Running "clean:sass" (clean) task
Cleaning .sass-cache...OK

Running "purgeinstall" task
Removing bower_components
Removing node_modules
Removing package-lock.json
You now must run "npm install && bower install" before attempting to re-build.
```

lucasponce:
thx @jmazzitelli 
LGTM

PR #381, opened 2017-07-10
add donut graph

jmazzitelli:
Here's my first attempt at adding a donut chart. Through the donut, you can see the total number of alerts that are required to be attended to (this means both open and acknowledge, but not resolved) where the open alerts is the red arc of the circle and the acknowledged is the orange arc (red meaning, very important, orange meaning important but not as important because someone already acknowledged it).

I did not add "resolved" alerts to the donut because that number is going to be constantly growing and so will eventually trend to becoming the full donut. If resolved alerts get deleted, we could add it as the third arc of the donut, but AFAIK, resolved alerts always remain and thus will eventually trend to take over the entire donut.

Here's an image:

donut


lucasponce:
The donut graph is great, showing the total and the percentages with open and resolved is good.
I think that perhaps it is good also to maintain the cards with the absolute numbers for open and resolved.
So, from those cards we can add a link to a future "alerts" new section where we can list all alerts and perform close filter searchs (where user can also add tagQuery to refine things).
@jmazzitelli would you like to take this task ? Thx !

PR #382, opened 2017-07-11
Port GroupTrigger examples

lucasponce:
Add basic pagination on Triggers UI

Added pagination to show large number of triggers on the UI.
Pretty basic yet, but the goal is to browse the GroupTriggers example without need of external REST query.


lucasponce:
Ah, left 'my-organization' tenant in the UI pre-selected, as most of the examples are using this one.
I guess during the development phase it will be useful to not type the tenant.

jshaughn:
Thank you for leaving in the tenant.  I have no desire to type that long tenant even one more time...

PR #383, opened 2017-07-11
Initial list of action history for selected timeline items.

jshaughn:


jshaughn:
screen

lucasponce:
If you use the example/autoresolve-process you can test it with a resolved alert.
Resolved Alert brings a new column on the Dashboard details and perhaps you can tweak in the actions are looking as you expect.
A minor one, in the Triggers/Actions I also used icons to represent the action plugin, visually is also interesting.
Perhaps we can bring the same icon of those pages here, too.
(I have also used icons for email and webhooks plugins).

jshaughn:
With latest updates:

screen

lucasponce:
LGTM
Perhaps the icon can be aligned on the actionId (like the user icon with the name).

jshaughn:
> Perhaps the icon can be aligned on the actionId (like the user icon with the name).

I spent about an hour trying to get that to work nicely but failed.

PR #384, opened 2017-07-11
Create a global app config object - for now store base URL information.

jmazzitelli:
This means we no longer have to manually edit services js files if using
grunt or if the server is bound to some other IP/port that the grunt server is not.


jshaughn:
This seems good and helpful. I will merge.

PR #386, opened 2017-07-12
fix hello-world so it can write to any server, not just localhost:8080

jmazzitelli:
if the server is bound to some other non-localhost, non-8080 port, be able to tell examples where to send the data.

The default remains the same, so nothing changes if you just "mvn validate" but now you can do this: "mvn validate -Dhost=192.168.1.100:8181" and it will work.

You can also set the tenant via -Dtenant.


lucasponce:
Ah, probably you may want to change it in all the examples, as they follow the same pattern.

jshaughn:
If you do this example you'll need to do all of the examples because people will expect it.

jshaughn:
Ha, Lucas beat me by a few milliseconds

jmazzitelli:
all examples are updated.

jshaughn:
LGTM, merging as groovy files are outside the build...

PR #388, opened 2017-07-13
Triggers work

jshaughn:
- Add toolbar with tag filter and kebab
  - note: the kebab will offer a "New" option, and maybe "Update"
- Make trigger entry more visually appealing (at least to me, feedback welcome)


jshaughn:
Here is a look at the filter and proposed layout for the trigger entry:

screen

PR #389, opened 2017-07-13
Be able to start grunt server without starting up the browser window.

jmazzitelli:
The default behavior is the same as always (browser starts up with "grunt server")
but if you pass in --nobrowser, it won't start the browser.


jshaughn:
Lovely.

PR #390, opened 2017-07-14
Add default values for match/mode when Trigger is loaded from Ispn

lucasponce:
This fixes a bad bug on the engine.
When server is shutdown and restarted, transient fields are not serialized (which is intended), but in the Trigger(Trigger) constructor some missing default values where carried over the engine, throwing a non expected NPE.

### DIFF
--- a/api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
+++ b/api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
@@ -283,8 +283,8 @@ public class Trigger implements Serializable {
         this.source = trigger.getSource();
         this.severity = trigger.getSeverity();
 
-        this.match = trigger.getMatch();
-        this.mode = trigger.getMode();
+        this.mode = trigger.getMode() != null ? trigger.getMode() : Mode.FIRING;
+        this.match = trigger.getMode() == Mode.FIRING ?  trigger.getFiringMatch() : trigger.getAutoResolveMatch();
     }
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
```diff
@@ -283,8 +283,8 @@ public Trigger(Trigger trigger) {
         this.source = trigger.getSource();
         this.severity = trigger.getSeverity();
 
-        this.match = trigger.getMatch();
-        this.mode = trigger.getMode();
+        this.match = trigger.getMatch() != null ? trigger.getMatch() : Match.ALL;
+        this.mode = trigger.getMode() != null ? trigger.getMode() : Mode.FIRING;
```
jshaughn: I think this should be:
```
this.mode = trigger.getMode() != null ? trigger.getMode() : Mode.FIRING
this.match = trigger.getMode() == Mode.FIRING ?  trigger.getFiringMatch() : trigger.getAutoResolveMatch();

```

lucasponce: Probably it won't hit that logic, as when this constructor is invoked it should initialize the this.mode.
(As it happens in the other constructors).
But I think the suggestion is good, so I will update it and rebase the other PR.

lucasponce:
Is this fine to merge now ?

jshaughn:
LGTM, thanks.

PR #391, opened 2017-07-14
Unify http client on prometheus/elasticsearch

lucasponce:
This PR brings several things:
- Fix a bug in alerters, a missing initialization is needed to load triggers that are previously in the backend.
- Unifies the HTTP libraries used, let's remember that .jar are not cheap, and every new archive is loaded and consumes footprint, so doing an effort to reuse same libraries is good. Elasticsearch client brings as third party dependencies the apache httpcomponents so, having the okhttp is redundant and we can unify here (and reduce also code). I wish we could do the same also in logging but each third party tool uses their own frontend, hopefully we have unified the implementation.
- I have tested and validated the Prometheus alerter locally, all existing tests work.
- Minor one, it was brought before, but as we can control now the handlers, I changed the '/hawkular/__alerter_prometheus/notification' into a more clear '/hawkular/alerter/prometheus/notification' and additioanl alerters can share the '/hawkular/alerter' prefix for these use cases.

I'm going to write an example for better out-of-the-box experience but as I am not sure if I can do it in time I split this PR, and example will come in a following one.


api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
```diff
@@ -283,8 +283,8 @@ public Trigger(Trigger trigger) {
         this.source = trigger.getSource();
         this.severity = trigger.getSeverity();
 
-        this.match = trigger.getMatch();
-        this.mode = trigger.getMode();
+        this.match = trigger.getMatch() != null ? trigger.getMatch() : Match.ALL;
+        this.mode = trigger.getMode() != null ? trigger.getMode() : Mode.FIRING;
```
jshaughn: I think this needs to be updated as in the other PR's comment.

dist/src/main/resources/config/hawkular.properties
```diff
@@ -29,3 +29,4 @@ hawkular-alerts.engine-delay=1000
 hawkular-alerts.engine-period=2000
 hawkular-alerts.base-url=/hawkular/alerts
 #hawkular-alerts.backend-reindex=true
+#hawkular-alerts.prometheus-alerter=true
```
jshaughn: Do you think perhaps the alerter should be enabled by default.  I think at the moment the elastic alerter is enabled by default but the p8s alerter is not.  Maybe that is confusing.

alerters/alerters-plugins/alerters-prometheus/src/test/resources/alertmanager.yml
```diff
@@ -0,0 +1,17 @@
...
+  receiver: "hawkular-alerts"
+  group_by: ['alertname', 'tenant']
+  group_wait:      10s
+  group_interval:  1m
+  repeat_interval: 2m
+
+receivers:
+- name: "hawkular-alerts"
+  webhook_configs:
+  - url: 'http://localhost:8080/hawkular/alerter/prometheus/notification'
```
jshaughn: I don't see jboss-web.xml path change in this PR, is it missing?  I almost made the same change when I wrote this orginally, so I am +1 on the new path.

api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
lucasponce: +1 I need to rebase this.

dist/src/main/resources/config/hawkular.properties
lucasponce: Yes, I think we can set up enabled by default.
Also I think we can collect all configuration properties we have spreaded over several places and document them commented in the hawkular.properties file, now we have a nice one to do it.

alerters/alerters-plugins/alerters-prometheus/src/test/resources/alertmanager.yml
lucasponce: We are in our custom server based on Vert.X. You won't see any *web.xml descriptor anymore.
The Prometheus Alerter has an "App" which basically defines a baseUrl where we can put our domains and endpoints in a structured way.

jshaughn:
Duh! Stupid me...

lucasponce:
I will rebase this with #390

jshaughn:
Great!

PR #393, opened 2017-07-14
show alert severity in a pie chart

jmazzitelli:
severity-pie-chart


jmazzitelli:
At this point there is no "onclick" functionality on these pie/donut graphs.

The thinking is we add some onclick functions so when yoy, say, click on  "Open" alerts in the donut or in the "Critical" slice in the pie, your browser goes to a page that lists those alerts in some kind of a table view.

lucasponce:
I think we need also the open and acknowledged cards back.
For me it is missing. 
First donut shows total, with % (ok we can change to absolute values), but if we add resolved and events in cards, open and acknolwedged needs to be there. IMO.

The second donut with the criterias is great.

And yes, clicking on any link you browse to a new page with the specific filter applied, that would be great.

jmazzitelli:
After doing this severity chart, I'm starting to think i want to change the donut chart to a stacked vertical bar chart - two bars, one open, one ack. each bar with severity stacked (so one bar is the "open" alerts - and each stacked portion of that bar shows which are critical, which are high, medium, and low)

This would be in effect your "cards" that you want to bring back.

We can put the cards back underneath and have all four on the bottom. It would be duplicating the data, but I don't care. Having a row of all four cards left to right below the graphs.

jmazzitelli:
So what I suggest we do is merge this PR to get it into the next branch.

I will then put back the cards the way they were before - all four, in one row (not stacked on top of one another) so we'd have a row of graphs, and under that, a row of cards and under that, the timeline.

That will be done in another PR.

lucasponce:
+1

PR #394, opened 2017-07-14
Add an Prometheus Alerter example

lucasponce:
Add an example for the Prometheus Alerter.
It would be nice to have also a section in the docs like the Elasticsearch one.
An example of integration between P8S and ES would be great.


jshaughn:
Excellent!

PR #395, opened 2017-07-14
dashboard with four cards and a stacked bar graph

jmazzitelli:
stacked-bar

If you hover over a bar in the stacked bar graph, you'll get the numbers based on severity for that bar (open or acknowledged):

stacked-bar-2


jmazzitelli:
@lucasponce @jshaughn see what you think

jmazzitelli:
The previous commit moves the cards to reduce whitespace on screen. The cards rearrange nicely when you narrow the width of the browser screen. See images:

With a narrow browser window:

cards1

With a wide browser window:

cards2

jshaughn:
Good!

PR #396, opened 2017-07-14
Trigger Work

jshaughn:
- Add new trigger action via toolbar kebab
- Add enable/disable/delete trigger actions


ui/src/main/ui/src/triggers/triggers-service.js
```diff
@@ -33,5 +31,33 @@ angular.module('hwk.triggersModule').service('hwk.triggersService', ['$resource'
         }
       });
     };
+
+    this.NewTrigger = function (tenantId) {
```
mtho11: This is just a nitpick but why the capital letters on method names in services? When elsewhere it seems like lowercase (i.e. `$scope.newTrigger`)

jshaughn:
> Just a general comment - Why no typescript or ES6? ES5 is sooo yesterday (non-Radical technology).

@mtho11 ...because I don't even know what your comment means :)  It may be nearsighted but I think the reason is simply that we want to get something working as fast as possible and this is the least blocking mechanism for us , given our limited abilities in the UI area (and that doesn't completely go against an acceptable UI tech, afaik).  Lucas was able to jump start things using this tech and mazz and I are just trying to do anything useful we can as we try and get familiar.  Remember that our UI experience is limited to SmartGWT :)

I certainly have nothing against modernizing what we have but not currently at the expense of slowing down our ability to get out an initial UI.   PR's accepted!

ui/src/main/ui/src/triggers/triggers-service.js
jshaughn: No good reason, I guess.  Just following what was there (yes, I'm a lemming).

mtho11:
@jshaughn fair enough :) We'll worry about modernization later

jshaughn:
> We'll worry about modernization later

@mtho11 Thanks for keeping tabs on what we are doing, if we are going completely off the rails please stop us, we don't want something that completely goes against the grain and would need to be replaced in its entirety.

lucasponce:
Yes. The idea is to focus on features first as our efforts are limited.
Once the basic UI is in a good state we can make an effort to integrate/modernize.

PR #400, opened 2017-07-20
FullTriggerModal Work

jshaughn:
- Triggers modal now supports Create, View/ReadOnly and Edit options.
- Fix issue with 'x' button cancel


jmazzitelli:
travis is still red, but its due to travis internal problems

```
[0K$ if [ "${cachedBytes}" -gt "${availBytes}" ] ; then echo "Cleaning the cached dirs (${cachedBytes} Bytes) because their size exceeds the free space (${availBytes} Bytes) left on the current partition" rm -Rf "$HOME/.m2" fi /home/travis/.travis/job_stages: eval: line 58: syntax error: unexpected end of file
```

PR #401, opened 2017-07-21
missing semis might be the cause of the travis errors.

jmazzitelli:


jmazzitelli:
this still has the odd failure, but we need this PR. I think without those semis we get those other "EOF" errors.

PR #405, opened 2017-07-21
stub out alerts page and link on left menu

jmazzitelli:
This provides a mock alerts list with composite expansion.
mock-alerts


jshaughn:
I like the drop down idea and the list entries look relatively clean.  Good start, this will be a work in progress as alerts/events have a lot of filtering and a lot of data to present.

PR #406, opened 2017-07-24
fix npe

jmazzitelli:
This happens when a person declares a condition as "threshold" but doesnt' specify a threshold value (this happens when the person really meant it to be "range' and thus had thresholdLow and thresholdHigh instead). Clean up NPE so error message is more clear what happened.


jmazzitelli:
Before I would get this:

```
16:23:40.940 [AlertsEngineImpl-Timer] ERROR o.h.a.e.i.AlertsEngineImpl - HAWKALERT220001: Error processing rules: [Exception executing consequence for rule "Threshold" in org.hawkular.alerts.engine.rules: java.lang.NullPointerException]
```

With this PR the error message is more clear - at least it tells you WHICH trigger has a problem:

```
17:13:39.812 [AlertsEngineImpl-Timer] ERROR o.h.a.e.i.AlertsEngineImpl - HAWKALERT220001: Error processing rules: [Exception executing consequence for rule "Threshold" in org.hawkular.alerts.engine.rules: java.lang.IllegalStateException: Invalid threshold for condition: ThresholdCondition [triggerId='simple-generator-trigger-low', triggerMode=FIRING, dataId='data-x', operator=null, threshold=null]]
```

jshaughn:
This is helpful for Threshold although it may be even better to prevent persisting these invalid conditions.  Maybe we could add an abstract Condition.isValid() and make sure to call it from setConditions().   Any desire to do that or are you happy as is?

jmazzitelli:
I'm not sure the condition WAS persisted. This is happening when the trigger is being created - so its possible this exception when thrown aborts the save. Anyway, I'm ok with this PR as is for now. At least it now tells you WHAT trigger is at fault and you can fix

lucasponce:
This sounds to me more a missing default value when reading a JSON object.
It seems in the .next version I missed some flow and it could happen a non expected NPE.
By anyway, additional protection is good.

PR #420, opened 2017-08-01
Support for Kafka Action Plugin

lucasponce:
This PR adds a new Action Plugin to integrate with Apache Kafka.
It reuses same design of Elasticsearch plugin adding transformation cappabilities.

A Kafka Alerter will be sent in a following PR, but this one is independent and it has an example to be tested.


actions/actions-plugins/actions-kafka/src/main/java/org/hawkular/alerts/actions/kafka/KafkaPlugin.java
```diff
@@ -0,0 +1,248 @@
...
+ *
+ * @author Jay Shaughnessy
+ * @author Lucas Ponce
+ */
+@Plugin(name = "kafka")
+public class KafkaPlugin implements ActionPluginListener {
+    public static final String PLUGIN_NAME = "kafka";
+
+    /**
+     * "url" property is used to indicate the Kafka server or servers to connect.
```
jmazzitelli: What is this "url" javadoc? I think this might be something that was left over after a cut-n-paste? Probably should be deleted.

jmazzitelli:
LGTM other than that small javadoc comment that seems out of place

actions/actions-plugins/actions-kafka/src/main/java/org/hawkular/alerts/actions/kafka/KafkaPlugin.java
lucasponce: Oh, true, I left over from another plugin. I will delete it.

lucasponce:
Thx for the review @jmazzitelli. Merging this.

PR #422, opened 2017-08-02
Dockerfile scripts

lucasponce:
Basic scripts for docker.
It is not automated as I think images should be controlled when are deployed, but basis to build images are in place now.


lucasponce:
@jmazzitelli please could you review ? thx

jmazzitelli:
> It is not automated as I think images should be controlled when are deployed

Typically what projects do (this is what HOSA does) is the docker image is automatically built and uploaded by travis to docker hub with the docker tag "latest". Thus, every master build is immediately available for use by anyone by pulling the docker hub image using the "latest" tag. This is nice for community users to pick up the latest builds and test with it without having to git clone and build themselves.

You can control other images you upload by manually releasing images using docker tags such as "1.0.0". Or you can automate that as well by using git tags as the docker image tag. Again, HOSA does this also. For examples, see: https://github.com/hawkular/hawkular-openshift-agent/blob/master/.travis.yml#L39-L63

lucasponce:
Ok, sounds good, I will follow the same config like HOSA.

lucasponce:
To test and troubleshooting the docker push I need to merge this one, as credential don't work from PRs for security reasons.

PR #423, opened 2017-08-02
better behavior in the charts

jmazzitelli:
This fixes the ugly way the charts in the top left of the dashboard were getting rendered. Before they would "flash" every few seconds (each update) because they were getting created each time. Now we only create them once and we only re-load the data each update. We get nice animations when we fix it like this and we do not see flashing anymore and the interactivity of the charts (e.g. clicking the legend to hide slices of data) works again (before they would reset themselves every update).

This also fixes the pie chart so it only shows non-resolved alerts. This makes it consistent with the bar chart so now both charts render the same alerts (that is, only open and ack'd alerts). This allows the pie chart to show more interesting data (i.e. the percentages of critical alerts that need to be resolved, compared to the percentages of high, med, low alerts needed to be resolved).

The pie chart tooltip now provides both percentage and raw data value for each slice (so hovering over a pie slice tells you how many unresolved alerts of that given severity are in the system along with the percentage).


lucasponce:
LGTM

PR #426, opened 2017-08-08
Remove cassandra dependencies from .next version

lucasponce:


lucasponce:
@jshaughn I rebased .next to solve some conflicts but this should help travis.
We can revert this commit to bring it back (if necessary for .next).

jshaughn:
I actually wish we had done this earlier :)  We spent time keeping it in sync.  Time to break up with Cassandra, "it's not us, it's you, Buh-bye!"

PR #427, opened 2017-08-09
HWKALERTS-275 Additional tests on Event queries

lucasponce:
This PR adds more testing on Events.
Also it will use travis to validate if tests are passing on the CI.


jshaughn:
This is good, although I will hold off merging until we can explain the issue being reported.

PR #429, opened 2017-08-09
HWKALERTS-275 Additional tests on Event queries

lucasponce:
This is a similar test for 1_6_x branch, also with a increased # of events.
I sent it also to check if travis has any issue with it.


lucasponce:
@jshaughn @gbaufake this test works locally for alerting with a pretty decent amount of data.
@gbaufake can you adjust some reproduce steps on your side ?
Thx

PR #430, opened 2017-08-10
Add Events tab into UI

lucasponce:
Ok, a first Events tab to complete the details of what we show.
Mostly reused the Alerts tab and linked from Dashboard.
Ah, also put in order Dashboard, Alerts, Events as it is close and leave to the end the definitions/actions.

I'm going to optimize some missing formats on Events but in a following PR.

image

### DIFF
--- a/ui/src/main/ui/src/alerts/alerts-controller.js
+++ b/ui/src/main/ui/src/alerts/alerts-controller.js
@@ -139,7 +139,7 @@ angular.module('hwk.alertsModule').controller( 'hwk.alertsController', ['$scope'
       }
     };
 
-    $scope.$on('ngRepeatDone', function(ngRepeatDoneEvent) {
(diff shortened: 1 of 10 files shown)


lucasponce:
@jshaughn or @jmazzitelli please take a look. Thx.

jshaughn:
LGTM for initial event support.

jmazzitelli:
looks good. FYI: Jay has new date-time picker - he should add it to this new events page to replace the date picker here.

lucasponce:
Yeah, I release the PR with minimal changes as I was aware Jay could be working on similar area.
So, to minimize conflicts you can rebase on this and add changes you need.

jshaughn:
Yes, my goal is to have a global datetime range applied to the relevant pages by the end of today.

PR #431, opened 2017-08-10
Add robust datetime range filtering for Alerts querying.

jshaughn:
This was a pita. It handles the alerts view, next will be to apply it globally.


jmazzitelli:
we can merge. next up we should get this to apply to events and dashboard, but if we make this global, that should be taken care of.

PR #432, opened 2017-08-10
example that shows alerts taking in data from multiple different sources

jmazzitelli:


jmazzitelli:
I suggest someone pull this, run "mvn test" and see if it works for you

jmazzitelli:
I think for now this is good. It at least shows alerts and events generated from kafka topic data, elasticsearch logs, and prometheus queries.

lucasponce:
I've tested the example.
It looks good but in the UI there are some things broken in the Events side.
Also this example is good to check that UI doesn't render correctly trigger definitions.
But it is good, the UI issues are not related to this.

PR #438, opened 2017-08-16
Set default JDK to openjdk8

josejulio:


lucasponce:
LGTM
Thanks @josejulio

PR #439, opened 2017-08-20
Pending features developed in master not brought into .next

lucasponce:


engine/src/main/antlr4/org/hawkular/alerts/engine/tags/parser/TagQuery.g4
```diff
@@ -127,4 +127,6 @@ fragment V: [vV];
 fragment W: [wW];
 fragment X: [xX];
 fragment Y: [yY];
-fragment Z: [zZ];
\ No newline at end of file
+fragment Z: [zZ];
+
+fragment NEG_OP: '~';
```
lucasponce: In the current tagQuery language http://www.hawkular.org/docs/rest/rest-alerts.html a <tag_value> can include a regexp if it is defined between simple quotes.
This '~' char is a potential code to be escaoed in examples like:
```
def tagQuery = "test_tag = '\\/t;hawkular\\/f;my-agent\\/r;Local%20DMR\\~\\~_Server Availability'"
```
And it was not supported by the lexer.
This one is fixed as it might look like a corner case but it is common in for Hawkular Services.

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
```diff
@@ -113,26 +114,28 @@ public void init() {
...
                     String value;
                     if (tokens.size() == 3) {
-                        query.append("'").append(tag).append("' and (");
                         op = tokens.get(1);
                         value = tokens.get(2);
                         boolean isRegexp = value.startsWith("'");
-                        String regexp = value.substring(1, value.length() - 1);
-                        regexp = regexp.equals("*") ? ".*" : regexp;
+                        String regexp = "";
+                        if (isRegexp) {
```
lucasponce: Bugfix.
This logic only has be processed when it is a regexp.

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
```diff
@@ -113,26 +114,28 @@ public void init() {
...
+                            regexp = value.substring(1, value.length() - 1);
+                            regexp = regexp.equals("*") ? ".*" : regexp;
+                        }
                         if (op.equalsIgnoreCase(EQ)) {
                             // tag =
                             if (isRegexp) {
-                                query.append("/").append(regexp).append("/");
+                                query.append("/").append(tag).append(TagsBridge.SEPARATOR).append(regexp).append("/");
```
lucasponce: The new format how internally tags are represented in the index engine is defined in TagsBridge class.
Potential constants to build expression will be pointed from there.

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/model/TagsBridge.java
```diff
@@ -44,9 +73,79 @@ public void set(String name, Object value, Document document, LuceneOptions luce
...
+            String tags = null;
+            try {
+                char[] buff = new char[1 * 1024];
+                StringBuilder buffer = new StringBuilder();
+                while ((nRead = input.read(buff, 0, buff.length)) != -1) {
+                    buffer.append(buff, 0, nRead);
+                }
+                tags = buffer.toString();
+            } catch (IOException e) {
+                e.printStackTrace();
```
burmanm: printStackTrace to stdout directly? Nothing else as error processing?

burmanm: printStackTrace to stderr directly? Nothing else as error processing?

lucasponce: +1

lucasponce:
Thanks for your review @burmanm.

PR #440, opened 2017-08-21
HWKALERTS-275 Use IN operator on Events/Alerts queries

lucasponce:
This change prevents to get an 
```
com.datastax.driver.core.exceptions.BusyPoolException
```
when the number of filtered Events/Alerts exceeds defined max queue size.
Now, instead to query individually, it uses the IN operator to group queries in a configurable max items size.


lucasponce:
PR tested
https://issues.jboss.org/browse/HWKALERTS-275
Merging and applying into 1.7.x branch.

jshaughn:
Nice way to avoid too many concurrent requests.

PR #447, opened 2017-08-30
Enable docgen for v2 from travis

lucasponce:
In .next I disabled docgen until we have migrated the API doc.
Now, travis should handle v1 and v2 versions.
Once these files are in the repository, we need to create a proper v1/v2 entries in the doc.

### DIFF
--- a/.travis.swagger.sh
+++ b/.travis.swagger.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+#
+# Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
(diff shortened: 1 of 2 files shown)


jshaughn:
I guess the .travis.swagger.sh file is just for consistency to get things to run, even though we're not using swagger?  LGTM, just confirming this is the filename we want.

lucasponce:
Yes, it has to do with the target repository which .adoc are stored (which is also called swagger), but all scripts are fine.

lucasponce:
@jshaughn does this one more steps ?

PR #448, opened 2017-08-30
HWKALERTS-275 Use IN operator on Events/Alerts queries

lucasponce:


lucasponce:
The 1.8.x was closed/missing without any reason, some problem from my side.

PR #453, opened 2017-09-01
fix docker push

jmazzitelli:
in .travis.swagger.sh which is called just before the docker push stuff, a mvn clean is performed:

```
./mvnw clean compile -Pdocgen -DskipTests -Dcheckstyle.skip -Dlicense.skip
```

So the clean deletes the docker stuff in dist/target which is required by docker build (the build-dist.sh takes the Dockerfile and runs docker build which copies files that should be in target/* but are now deleted).

The error you get will be:

```
Step 4/8 : COPY target/hawkular-alerting* /opt/hawkular-alerting
No source files were specified
```

That's the error happening in build-dist.sh (which is the docker build stuff), and thus the docker image never gets created and thus why this happens:

```
An image does not exist locally with the tag: hawkular/hawkular-alerts
```

We need to make sure the dist/target/* files remain intact when we build the docker image and push to docker hub


jshaughn:
LGTM

PR #461, opened 2017-09-08
Migrate tutorials to V2

jshaughn:
- Remove use of HAM/Cassandra containers and replace with hAlerts container
  - So, tutorial now uses only hAlerts
  - Update impacted scripts, instructions, etc
- Insert new UI tutorial as lesson-04
  - Update impacted asciidoc files with new lesson numbering
- A few minor cleanups along the way


lucasponce:
Good changes.

PR #462, opened 2017-09-11
Fix the dash range filter widgets

jshaughn:
... (touchspin and calendar were not working)

- remove some superfluous ng-hide/ng-if attributes, they confused the rendering
- fix issue with "Set Range To" button click
- stop the controller from rendering timeline when there is nothing to show
  - this saves work and stops some console errors

Also: remove preset tenant


ui/src/main/ui/src/app/controllers/app-controller.js
```diff
@@ -67,7 +67,7 @@ angular.module('hwk.appModule').controller( 'hwk.appController', ['$scope', '$ro
     $scope.newTenant = {};
 
     // [lponce] comment if you don't want a pre-filled tenant
-    $scope.newTenant.tenant = 'my-organization';
```
lucasponce: -1
I think that at present status, where the UI is a selling tool, having the default tenant used in examples is good.

jmazzitelli: I agree with @lucasponce - why remove the pre-set tenant?

jshaughn: I disagree,  I think it is confusing.  Also, the tutorials use tenant 'tutorial'.  If we are going to preset it then I would vote for 'tutorial', but really I think it should be empty.  Let's let @jmazzitelli cast the deciding vote.

lucasponce: Perhaps if there is some example/tutorial using a different tenant, we should use the same to let people to not introduce the tenant in the UI.

jshaughn: We could make it a dropdown with some pre-fill choices :)  [my-organization, tutorial, test]

jshaughn: Democracy rules, I will return the 'my-organization' tenant, even though you're both wrong :)

jmazzitelli: Why are the tutorials using "tutorial"? Why are we using different defaults? From what I remember, all the examples up to now have been using the same "my-organization"? We should be using the same for all examples/tutorials.

lucasponce: I think that tutorial could use my-organization (or the name you like here) but ALL examples should use the same.
User should drive this and less steps for him/her is better.

jmazzitelli: > but ALL examples should use the same.

+1 Why make it harder on ourselves by using different names in different examples? Use the same everywhere.

jshaughn: The tutorials are using 'tutorial' because it's simple and made sense to me.  And  it ensured at least that the examples and tutorials won't step on each other.  If we want to consolidate why not just change everything to something simple, like 'tenant'.  I don't care much for 'my-organization', it's verbose.  Anyway, I don't relish the idea of editing all of the files, so perhaps mazz can do that, in which case he can choose anything he likes :)

PR #465, opened 2017-09-13
license/checkstyle/xml-format compliance

jshaughn:
- update files to pass all of the standard hawkular checkers
  - note, we don't apply an canonical formatting to .java although
    it is expected that either eclipse or IJ formatting is applied.
- use latest hawkular parent pom (version 55) to set up the various
  checkers.
- add new -Dcheck to run the checkers, they won't run by default


alerters/alerters-plugins/alerters-prometheus/src/main/java/org/hawkular/alerter/prometheus/PrometheusAlerter.java
```diff
@@ -146,7 +146,8 @@ public void stop() {
     private void initialRefresh() {
         try {
             Collection<Trigger> triggers = definitions.getAllTriggersByTag(ALERTER_ID, "*");
-            triggers.stream().forEach(trigger -> activeTriggers.put(new TriggerKey(trigger.getTenantId(), trigger.getId()), trigger));
+            triggers.stream().forEach(
```
lucasponce: Changes on license files, imports, or config files are ok.
Only license are necessary, others are optional and are a matter of style agreement.
I don't like changes on code like this, to be honest, it's not value doing this.
The style checker should avoid these type of changes are least in 2.x project.

lucasponce:
Merging, perhaps some rules to review indentation could be revisit.

alerters/alerters-plugins/alerters-prometheus/src/main/java/org/hawkular/alerter/prometheus/PrometheusAlerter.java
jshaughn: That was probably just my fault while dorking with the file for some other reason.  I maybe applied the IDE formatter by mistake.

jshaughn: Because we use different IDE's we will likely always have a few of these issues.

PR #467, opened 2017-09-15
Gnocchi Alerter

lucasponce:
Background

Hawkular Alerting is a federated alerting engine. Today it offers a REST API to integrate with any system and a set of pre-built integrations with Elasticsearch, Prometheus and Apache Kafka.

This PR adds Gnocchi into the alerter plugins.

In the context of the discussion
https://github.com/gnocchixyz/gnocchi/issues/71

We commented that we could add a poll alerter as a first step to cover Gnocchi and collect requeriments.
If the features Hawkular Alerting offers are interesting we can propose a fully streaming approach being Gnocchi owner to send data into Hawkular Alerting.

Anyway, this is just a first step to add alerting capabilities into Gnocchi.

cc @seuf @chungg @sileht @jd

### DIFF
--- a/alerters/alerters-plugins/alerters-gnocchi/src/main/java/org/hawkular/alerter/gnocchi/GnocchiQuery.java
+++ b/alerters/alerters-plugins/alerters-gnocchi/src/main/java/org/hawkular/alerter/gnocchi/GnocchiQuery.java
@@ -17,6 +17,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
@@ -130,9 +131,9 @@ public class GnocchiQuery implements Runnable {
             for (int i = 0; i < splitMetricNames.length; i++) {
                 gnocchiNames.add(splitMetricNames[i].trim());
             }
-            List rawAllMetrics = getAllMetrics();
-            for (int i = 0; i < rawAllMetrics.size(); i++) {
-                Map<String, String> metric = (Map<String, String>) rawAllMetrics.get(i);
+            List rawMetrics = getMetrics(gnocchiNames);
+            for (int i = 0; i < rawMetrics.size(); i++) {
+                Map<String, String> metric = (Map<String, String>) rawMetrics.get(i);
                 if (gnocchiNames.contains(metric.get("name"))) {
                     nameIdMetrics.put(metric.get("name"), metric.get("id"));
                 }
@@ -290,6 +291,31 @@ public class GnocchiQuery implements Runnable {
             log.errorf(e,"Error querying Gnocchi metrics %s", allMetricsUrl);
         }
         return Collections.EMPTY_LIST;
+    }
+
+    private List<Map<String, String>> getMetrics(List<String> gnocchiNames) {
+        String metricsUrl = baseUrl + "/v1/metric";
+        List<Map<String, String>> rawMetrics = new ArrayList<>();
+        if (isEmpty(gnocchiNames)) {
+            return rawMetrics;
+        }
+        for (String metricName : gnocchiNames) {
+            String metricUrl = metricsUrl + "?name=" + metricName;
+            try {
+                URL url = new URL(metricUrl);
+                HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+                conn.setRequestProperty(AUTHORIZATION, basicAuth);
+                conn.setRequestMethod(GET);
+                conn.setDoInput(true);
+                List rawMetric = JsonUtil.getMapper().readValue(conn.getInputStream(), List.class);
+                conn.disconnect();
+                log.debugf("Gnocchi Metrics found %s", rawMetric);
+                rawMetrics.addAll(rawMetric);
+            } catch (IOException e) {
+                log.errorf(e,"Error querying Gnocchi metrics %s", metricUrl);
+            }
+        }
+        return rawMetrics;
     }
 
     private List<Map<String, String>> getResourceMetrics() {
--- a/alerters/alerters-plugins/alerters-gnocchi/src/main/java/org/hawkular/alerter/gnocchi/GnocchiQuery.java
+++ b/alerters/alerters-plugins/alerters-gnocchi/src/main/java/org/hawkular/alerter/gnocchi/GnocchiQuery.java
@@ -50,6 +50,8 @@ public class GnocchiQuery implements Runnable {
     private static final String METRIC_NAMES_REGEXP = "metric.names.regexp";
     private static final String METRIC_RESOURCE_QUERY = "metric.resource.query";
     private static final String METRIC_AGGREGATION = "metric.aggregation";
+    private static final String METRIC_GRANULARITY = "metric.granularity";
+    private static final String METRIC_GRANULARITY_DEFAULT = "300";
 
     private Trigger trigger;
     private Map<String, String> properties;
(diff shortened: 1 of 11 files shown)


alerters/alerters-plugins/alerters-gnocchi/src/main/java/org/hawkular/alerter/gnocchi/GnocchiQuery.java
```diff
@@ -0,0 +1,391 @@
...
+                    conn.setRequestMethod(GET);
+                    conn.setDoInput(true);
+                    List measures = JsonUtil.getMapper().readValue(conn.getInputStream(), List.class);
+                    List<Data> data = new ArrayList<>();
+                    for (int i = 0; i < measures.size(); i++) {
+                        List measure = (List)measures.get(i);
+                        if (measure != null && measure.size() == 3) {
+                            String timestamp = (String)measure.get(0);
+                            Double value = (Double)measure.get(2);
+                            data.add(Data.forNumeric(trigger.getTenantId(), metricName, parseTimestamp(timestamp), value));
```
chungg: i think there is something missing here (or when you're building the queries). in gnocchi, if you dont' specify a granularity (http://gnocchi.xyz/operating.html#default-archive-policies), it will return back all granularities associated with metric. if you're using 'low' archive policy, this is only one granularity and you're fine. but if you're using 'medium', 'high', and presumably any custom policy, it will have multiple granularities returned and if you ignore it or don't specify it, you'll be effectively squashing multiple series into one.

lucasponce: Yeah, in this first draft I'm taking the default granularity.
So effectively ignored.
I think we can add the granularity as part of the configuration (and if not present then fall into the default one).
I think yes, but in Gnocchi is a normal task to ask for same granularity in a group of metrics ?
i.e.
The example there there are 8 metrics for cpu, cpu-.@cpu-user-0, I guess that we can use the same granularity for all metrics fetched.

Or, is there another way where gnocchi users are familar ? I would like that the integration will be most natural for them (in the context that they need to define them).

chungg: hmm... there really isn't a default granularity. when you create a metric, you must specify an archive policy (if i remember correctly). if the metric is a 'low' archive policy, then there is only one granularity (a timeseries of 5min aggregates over 30 days) and you have a 'default' granularity. but if they created a metric using another policy such as 'medium', then there will be multiple time-series (captures one series at 1min aggregates for 7 days, and another series at 1hr aggregates for a year). in the 'medium' case, if you get_measures without a granularity,  it will return both series, which will give you two different, partially overlapping (at different granularities), series.

for example, if you send in a point for 2017-01-01T00:05:32 with value 5. you will get back something roughly like:

`[(2017-01-01T00:00:00, 3600, 5),
  (2017-01-01T00:05:00, 60, 5)]`

two different series in a single result

lucasponce: Good point, I will adapt the alert to take this case.

jshaughn: That's interesting, so you can get potentially multiple running aggregates depending on the archive policy.  @lucasponce, I guess we either need to limit to one result or handle multiple results in a sane way.

lucasponce: @jshaughn yes, that is was I try to comment.
The alerer takes granularity by default thinking that only one result was given but as multiple can be returnerd, granularity needs to be explicit in the trigger configuration to get a single result.

lucasponce:
Added granularity configuration and propagate this info into Hawkular Data for tracing.
Thx for your review @chungg, please let us know if you think alerter could be merged as a first draft.

lucasponce:
Let's see if we have today some coment about it or we can merge and prepare an article about it to demo it.
As commented in the IRC, this is just a preliminar integration plugin, future changes can go in next community versions of alerting 2.x series.

alerters/alerters-plugins/alerters-gnocchi/src/main/java/org/hawkular/alerter/gnocchi/GnocchiQuery.java
```diff
@@ -0,0 +1,403 @@
...
+        }
+
+        // Case 2, fixed list of metric names, a query to get gnocchi ids is necessary
+        if (!isEmpty(metricNames) && isEmpty(metricResourceQuery)) {
+            String[] splitMetricNames = metricNames.split(",");
+            List<String> gnocchiNames = new ArrayList<>();
+            for (int i = 0; i < splitMetricNames.length; i++) {
+                gnocchiNames.add(splitMetricNames[i].trim());
+            }
+            List rawAllMetrics = getAllMetrics();
```
jd: I'd suggest passing at least `?name=metric_name` to the query to limit the output. If you got millions of metrics it's gonna be a problem. :)

alerters/alerters-plugins/alerters-gnocchi/src/main/java/org/hawkular/alerter/gnocchi/GnocchiQuery.java
```diff
@@ -0,0 +1,403 @@
...
+                    nameIdMetrics.put(metric.get("name"), metric.get("id"));
+                }
+            }
+            log.debugf("Gnocchi Metrics configured %s", nameIdMetrics);
+            return;
+        }
+
+        // Case 4, a resource query is defined, so metrics are filtered by this query
+        if (!isEmpty(metricResourceQuery)) {
+            List rawResourceMetrics = getResourceMetrics();
```
jd: That seems pretty wide as a query. You can use the search API to have a smaller set of results.

lucasponce: +1 indeed it is used, read from the trigger configuration
https://github.com/hawkular/hawkular-alerts/pull/467/files/9348a574844de42dd4ef19ddaa853020c43b154a#diff-f7efcfbcd8455ebf114f6e51ced5131bR155
So, basically, it uses the search API to narrow the resources to fetch.

alerters/alerters-plugins/alerters-gnocchi/src/main/java/org/hawkular/alerter/gnocchi/GnocchiQuery.java
lucasponce: +1, it is the default case when it is not defined, an alternative is to throw some error message forcing user to define it, but yes, idea is that user can define metrics in several ways
https://github.com/hawkular/hawkular-alerts/pull/467/files#diff-f7efcfbcd8455ebf114f6e51ced5131bR111
i.e.
metric.ids, or
metric.names, or
metric.names.regexp, and
metric.resource.query, to even narrow the potential metrics to search.

Also, this is performed only once, to resolve internally metric name -> metric id which will be used for query measures.

lucasponce:
So, added suggestions from @jd, if metrics.names defined, then it uses the name filter to not fetch all metrics on client.
The resource metrics query is already filtered by a gnocchi resource query using search endpoint.

Only in the worst case, that there is nothing defined it will fetch all metrics defined.

lucasponce:
@jshaughn If there is no more comments/suggestions I think the alerter is in good shape to be merged.

jshaughn:
OK, this seems to be in a good initial state,  merging.  Further comments/revisions can be handled in follow-up PRs.  Thanks @lucasponce, @jd and @chungg !

jd:
Guys, if you want to send a PR on Gnocchi to update our doc to point to whatever makes sense on Hawkular side, please feel free to open one!

lucasponce:
+1
We need to update the Gnocchi Alerter doc and prepare a blog post but it is a good idea.
Thanks !

PR #468, opened 2017-09-15
Fix root pom profile issue

jshaughn:
- continue to skip all the checkers by default
- enable checking with -Pcheck

@lucasponce sorry for the confusion, should have created this PR yesterday.  I think this simplifies things, just add -Pcheck if you want to run the checkers.


jshaughn:
@lucasponce Low priority but is this OK?  It gets rid of the parent pom and adds an explicit profile -Pcheck to run the various checkers.

lucasponce:
LGTM thanks.

PR #469, opened 2017-09-21
Minor Tweaks

jshaughn:
- remove some leftover cassandra references
- fix small issue in severities query

@lucasponce A few small cleanups I came across when working on 1.8.

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -48,13 +48,13 @@
 
     <dependency>
       <groupId>org.hawkular.alerts</groupId>
-      <artifactId>hawkular-alerts-filter-api</artifactId>
+      <artifactId>hawkular-alerts-events-aggregation</artifactId>
       <version>${project.version}</version>
     </dependency>
 
     <dependency>
       <groupId>org.hawkular.alerts</groupId>
-      <artifactId>hawkular-alerts-events-aggregation</artifactId>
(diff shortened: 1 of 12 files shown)


engine/pom.xml
```diff
@@ -48,13 +48,13 @@
 
     <dependency>
       <groupId>org.hawkular.alerts</groupId>
-      <artifactId>hawkular-alerts-filter-api</artifactId>
+      <artifactId>hawkular-alerts-events-aggregation</artifactId>
```
lucasponce: mmm ?

engine/pom.xml
```diff
@@ -48,13 +48,13 @@
...
       <version>${project.version}</version>
     </dependency>
 
     <dependency>
       <groupId>org.hawkular.alerts</groupId>
-      <artifactId>hawkular-alerts-events-aggregation</artifactId>
+      <artifactId>hawkular-alerts-filter-api</artifactId>
```
lucasponce: Same mmm ?

lucasponce:
Good clean-up of old references.
Merging.

engine/pom.xml
jshaughn: yes, no changes to the deps, just flipped the order to be alphabetical.  Probably didn't need to do that :)

PR #470, opened 2017-09-23
Openshift template

gbaufake:
Hi @jshaughn and @lucasponce.

Based on @jmazzitelli k8s template, I created a openshift templated

I think we can use that template to some testing under openshift. On the future, we can create a setup running some of the hawkular-alerts integration running on openshift such as Kafka, Prometheus, Gnocci, etc.

### DIFF
--- a/dist/hawkular-alerts-openshift.yml
+++ b/dist/hawkular-alerts-openshift.yml
@@ -1,0 +1,75 @@
+#
+# Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


dist/hawkular-alerts-openshift.yml
```diff
@@ -0,0 +1,75 @@
...
+id: hawkular-alerts
+kind: Template
+apiVersion: v1
+name: Hawkular Alerts
+metadata:
+  name: hawkular-alerts
+
+
+objects:
+- apiVersion: extensions/v1beta1
```
lucasponce: This "v1beta1" refers to the API version of OpenShift or Alerting ?

jmazzitelli: that is a specific OpenShift (actually it might be K8S) version. Has nothing to do with alerts.

lucasponce:
LGTM

PR #471, opened 2017-09-25
Remove unused schema cache

jshaughn:
@lucasponce Some minor cleanup


lucasponce:
Good catch.
Note to self in the future: 
This cache was used to synchronize C* initializataion on multiple nodes.

PR #473, opened 2017-09-27
ISPN Backend

jshaughn:
@lucasponce A large PR but of course it is a large change.
- Replace Cassandra with ISPN/HibernateSearch (9.1.1/5.8.0 non-embedded) backend
- Remove Metrics deployments

It is currently using a src-dep for commons mwm-wildfly branch.


jshaughn:
This will need to be properly squashed, but for now it is available for review and travis.

hawkular-alerts-api/src/main/java/org/hawkular/alerts/api/model/trigger/Trigger.java
```diff
@@ -58,154 +60,115 @@
 
     private static final long serialVersionUID = 1L;
 
-    @ApiModelProperty(value = "Tenant id owner of this trigger.",
-            position = 0,
-            required = true,
-            allowableValues = "Tenant is overwritten from Hawkular-Tenant HTTP header parameter request")
+    @ApiModelProperty(value = "Tenant id owner of this trigger.", position = 0, required = true, allowableValues = "Tenant is overwritten from Hawkular-Tenant HTTP header parameter request")
```
lucasponce: Why these changes of formatting ?
Please, let's respect formats.
IDE should be put in mode "don't modify formats automatically", in this case is painful, as these document annotations are hard to read without proper formatting.
I'm tired to battle about this.

I think the rule about "ok, if someone defines a format first, let's respect it" is enough fair.

hawkular-alerts-commons/src/main/java/org/hawkular/alerts/cache/IspnCacheManager.java
```diff
@@ -0,0 +1,88 @@
...
+            cacheManager = null;
+        }
+    }
+
+    private static synchronized void init() {
+        if (cacheManager == null) {
+            try {
+                distributed = Boolean.valueOf(HawkularProperties.getProperty(ALERTS_DISTRIBUTED, ALERTS_DISTRIBUTED_ENV,
+                        ALERTS_DISTRIBUTED_DEFAULT));
+                String configPath = System.getProperty(CONFIG_PATH);
```
lucasponce: Not for this PR.
Inventory is using ${jboss.server.config.dir}/hawkular to read config files to read config files.
It could be good to unify them at some point.
(If not found, it read from classloader).

lucasponce:
A first review looks good.
I have a minor but frustrating comment about formatting wars.

jshaughn:
@lucasponce Feedback incorporated and squashed.  DNM label removed.  Thanks for review, waiting now to go green.

PR #475, opened 2017-09-29
Align status endpoint with UP/DOWN

lucasponce:
Minor change to aling the same status value from inventory and alerting.


lucasponce:
Found a minor issue with this change. Added DNM label.

PR #476, opened 2017-10-04
Prometheus Alerter Group Trigger Support

jshaughn:
- Support tokenized expressions so that member triggers
  can customize the promql expression string. Tokens are
  supplied as additional context on the member.
- Add itest module and fix up all tests
- Add -Pprometheus to run both unit and itests
  - requires a P running on default ports, and some other

Also:
- Enable prometheus alerter by default
- Fix some doco in ExternalCondition

@lucasponce Please review.  I think this is what we need in order to use this in a live alerting scenario.


hawkular-alerters/hawkular-prometheus-alerter/hawkular-prometheus-alerter-itest/pom.xml
```diff
@@ -0,0 +1,367 @@
...
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <profiles>
+
+    <profile>
+      <id>prometheus</id>
```
lucasponce: +1 to have a prometheus profile
Also, thinking loud I guess we could configure travis to install a prometheus to automate also this in some way.
(We did similar with Cassandra, and after some battles we could have a decent CI scripting).
Not for this PR, of course, but something to keep in mind we will need here in Alerting and in hServices sooner than later.

hawkular-alerters/hawkular-prometheus-alerter/hawkular-prometheus-alerter-itest/src/test/groovy/org/hawkular/alerter/prometheus/PrometheusITest.groovy
```diff
@@ -84,11 +82,10 @@ class prometheusITest {
         assertEquals(200, response.status)
     }
 
-    @Ignore
     @Test
     void prometheusAlerterAlertTest() {
         def start = String.valueOf(System.currentTimeMillis())
-        // load external trigger
+        // load external group trigger
```
lucasponce: Minor, I think this test is using prometheus-alert-trigger.json which doesn't use group trigger, so the comment led me to think the json file was wrong, but I think is the comment which is bad.

hawkular-alerters/hawkular-prometheus-alerter/hawkular-prometheus-alerter-lib/src/main/java/org/hawkular/alerter/prometheus/ConditionManager.java
```diff
@@ -81,7 +91,7 @@
 
     private static final String PROMETHEUS_ALERTER = "hawkular-alerts.prometheus-alerter";
     private static final String PROMETHEUS_ALERTER_ENV = "PROMETHEUS_ALERTER";
-    private static final String PROMETHEUS_ALERTER_DEFAULT = "false";  // do not turn on prom-alerter by default
+    private static final String PROMETHEUS_ALERTER_DEFAULT = "true";
```
lucasponce: +1

lucasponce:
Apologizes for my delay reviewing this one.
LGTM merging

PR #479, opened 2017-10-09
Triggers API: Orphan/Unorphan to use PUT instead of POST

burmanm:
Proposal to make REST-API more consistent.

Change orphan/unorphan triggers to follow PUT as they should behave like idempotent methods. Update is already PUT method and these should be for that reason too.


lucasponce:
LGTM
examples and tutorials are not automatically tested, I don't remember if these methods were used there.
Could you take a look and leave a comment if everything is fine ?
I will merge then.

jshaughn:
I'm OK with this despite an API change.  I know these are not used in the tutorial as their is not yet a Group Trigger tutorial.  @burmanm , you're everywhere :)

Issue #480, opened 2017-10-11
REST-interface documentation is not created correctly


burmanm:
Some methods are not visible in the REST-API documentation, such as

```java
router.get(path + "/:triggerId").handler(this::getTrigger);
```

The Java-code has the API documentation, but it's not in the hawkular.org website generated documentation.

lucasponce:
Thx for the issue report.
Yes, this seems like a bug.

PR #481, opened 2017-10-11
[HWKALERTS-287 REST-API is missing multiple methods for same path 

burmanm:
Change hasProperty to containsKey to get all the methods printed in the documentation for same path.


burmanm:
Fixes #480

jshaughn:
Thanks @burmanm !

PR #482, opened 2017-10-12
Fix groupId related documentation in REST API

burmanm:
There were few copy-paste mistakes as well as inconsistent description for triggerId


jshaughn:
Thanks, @burmanm. Cleaner now.

PR #483, opened 2017-10-18
Ui

jshaughn:


hawkular-alerts-ui/src/main/ui/bower.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "hawkular-alerts-ui",
+  "version": "2.0.0",
```
lucasponce: Minor, this is just informational and it doesn't mean to deploy anything, but I guess we should put here a 1.8.0.

hawkular-alerts-ui/src/main/ui/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "hawkular-alerts-ui",
+  "version": "2.0.0",
```
lucasponce: Same here

lucasponce:
A couple of minor comments.
Also I have a suggestion to put the version of the hawkular alerting somewhere in the UI dashboard, as we have 1.8.x and 2.0, it could be good to have some minor text on the UI.

hawkular-alerts-ui/src/main/ui/bower.json
jshaughn: To be honest, this manually maintained versioning is unlikely to stay updated.  But at least we can start with a reasonable version.  I'll change.

jshaughn:
@lucasponce I updated the version in the two files.  More importantly, I added an "About..." option to the dash which pops up an about box showing our /status info.

lucasponce:
thanks

PR #487, opened 2017-11-13
Fixing references to old alerts page

xeviknal:
This PR fix all URL references to old alert page (alerts.html) to the new one (alerts-v2.html).
Also includes one adoc syntax problem on lesson-2: using `gauge-1' instead of `gauge-1`


lucasponce:
LGTM

PR #488, opened 2017-11-13
HAWKULAR-1331 Enable JDG profile on 1.8

lucasponce:
- Align with Inventory for consistency on config files
- Bug fixes


jshaughn:
Good job getting this working on both app server platforms.  I suppose a similar approach is needed for inventory?

lucasponce:
Inventory is already done (with similar layout of profiles and variables to use the same way to build).
Services is not yet prepared but that perhaps is not necessary, as it will depend on how it will be productized.

PR #489, opened 2017-11-17
Expose Prometheus endpoint for Hawkular Services uses

lucasponce:
Minor one, as we might want to reuse same property for Hawkular Services and then it makes sense to not make it dependent on alerting.


lucasponce:
DNM label.
I'm not fully convinced of this PR I'm going to change it a little bit.

hawkular-alerters/hawkular-prometheus-alerter/deployments/hawkular-prometheus-alerter-standalone/src/main/webapp/WEB-INF/jboss-web.xml
```diff
@@ -18,5 +18,5 @@
 
 -->
 <jboss-web>
-  <context-root>/hawkular/__alerter_prometheus</context-root>
+  <context-root>/hawkular/alerter/prometheus</context-root>
```
lucasponce, line 21: I unified the url with 2.x.
I guess there is no clash there and in this way, this endpoint are consistent with rest use cases.

lucasponce:
Please @jshaughn take a look at this as I would need it to bump version in hServices.
Thx

jshaughn:
Right, since we actually do expose an endpoint for notification consumption I think this makes sense.

PR #490, opened 2017-11-22
At startup we need to perform ispn reindexing

jshaughn:
@lucasponce please review.


lucasponce:
As commented, it could be nice to maintain the flag just in case we might want to skip this ?

hawkular-alerts-commons/src/main/java/org/hawkular/alerts/cache/IspnCacheManager.java
```diff
@@ -34,6 +37,8 @@
  */
 public class IspnCacheManager {
     private static final MsgLogger log = MsgLogging.getMsgLogger(IspnCacheManager.class);
+    private static final String ISPN_BACKEND_REINDEX = "hawkular-alerts.backend-reindex";
+    private static final String ISPN_BACKEND_REINDEX_DEFAULT = "false";
```
lucasponce, line 41: Perhaps we can set default true, but that's minor, having the ability to configure this is good.

PR #492, opened 2017-11-22
Fix an Event constructor bug where eventType is left unset.  

jshaughn:
This can then cause failed queries.


lucasponce:
Good catch.

PR #495, opened 2017-12-18
HWKALERTS-291 Split backend caches

lucasponce:
I wanted to perform this change some time ago, since we introduced it in inventory.
Now, we have a better split in alerting too, and as definitions population will be expected to be minor than events/alerts/actions population, then this will improve the search use cases (specially from MiQ use cases).


jshaughn:
Makes sense to me.

PR #500, opened 2018-06-04
Update to parent POM for Wildfly 11

rubenvp8510:
@jsanda please review.


jsanda:
The output says that the build passed, but travis is still reporting a failure. @rubenvp8510 can you push another commit to trigger a new build? Just make a minor change to a comment or something like that.

PR #502, opened 2018-06-11
Update to parent POM for Wildfly 11

rubenvp8510:


jsanda:
It is still failing. I think the problem might in ```.travis.yml``  when trying to clean the cache dir. Maybe try commenting out the code that removes the cache dir?

